PR #77, opened 2018-08-26
Explain STEM

probonopd:
Explain STEM as "scanning transmission electron microscopy"


probonopd:
Might even want to link to https://en.wikipedia.org/wiki/Scanning_transmission_electron_microscopy

PR #89, opened 2018-08-28
Spotselector

uellue:
This branch contains an example that extracts intensity from diffraction peaks above background.

The only change is 15bb955, the rest is from tracking the upstream branch resp. previous pull requests that were already merged.


sk1p:
Hm. We somehow need to fix this git problem, and document the right way™ to work on (and troubleshoot) feature branches for pull requests. I tried to quickly fix it myself but ran into some trouble, will need to defer for now.

Btw. there is a nice notebook viewer at 

http://nbviewer.jupyter.org/github/LiberTEM/LiberTEM/blob/a73ba4d66badec7443c859241b44aabdb6e58b8c/examples/peak_selector.ipynb

(you can construct that URL using the "view" button on the [files](https://github.com/LiberTEM/LiberTEM/pull/89/files) tab here

PR #94, opened 2018-08-30
WIP: frame cropping

sk1p:
This is just a proof-of-concept implemented for the combination of mask job on a raw dataset. Observation: works well for really small masks, for large masks it is slower than without cropping. Needs to be profiled etc.

Should have larger impact with k2is data or other large frames


sk1p:
Don't think this is still useful, needs to be re-done after #195 is merged anyways.

PR #95, opened 2018-08-31
Full encoder and decoder for little endian uint12 on little endian

uellue:
The code added to encode-decode.ipynb handles encoding and decoding
of arrays with arbitrary length with all index calculations
and treatment of dangling blocks. It was written to closely
mimick the way it would be written in C or C++ so that
it can serve as a prototype for a correct C/C++ implementation.


sk1p:
Merged, thanks!

PR #97, opened 2018-08-31
Describe mathematical operation for virtual detectors

### LINKED ISSUE #14: Mathematical formulation to allow using math libraries for applying masks to frames
For the mathematical operation it doesn't matter in what order or arrangement the frames are scanned. That means the "scan dimension" can be flattened to a linear array. The same way the arrangement of the pixels in a frame doesn't matter, that means masks and frames can be flattened to linear arrays as well.

With those extra dimensions flattened, applying a set of _m_ masks (_B_) to a collection of _n_ frames with _c_ pixels per frame (_A_) can be expressed as a [rectangular matrix product](https://en.wikipedia.org/wiki/Matrix_multiplication#Rectangular_matrices) of _A_ and _B_. _A_ is a _n x c_ matrix and _B_ a _c x m_ matrix.

That means we can use the BLAS function [GEMM](https://en.wikipedia.org/wiki/Basic_Linear_Algebra_Subprograms#Level_3). For that we should be able to find a suitable [fast implementation](https://en.wikipedia.org/wiki/Basic_Linear_Algebra_Subprograms#Implementations).

Notably:
* Scipy https://docs.scipy.org/doc/scipy/reference/linalg.blas.html#blas-level-3-functions
* Intel MKL https://software.intel.com/en-us/mkl-developer-reference-c-blas-level-3-routines
* CUDA http://docs.nvidia.com/cuda/cublas/index.html#cublas-lt-t-gt-gemm
* ATLAS http://math-atlas.sourceforge.net/
* OpenBLAS http://www.openblas.net/

uellue:
Extended the documentation with info from Issue #14. Closes #14.

### DIFF
--- a/docs/source/architecture.rst
+++ b/docs/source/architecture.rst
@@ -66,3 +66,32 @@ running on the local node.
 Processing is done in an asynchronous fashion; if you start a job the request
 immediately returns, but you get notifications about status changes, or you can
 explicitly query the API server about a specific job.
+
+Mathematical expression for applying masks
+------------------------------------------
+
+The most basic operation with pixelated STEM data is multiplying each frame
+element-wise with a mask of the same size and summing up the result for each
+frame. This way one can implement virtual detectors, center of mass and
+darkfield imaging, to name a few examples.
+
+Since each frame is processed individually and there's a 1:1 match between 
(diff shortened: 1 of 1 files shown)


sk1p:
Merged, thanks!

PR #110, opened 2018-09-06
Work on the peak selector example

uellue:
Rearranging this and that to make it presentable

I hope we can merge this before IMC so that people can immediately see this during the presentation


sk1p:
Looking good, merged! You may want to set an email address for your git user though, at the moment it is `uellue@p70.wlan.kfa-juelich.de`

PR #111, opened 2018-09-07
Some more polish on the example

uellue:
@sk1p I remembered to set my e-mail after comitting, sry. Now it is set. Does amending the commit update the author?

Disable scrolling, recalculated


uellue:
Apparently the commit still has the old e-mail. :-/ Well, now it is set.

sk1p:
It's fine, we can add a .mailmap later

PR #113, opened 2018-09-10
Rewritten center of mass example to use the Python API

uellue:
That means it is not live updating anymore. Should we have an example
dedicated to a live updating display? Or even support live updating in the Python API? As far as I can see,
it means we provide or obtain a result buffer and register a callback with run().


sk1p:
The live updating was more of an experiment/hack, to see how far I could go with the notebook user Interface. Changes are coming to ipython/jupyter that will allow truly async notebooks in the near future - then we can think about providing an AsyncContext (?) with a proper async API. 

How that API would look like I'm not sure yet, maybe even something simple like:

```
for part in await ctx.run(job, out=result):
    update_plot_with_data(result)
```

(What is part here?)

We can also think about tighter integration with hyperspy, as long as we don't have to pull it in as a dependency. Something duck-type-y.

sk1p:
I wonder... Should we clarify that the same thing is already built into LiberTEM as a complete analysis? Other than that, I like it! I'll merge as-is.

PR #135, opened 2018-10-05
First detailed description of integrated and acquisition workflows

uellue:
Starting to develop a detailed description of workflows for acquisition. Work in progress, discussion and extension highly appreciated!


sk1p:
We should look at existing software that has such integrations, what problems occur and how they solve these problems. Can't think of any on top of my head, need to research.

sk1p:
Sure, but also more generally, not limited to scientific data: software that is written as a plugin to other software, which then works on remote data and uses remote compute resources. There should be examples in the "cloud computing" area.

But the software in cloud computing is most of the time self-contained, web-based, "cloud-native", and not integrated into legacy systems. As an example: https://medium.com/pangeo/cloud-native-geoprocessing-of-earth-observation-satellite-data-with-pangeo-997692d91ca2

uellue:
I've added a workflow for @kruzaeva s work as well. Feedback highly appreciated!

uellue:
Closing since this is superseded by actual live acquisition and processing workflows by now.

PR #159, opened 2018-11-16
Logo in documentation; version info

uellue:
Copied the corporate design logo to static directory and included in configuration

Populate the version info in conf.py with info from `src/libertem/__version__`


docs/source/conf.py
```diff
@@ -28,16 +28,20 @@
...
 # -- Project information -----------------------------------------------------
 
 project = 'LiberTEM'
 copyright = '2018, LiberTEM Authors'
 author = 'LiberTEM Authors'
 
+_version_bits = libertem.__version__.split('.')
 # The short X.Y version
-version = ''
+version = _version_bits[0] + '.' + _version_bits[1]
```
sk1p: Hm, what about the patch/fix part of `__version__`? Should they not be part of `version`?

uellue: Conf has two variables, version and release. Version is supposed to be "X.Y" according to the inline documentation in conf.py.

uellue: `release` is the full `libertem.__version__`

uellue: To be honest, I still didn't find a place where these variables actually show up in the documentation.

sk1p: Ok, looks good, we will still need to adapt the template to include the version in the footer etc.

PR #162, opened 2018-11-20
Included big endian example in RAW opening dialog

### LINKED ISSUE #144: Implement big-endian raw import
Another field to select between little- and big-endian raws in the open data menu would be  helpful.

uellue:
Include a numpy dtype example for big endian raw data in the RAW file opening dialog. Fixes #144

This commit includes fixes in setup.py to allow building the JS client on Windows. Credit goes to @sk1p !


sk1p:
Looks good on Linux too, merging!

PR #173, opened 2018-11-22
Fixed correct data in mask analysis result

uellue:
The mask analysis now returns the correct data, there was a mixup.

The __str__ method for an AnalysisResult now gives more verbose output
that helps to inspect the contents of analysis results.


uellue:
Removed the bugfix because @sk1p fixed it in parallel :-)

PR #174, opened 2018-11-22
Examples use the newly created analysis facilities

uellue:
Use analysis instead of job where appropriate.


sk1p:
Looking good, thanks!

PR #176, opened 2018-11-23
Extended example to use all functions

uellue:


sk1p:
Merged. Wish there was a notebook diff function in github, :grin:

PR #177, opened 2018-11-23
Support for sparse matrices

### LINKED ISSUE #178: Allow using sparse matrices
Sparse matrices can, in principle, be more efficient than dense ones for operations like binning, point analysis, reshuffling. 

However, they can't be used transparently, currently. #177 introduces very basic support, but the matrix multiplication itself doesn't work yet for unknown reasons.

uellue:
Edit: The second commit implements experimental sparse matrix support. Closes #178.

The code should be reviewed, tested and refactored. In particular, mixing dense and sparse matrices should be handled. It might be cleaner to create a new sparse_mask_job job type and forbid mixing? @sk1p how would you handle this?

As a side note, sparse matrices can be constructed correctly from dense matrices as well, it is just not efficient if the matrix is actually dense.

----
Was:
This commit adds very basic support for sparse matrices

The modification of _make_mask_slicer() allows handling sparse matrices
since scipy.sparse matrices allow reshaping into matrices with (-1, 1),
not into vectors with (-1, ), and they support hstack() and vstack(),
but not stack(). The result of the function is the same.

The computation itself doesn't work yet for reasons that have to be
investigated. The test for sparse matrices allows to handle the situation
gracefully with a NotImplementedError instead of crashing.

The example currently triggers this error.


uellue:
Added a comment on why there's so much transpose going on

uellue:
Changing the point analysis to sparse matrix gives a substantial speedup on my laptop. To be tested on Linux and on systems with more RAM and CPU.

src/libertem/job/masks.py
```diff
@@ -83,8 +89,37 @@ def _compute_masks(self):
...
+            else:
+                return a
+
+        def to_sparse(a):
+            if sp.issparse(a):
+                return a
+            else:
+                return sp.csr_matrix(a)
+        raw_masks = [f().astype(self.dtype)
+            for f in self.mask_factories]
```
sk1p: Minor nitpick: I like to have multi-line list comprehensions aligned, maybe something like this:

```suggestion
        raw_masks = [
            f().astype(self.dtype)
            for f in self.mask_factories
        ]
```

(hope this suggestion stuff actually works, I think it doesn't get that I'm replacing two lines here...)

src/libertem/job/masks.py
```diff
@@ -19,20 +20,24 @@ def _get_masks_for_slice(slice_):
             # .reshape((-1,)) -> like flatten, but no copies
             # should save us one copy as we np.stack() immediately afterwards
             # https://stackoverflow.com/a/28930580/540644
-            slice_.get(mask, signal_only=True).reshape((-1,))
+            slice_.get(mask, signal_only=True).reshape((-1, 1))
```
sk1p: Can you explain why this is necessary? Just a difference between numpy arrays and scipy sparse matrices?

src/libertem/analysis/masks.py
```diff
@@ -16,17 +16,25 @@ def dtype(self):
...
-        job = ApplyMasksJob(dataset=self.dataset, mask_factories=mask_factories)
+        use_sparse = self.get_use_sparse()
+        job = ApplyMasksJob(dataset=self.dataset,
+                mask_factories=mask_factories, use_sparse=use_sparse)
         return job
 
     def get_mask_factories(self):
         raise NotImplementedError()
 
+    def get_use_sparse(self):
```
sk1p: As discussed, default impl which returns `False` would be nice, so subclasses can opt-in to sparse matrices.

src/libertem/api.py
```diff
@@ -66,7 +66,7 @@ def load(self, filetype: str, *args, **kwargs) -> DataSet:
 
     load.__doc__ = load.__doc__ % {'types': ", ".join(filetypes.keys())}
 
-    def create_mask_job(self, factories, dataset):
+    def create_mask_job(self, factories, dataset, use_sparse=None):
```
sk1p: As discussed, add `use_sparse` to docstring.

src/libertem/api.py
```diff
@@ -91,9 +91,10 @@ def create_mask_job(self, factories, dataset):
         return ApplyMasksJob(
             dataset=dataset,
             mask_factories=factories,
+            use_sparse=use_sparse
         )
 
-    def create_mask_analysis(self, factories, dataset):
+    def create_mask_analysis(self, factories, dataset, use_sparse=None):
```
sk1p: As discussed, add `use_sparse` to docstring.

src/libertem/analysis/point.py
```diff
@@ -13,12 +13,15 @@ def get_results(self, job_results):
...
+
     def get_mask_factories(self):
         cx = self.parameters['cx']
         cy = self.parameters['cy']
 
         def _point_inner():
-            a = np.zeros(self.dataset.shape[2:])
-            a[int(cy), int(cx)] = 1
-            return a.astype(self.dtype)
+            a = sp.csr_matrix(((1,), ((int(cy),), (int(cx),))),
```
sk1p: Does this need to be a tuple here? I try to avoid tuples of length 1 in more complicated expressions, as the syntax can become a bit ugly. Maybe:

```suggestion
            a = sp.csr_matrix(([1], [[int(cy)], [int(cx)]]),
```

If this is equivalent.

src/libertem/job/masks.py
uellue: @sk1p Sparse matrices can only be reshaped to 2D, not 1D. Technically one could write this somehow that dense arrays get a different treatment here. I felt that it would be cleaner to have equivalent code for sparse and dense.

src/libertem/analysis/masks.py
uellue: Done! :-)

src/libertem/api.py
uellue: Done! :-)

src/libertem/api.py
uellue: Done! :-)

src/libertem/analysis/point.py
uellue: Indeed better, coming with next commit.

src/libertem/job/masks.py
uellue: A new version is coming with next commit, I hope it is what you meant?

src/libertem/analysis/point.py
uellue: The sparse matrix constructor was a bit picky and only accepted lists for the innermost part. In any case, more readable now, thx! :-) 👍

PR #183, opened 2018-12-03
Catch an exception on Windows and handle it gracefully

uellue:
Attempting to determine the owner of files like C:\pagefile.sys
on Windows triggers an "Access is denied" error.

* Transform the specific Windows exception to a general IOError
in win_tweaks.py
* Handle the IOError gracefully in web/server.py


sk1p:
Merged, thanks!

PR #185, opened 2018-12-05
Support using an external local dask cluster

uellue:
Change the default behavior of `get_location()` to return `None`.
dask interprets this as "any worker".

Use `get_location()` independent of `is_local`. Verify that it is either
"None" or not empty. (thx @sk1p!)

Remove the non-functional get_locations() implementations from the
subclasses. "127.0.0.1" doesn't always work as a default identifier.

The remaining purpose of is_local is to determine if we should tidy
up the cluster or leave it running.


sk1p:
Thanks, merged! And sorry for the time wasted by the incomplete impl. there :grin:

Issue #157, opened 2018-11-13
Switch computational backend to 2D/3D partitions / tiles


sk1p:
* support for vector scans / spectral data
 * general support for time/tilt/... series
 * some file formats are inherently 3D in shape (that is, stacks of frames) which makes the mapping file <-> 4D partition awkward
 * keep 4D info in metadata

sk1p:
See also: #99

uellue:
I guess #195 closes this for the back-end?

Issue #99, opened 2018-09-03
Tileshapes refactor


sk1p:
Currently, we have tiles that are always 4D shapes. This is a problem if the number of frames per tile doesn't evenly divide the scan width: tiles near the right edge will not be of the same shape as others. Example:

Dataset shape: `(16, 16, 128, 128)`
Tileshape: `(1, 6, 128, 128)` 

For the first row, we will have three tiles. The first two will have shape `(1, 6, 128, 128)`, the last one `(1, 4, 128, 128)`. That means for the whole right border we have potentially inefficient cache usage while applying masks.

We can fix this by using 3D tileshapes: they can "spill over" into the next row of frames.

Pro: 
- other than the very last tile, they will all be of the same shape
- less logic for the borders in dataset implementations
- processing non-rectangular scans is easier

Con: 
- no easy way to have a local environment of frames in one go (note that this is also not yet possible for all readers: HDF5 and mmap'ed readers could easily support this, others need some work)

The idea would be to support both 3D and 4D tiles and use the most appropriate for the current job.

uellue:
I guess #195 closes this?

Issue #86, opened 2018-08-28
Test files for datasets


sk1p:
We need better test coverage for datasets. This could be done by having a small (<1MB?) example file for each file format we support in our repository, and testing all relevant functions on real data.

- [ ] MIB (different bit depths, different partitioning (number of frames per `.mib` file)) #129 
- [x] RAW (different variants, including endian-ness #144 ,can easily be generated)
- [x] HDF5 (can easily be generated)
- [ ] BLO 
- [ ] K2IS #63 #109 
- [ ] FRMS6  #155
- [ ] MRC #152 
- [ ] Norpix Seq #153 
- [ ] TIA SER/EMI

For K2IS we should also have some additional files that exhibit the weird corner cases of the format.

sk1p:
cc @woozey

sk1p:
We need to make sure to not include the test files in the Python distribution, otherwise we will run into issues, see for example: https://github.com/hyperspy/hyperspy/issues/2048 - I'd put the files onto a dedicated hosting service (maybe bintray?) and download them for testing. That also allows for more realistic test files if we don't have to watch size as mutch (e.g. small but real scans instead of all 0s).

sk1p:
There are multiple K2IS files available, for which we have the permission to publish them for testing purposes. They range from 8GB to 177GB. So: good for testing by users, not so good for automated testing in the CI pipeline.

@woozey will record new test data in the next weeks, which will be used for validating our reader. There is a minimum scan size of 16x16, which should be approx. 1.2GB of data in total. This could work with [TravisCI](https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system) where we should have ~9GB of storage available in total. If we don't care about the actual data, we could null those areas in the file and just keep the headers, and then heavily compress the result.

The testing should remove the files after all testcases that require them have run, which is easily doable with [pytest fixtures](https://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code).

As the disk space requirements and run time are not trivial, we should think about only running these tests when they are explicitly requested, and in CI.

woozey:
@sk1p As far as I remember the acquisition software limits the scan size to 20x20. In any case I will double check it and generate new datasets for you also for troubleshooting reasons.

sk1p:
We now have a 20x20 test dataset from the K2, thanks to @woozey ! We can convert this to raw binary (even different pixel formats), and set up I/O tests and benchmarks.

[...]

PR #195, opened 2018-12-10
Dimensionality refactoring

### LINKED ISSUE #182: Slow tests
There are two tests that take much longer than all others (output of `tox -e py36 -- --durations=5`):

```
========================================================================= slowest 5 test durations ==========================================================================
4.56s call     tests/test_analysis_masks.py::test_weird_partition_shapes_1
1.11s call     tests/test_analysis_masks.py::test_subframe_tiles
0.14s call     tests/io/test_k2is_uint12.py::test_encode_decode_uint12_ref
0.14s call     tests/io/test_k2is_uint12.py::test_encode_decode_uint12
0.12s call     tests/test_analysis_masks.py::test_mask_uint
========================================================================= 58 passed in 8.27 seconds =========================================================================
```

`test_weird_partition_shapes_1` runs three mask analyses, each over 64 quite small partitions. In 

`test_subframe_tiles` it is just many small tiles.

This needs to be profiled and then we can decide if this is a performance bug in mask application or if we should reduce the number of partitions for improving testing speed.

sk1p:
First part of making LiberTEM more flexible regarding dimensionality of data. Introduces a `Shape` class that knows how many dimensions the signal has, a bit like hyperspy (simpler, keeping numpy ordering of dimensions etc.).

Tests should pass, but support in the I/O layer is not complete yet.

Closes #182, #157, #99


src/libertem/common/slice.py
```diff
@@ -1,68 +1,64 @@
...
+from libertem.common.shape import Shape
 
 
 class Slice(object):
+    __slots__ = ["origin", "shape"]
+
     def __init__(self, origin, shape):
         """
-        A slice into a 4D dataset, defined by origin and shape
+        A ND slice, defined by origin and shape
```
uellue: n-dimensional for clarity?

src/libertem/common/slice.py
```diff
@@ -1,68 +1,64 @@
...
+                    len(self.origin), len(self.shape)
+                )
+            )
+        if not hasattr(shape, 'to_tuple'):
+            raise ValueError("please use libertem.common.Shape instance as shape parameter")
 
     def __repr__(self):
         return "<Slice origin=%r shape=%r>" % (self.origin, self.shape)
 
     def __hash__(self):
```
uellue: Perhaps a comment what this is used for?

tests/test_analysis_masks.py
```diff
@@ -55,7 +55,7 @@ def test_merge_masks(masks):
```
uellue: # FIXME test coverage for permutations (nav, sig): (1, 2) (time series); (2, 1) (spectral), (2, 3) (hyperspectral)

This is critical because the whole point of moving to n-dimensional data sets is breaking the (2, 2) assumption. This needs test coverage.

tests/test_shape.py
```diff
@@ -0,0 +1,71 @@
+from libertem.common import Shape
```
uellue: 😃  :+1:

src/libertem/job/sum.py
```diff
@@ -9,25 +9,21 @@ def get_tasks(self):
...
             data = data_tile.data
             if data.dtype.kind == 'u':
                 data = data.astype("float32")
             result = data_tile.data.sum(axis=(0, 1))
-            tile_slice = data_tile.tile_slice.get()
-            part[
-                tile_slice[2],
-                tile_slice[3],
-            ] += result
+            part[data_tile.tile_slice.get(sig_only=True)] += result
```
uellue: A bit of documentation on how to use the Shape could be nice. This point here is extremely elegant and nice. Just not obvious to do it like that for someone using Shape. Could be a good example for documentation.

uellue: Idea for follow-up: How does Hyperspy do this? Perhaps comparison or improvement?

src/libertem/job/masks.py
```diff
@@ -128,7 +127,7 @@ def _compute_masks(self):
     def get_masks_for_slice(self, slice_):
         if self._get_masks_for_slice is None:
             self._get_masks_for_slice = _make_mask_slicer(self.computed_masks)
-        return self._get_masks_for_slice(slice_.discard_scan())
```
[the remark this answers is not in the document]
sk1p: Yep, still needed!

src/libertem/io/dataset/blo.py
```diff
@@ -2,7 +2,7 @@
 
 import numpy as np
```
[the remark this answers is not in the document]
sk1p: Good point. I started to work on this, but there is still the task of hosting the test data somewhere (Zenodo as open data?), and even maybe getting more test data (different bit depths for MIB for example). Right now I'm just making it work on my local computer, and skip the tests if the test data doesn't exist.

See also #86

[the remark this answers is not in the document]
uellue: Zenodo sounds good. We'd need permission and a proper description of the data sets. :-)

tests/test_analysis_masks.py
uellue: And perhaps (1, 1) for good measure. :-)

sk1p:
Remaining topics of discussion:

 * naming: I'm not convinced `DataSet.effective_shape` is the best name for its purpose
 * integration of n-dimensional methods into the API
 * tests for `libertem.io`
 * GUI changes? Can be done after this is merged though

uellue:
> * naming: I'm not convinced `DataSet.effective_shape` is the best name for its purpose

What about `raw_shape` for the native dataset shape, and just `shape` for the best way to interpret the data as a user, how I understand `effective_shape` right now?

> * integration of n-dimensional methods into the API

As a general opinion, I'd keep the specific Analyses that make assumptions about the data as-is. COM, disk, ring only make sense for 2D frames. Where there are no implicit assumptions, the functions return data in the "correct" shape following the shape of the data set, i.e. a mask job on a (8,8,8) nav and (16,16) sig data set will accept (16,16) masks and return a (8,8,8) result, like it is already implemented. For spectral or hyperspectral data we create new Analyses as needed. The goal for an Analysis is to have a simple and very clear tool for scripting and the GUI. 

A Job, as opposed to an Analysis, should offer a more general interface to the back-end. The only change I'd make is to `create_pick_job()`. Discussion with @sk1p: Accept `origin=(z, y, x, ...)` and `shape=(z, y, x, ...`) tuples. If `shape is None`, return a point.

To have a convenient interface, add a create_pick_analysis(self, dataset, x, y=None, z=None) that throws an exception if the given parameters don't match the nav dimension. 

* One nav dimension: Require x, throw exception if y or z are not None
* Two nav dimensions: Require y and x, throw exception if z is not None
* Three nav dimensions: Require z, y and x
* More than three nav dimensions: Throw exception and suggest using create_pick_job()

> * GUI changes? Can be done after this is merged though

That should be a separate pull request, probably. :-) There we can also create the Analyses as needed. :-)

sk1p:
> What about `raw_shape` for the native dataset shape, and just `shape` for the best way to interpret the data as a user, how I understand `effective_shape` right now?

Agreed and done!

sk1p:
`create_pick_job` changes and `create_pick_analysis` impl pushed. Should be ready for check-in. Do you want to look over it again or should I merge?

PR #203, opened 2018-12-12
Correct method to launch workers with embedded interpreter

uellue:
Thank you @itamarst for pointing in the right direction! 

See https://github.com/dask/distributed/issues/2401


sk1p:
Looking good. To get the test coverage back up, I'll soon work on tests for the dask executor(s) and the web API.

Issue #205, opened 2018-12-12
Do all filesystem operations via the executor


sk1p:
We want to do all filesystem operations via the executor. This actually has two reasons:

1) They may take some time, for example when listing a large directory, or listing a NFS share. Running them via the executor means we get to do this in the background.
2) The worker nodes may not have the same filesystem as the API server or `dask.distributed` `Client`

In the first iteration, this means we make sure `Context.load(...)` pushes the dataset load operation into the cluster, including the `check_valid()` call. The web API basically needs the same changes.

The next step would be to update the web API to also do directory listing / existence checks etc. via functions pushed into the cluster. 

This is highly related to #154 which also may need to do more work on dataset initialization. Also related is #186 as listing a directory and opening files should take priority over another job that may take a while to finish.

- [x] add ability to run simpler functions via the executor
- [x] update `Context` and `DataSetDetailHandler` to  do `load` and `check_valid` via executor
- [x] test connecting via `Context` to a remote dask scheduler
- [x] update `DataSetDetectHandler` 
- [x] update `LocalFSBrowseHandler`
- [x] test starting the web client with a connection to a remote dask scheduler

sk1p:
As of ad09ac8798ca72b13bf5ab67c603e5d38eeacd0c it is possible to spawn a dask cluster on a remote host, connect to it (for example by starting a LiberTEM server locally, and then connecting via GUI) and work with the files on the remote system.

Currently, this is a bit brittle, as certain methods on the dataset need to make sure they don't access the filesystem (they are used on the client to calculate partitioning etc.), and that the initialization function is called. If I get an idea how to structure this better, maybe it can be made more robust.

uellue:
Tested with the async Digital Micrograph prototype with processing on moellenstedt. It ROCKS! 👍 👍 👍

PR #209, opened 2018-12-13
[skip ci] Bug reproducer for incorrect data shape

uellue:
Somewhere somehow the dimensions are mixed up when Digital Micrograph
constructs the image buffer numpy array.

Current workaround is a reshape of the data when writing into the buffer

This will likely change in the future when this is corrected.

### DIFF
--- a/prototypes/digitalmicrograph/components.py
+++ b/prototypes/digitalmicrograph/components.py
@@ -1,30 +1,20 @@
-import inspect
-
 import numpy as np
 
-changed = False
 
 class EventHandler(DM.Py_ScriptObject):
 
-        
-    def HandleComponentChangedEvent(self, img_disp_event_flags, img_disp, component_change_flag, component_disp_change_flags, component):
-        changed = True
-        print(changed)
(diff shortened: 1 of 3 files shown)


uellue:
Sorry, I forgot that I still had an open pull request in the gatan branch. Now it is two changes in one. Should I wrestle with git to pick the two apart or is it OK like that?

The second commit successfully tests #205 

Even on moellenstedt this is impressively fast. I can't wait to see it on the cluster!

sk1p:
> Sorry, I forgot that I still had an open pull request in the gatan branch. Now it is two changes in one. Should I wrestle with git to pick the two apart or is it OK like that?

OK like this. Do you want to keep the commented parts, or clean it up a bit?

> The second commit successfully tests #205 
> Even on moellenstedt this is impressively fast. I can't wait to see it on the cluster!

:+1:

uellue:
I've cleaned it up.

I've left the code to start a local cluster and left the alternative data set because that might come in handy later.

PR #226, opened 2018-12-21
Better random data with guaranteed asymmetry; numerical test

### LINKED ISSUE #225: More stringent mathematical tests
Right now, the tests do cover the basics regarding mathematical correctness. However, two items should be addressed:

- [ ] We should test for numerical errors and instabilities, for example sums of many very small numbers or sums of values with a high dynamic range, similar to our input data.
- [ ] We should make sure that our test data has "good randomness", i.e. strong asymmetries and value spread. Right now, arrays are usually populated with `np.random.choice(a=[0, 1], ...)` and results are checked with `np.allclose(..)`. That could miss errors like flipped dimensions, mixed up partitions etc. because errors would average out if the data sets are large enough.

uellue:
Closes #225

### DIFF
--- a/tests/io/test_hdf5.py
+++ b/tests/io/test_hdf5.py
@@ -1,10 +1,10 @@
 import numpy as np
 
-from utils import _naive_mask_apply
+from utils import _naive_mask_apply, _mk_random
 
 
 def test_hdf5_apply_masks_1(lt_ctx, hdf5_ds_1):
-    mask = np.random.choice(a=[0, 1], size=(16, 16))
+    mask = _mk_random(size=(16, 16))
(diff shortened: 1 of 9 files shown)


sk1p:
Failing testcase seems to be caused by missing import of `_mk_random` in the local cluster tests.

uellue:
WIP because I'll try sth about the numerical precision as well.

sk1p:
On the topic of numerical stability (for `np.sum`, similar problems may apply to OpenBLAS): https://github.com/numpy/numpy/issues/8786

uellue:
Closes #225

sk1p:
Thanks, looks good! Btw. I don't think "closes" etc. works as comment, edited this into the description of the PR.

Issue #243, opened 2019-01-11
Support for ser/emi files (TIA)


uellue:
Support for ser/emi is required for a specific project. 

http://www.er-c.org/cbb/info/TIAformat/
http://www.er-c.org/cbb/info/TIAformat/TIAseriesformat.pdf
http://hyperspy.org/hyperspy-doc/current/user_guide/io.html#fei-tia-ser-and-emi
https://ercius.wordpress.com/2014/01/21/electron-microscopy-data-files-emi-ser-update/
https://imagej.nih.gov/ij/plugins/tia-reader.html

@ercius https://github.com/ercius/openNCEM/blob/master/ncempy/io/ser.py that is the reader you wrote? Any chance that you could re-license it under a more permissive open source license? :-)

- [x] initial support, available via API
- [ ] integration tests
- [x] GUI support
- [ ] performance analysis

ercius:
Yes, that's our reader! :)

Changing the license is on my list of things to do. I'm putting together a
new release with a few updates and bug fixes. The new license will be
included. I'll do it by next week. Thanks for your patience.

This will 100% happen. You can start using it if you need it sooner.

uellue:
@ercius 👍 👍 👍  Great, thank you! :-)

uellue:
@ercius 🎉 🍾  Thank you! @sk1p just said he'll get started with that today. 👍

sk1p:
Initial support for SER/EMI files via ncempy is now available in the Python API, thanks again @ercius! I still need to add the GUI part, write integration tests and do a performance analysis. As I only have 3D datasets available (line scans), which are not yet really suported in the GUI, I'll postpone the GUI part for now. I've updated the issue description with the items still TODO.

@uellue for whatever reason, the link checker doesn't like the ER-C link, which makes the travis build fail. Any idea about that?

uellue:
Wasn't that implemented in #243? Can we close this?

sk1p:
So, the two unchecked boxes are still relevant: there are no integration tests yet, and I don't know about the performance - I haven't seen any large sample files yet. The second point is optional, I'd say (if no one screams...), the first one still needs to be done.

uellue:
We do run this in CI and performance hasn't really been an issue so far, as far as I am aware. Closing!

PR #251, opened 2019-01-23
Fix broken links

uellue:
Conda documentation has moved

fz-juelich.de filters the user agent string since recently. I asked them to stop filtering. Until then, fz-juelich.de is on the ignore list of the link checker as a workaround.

Refs #243


sk1p:
Thanks!

PR #255, opened 2019-01-26
Gridreduction

uellue:
This prototype code can reduce coordinates that correspond to the output of blob detection and blob tracing to a best fit parallelogram grid. It can separate and reconstruct several mixed grids in one collection and separate a grid from outliers.

TODO items until production:
- [ ] Optimize robustness and "magic number" parameters with real-world data
- [x] Optimize performance. The code is still very inefficient and way too slow for application.
- [x] Optimize real-world application on frames. Usually, not every frame needs a full detection since many, many frames share the same lattice vectors.


uellue:
This should be ready for integration now.

Issue #181, opened 2018-11-28
Basic tests for Web API


uellue:
With the experience of #180 and #179 it would be desirable to have at least basic test coverage for the Web API. That would have caught these bugs. In particular testing both on Linux and Windows on a clean install is a great help.

sk1p:
Relevant projects:

 - https://github.com/eugeniy/pytest-tornado - sadly not maintained currently
 - https://github.com/pytest-dev/pytest-asyncio - this could help to start a temporary test server
 - https://www.tornadoweb.org/en/stable/testing.html - built in testing support of tornado using the unittest framework, should be runnable via pytest, but means writing the test cases differently

My current idea is to implement something using pytest-asyncio, which then is also less dependent on tornado, should we get the idea of using a different framework for the web API

sk1p:
There is an [active fork of pytest-tornado](https://github.com/vidartf/pytest-tornado) that is published at https://pypi.org/project/pytest-tornado5/ - this may be another option that could be more straight-forward than using pytest-asyncio. As an example project using pytest-tornado (not the fork though) I looked at jupyterhub:

 - fixture to create a mock tornado app: https://github.com/jupyterhub/jupyterhub/blob/9a3dbedc52b1894b08c4a7fa9099e20b3f9a3f69/jupyterhub/tests/conftest.py#L63
- example tests: https://github.com/jupyterhub/jupyterhub/blob/9a3dbedc52b1894b08c4a7fa9099e20b3f9a3f69/jupyterhub/tests/test_pages.py

uellue:
Looks simple enough! :-)

sk1p:
Maintenance discussion: https://github.com/eugeniy/pytest-tornado/issues/41

PR #269, opened 2019-02-21
Use distributed.Client instead of distributed.asyncio.AioClient

uellue:
As suggested by @danpf in https://github.com/dask/distributed/issues/2537#issuecomment-466050723

Many thanks @danpf, this seems to work!

Confirmed to work with distributed==1.25.3

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -132,7 +132,7 @@ setup(
     install_requires=[
         "numpy",
         "scipy",
-        "distributed<=1.25.2,>=1.23.3",
+        "distributed>=1.23.3",
         "click",
         "tornado",
         "matplotlib",
--- a/setup.py
+++ b/setup.py
@@ -134,7 +134,7 @@ setup(
         "scipy",
         "distributed>=1.23.3",
         "click",
-        "tornado",
+        "tornado>=5",
         "matplotlib",
         "pillow",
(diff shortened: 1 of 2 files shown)


sk1p:
Thanks for handling this! We may to also need to depend on `tornado>=5` for this to work, right?

Also another reason to work on #181 soon, which would have detected this issue for us.

uellue:
> Thanks for handling this! We may to also need to depend on `tornado>=5` for this to work, right?

Yes, confirmed with test. See new commit in this PR for new version dependency!

> Also another reason to work on #181 soon, which would have detected this issue for us.

Yes! :-)

sk1p:
Btw, the reason it requires tornado 5: https://www.tornadoweb.org/en/stable/releases/v5.0.0.html#highlights

> The focus of this release is improving integration with asyncio. On Python 3, the IOLoop is always a wrapper around the asyncio event loop, and asyncio.Future and asyncio.Task are used instead of their Tornado counterparts. This means that libraries based on asyncio can be mixed relatively seamlessly with those using Tornado

As I understand it, before tornado 5, interoperability between tornado and asyncio apps was limited, hence `AioClient`.

sk1p:
Merged, thanks!

PR #274, opened 2019-02-22
Fit quality and weighted fit

uellue:
The correlation code now calculates a fit quality function, and the grid matching code
uses that to eliminate really bad peaks and to assign weights.

Work in progress: The gridmatching implementation has grown rather messy and needs
a thorough cleanup.

Independent of that, the result in the jupyter notebook is looking very promising.


src/libertem/analysis/gridmatching.py
```diff
@@ -221,21 +218,19 @@ def fastmatch(cls, correlation_result: CorrelationResult, zero, a, b, parameters
     def full_match(cls, correlation_result: CorrelationResult, zero, cand=[], parameters={}):
```
sk1p: Using mutable objects as default value means they are shared between calls. The best practice here is to use None instead, check with `is None` and initialize with a default value in the function body.

PR #298, opened 2019-03-18
Fixing issues in the Simple Libertem.ipyn

irahulcse:
I had converted the .ipynb file to .py file and then I  had tried to change it into the single frame.
Ad

- Changed the ctx_create_pick_job(dataset=ds,scan_x//2,scan_y//2)
with

- origin=(scan_x//2,scan_y//2,scan_z//2)
pick_job = ctx.create_pick_job(dataset=ds,origin=origin)

That have three parameters in them.
As read from the given link :https://libertem.github.io/LiberTEM/reference.html#libertem.api.Context.create_pick_analysis


irahulcse:
If there is some other way to change it to the single frame then pls comment on it.

sk1p:
Hi Rahul and welcome to LiberTEM.
First of all, thank you for your first pull request!

> I had converted the .ipynb file to .py file 

We actually would like to keep the example in the jupyter notebook (`.ipynb`) format. If you need help to get jupyter running, please let me know (I'm also available in the gitter chat). If you want to add an additional, non-jupyter example file, I'm open to that as well. In that case, some more cleanup is needed.

> ```origin=(scan_x//2,scan_y//2,scan_z//2)```

This is almost correct; please note the following from the documentation you linked: 

> Number of parameters must match number of navigation dimensions in the dataset, for example if you have a 4D dataset with two signal dimensions and two navigation dimensions, you need to specify x and y.

In this case, the dataset is indeed 4D with two signal dimensions, which means you must only specify x and y parameters. `scan_z` is not even defined in this context, so if you try to run it, you will get the following error: `NameError: name 'scan_z' is not defined`.

Did @uellue already give you access to our test dataset? You can run this example code with that dataset by just changing the `path` parameter in the `ctx.load` statement.

That's it for now! Please, don't be discouraged by the long comment, it is very normal that a first pull request requires some more work and iterations to get it right.

sk1p:
Sorry, the documentation I quoted is actually in `create_pick_analysis`, not `create_pick_job`. Sorry, even I mix these up :) Nevertheless, the `origin` parameter needs to either specify only the navigational dimendions (2 in this case), or all dimensions (4 here). For reading whole frames, specifying 2 dimensions should work. As always, you can test this by running the notebook.

I realize the concepts of navigation/signal etc. are not well-documented; if you have any questions, please do ask!

irahulcse:
Thanks @sk1p I will take care of the mistakes in the future. I got slightly confused how to test the file in the .ipynb format that's why I had converted into the .py file and tried to change according to the [link](https://libertem.github.io/LiberTEM/reference.html#libertem.api.Context.create_pick_analysis). And I will update the commit from three parameters to the two. I wanted to know what was the issue with the .ipynb file?. According to me, I had changed the file according the new single frame work.

irahulcse:
check

sk1p:
> I got slightly confused how to test the file in the .ipynb format that's why I had converted into the .py file and tried to change according to the [link](https://libertem.github.io/LiberTEM/reference.html#libertem.api.Context.create_pick_analysis). 

`.ipynb` files are jupyter notebooks. The way to test them is to install jupyter (for example via `pip install jupyter`, see [their documentation for details](https://jupyter.readthedocs.io/en/latest/install.html)), and run `jupyter notebook` in the directory containing the `.ipynb` file. You can then execute the notebook cell-by-cell, or the whole file. There is quite a lot of documentation on jupyter available on the web, for example [this document on notebook basics](https://nbviewer.jupyter.org/github/jupyter/notebook/blob/master/docs/source/examples/Notebook/Notebook%20Basics.ipynb). You do need to change the path in the example to make it work.

Make sure to install jupyter into the same python environment as libertem for the example to work.

> I wanted to know what was the issue with the .ipynb file?

The parameters of the `create_pick_job` method changed and the file was not updated to the new interface. `create_pick_job` used to accept x and y coordinates, now it is a bit more general and accepts shape/origin. 

> According to me, I had changed the file according the new single frame work through the create_pick_analysis().

Btw. there was an error in the documentation; the `create_pick_analysis` documentation used the same example as `create_pick_job`; this is now fixed. The example is now fixed.

Let me know if you need any more help, with jupyter or with testing the changes; and sorry for the misleading documentation!

sk1p:
I'm going to close this one in favor of the new PR, #309. In the future, you can do this yourself when you create a new pull request that should replace an old one: just close the old PR (close button is just below this comment box).

PR #300, opened 2019-03-19
executor: use sync dask executor and wrap it for async execution

sk1p:
After this, we no longer have code duplication in the dask executor, and can use an executor instance in both sync and async code.

@uellue can you re-test your DM integration prototype with this branch? I put in a compatibility implementation of `AsyncDaskJobExecutor`, so it should work without changes to your code, but as there threading changes involved, I want to make sure this doesn't break anything under the hood.


sk1p:
The AppVeyor failure appears to be because I cherry picked some commits to master already, let me fix that.

uellue:
@sk1p I've tried it and ran into some issues. It seemed to occur on the master branch as well. I'll have to investigate more tomorrow what exactly was going on.

uellue:
Fun fun fun! Doing things with an embedded interpreter runs into a lot of edge cases that have poor coverage in tests and real world use. https://github.com/python/cpython/pull/12463

Workaround:

```python
if not hasattr(sys, 'argv'):
    sys.argv  = []
```

Now that the base version is working again, I can start testing your change.

sk1p:
> As far as I can tell, it works the same as before. 

:+1: 

> Just as a question, that means we can run any executor asynchronously, including the local executor you were considering?

Yep! Now you can also use the "stupid" `InlineExecutor` as a backend for the web API, for easier debugging, for example to check if a problem is with dask or with our code.

> Should we keep the `AsyncDaskJobExecutor` or drop it and modify the DM integration example to use AsyncAdapter directly?

It depends. We don't yet have a common interface for creating (`make_local`, `connect`), which means if you want to use `AsyncAdapter` directly, you either have some blocking code when connecting / spinning up the local cluster, or you have to manually wrap it in `sync_to_async`. For a prototype/example, blocking code would be fine, but for anything more serious, you absolutely don't want to block the event loop.

Another possibility: change the `make_local` and `connect` classmethods to create sync/async executor based on an argument, a bit like dask itself. So you would call `DaskJobExectuor.make_local(asynchronous=True)` and get back an async executor (actually, a coroutine that 'resolves' to an `AsyncAdapter` instance, which wraps the `DaskJobExecutor`). That way we can get rid of the `AsyncDaskJobExecutor`. What do you think?

I'm going to merge this now; we can do these changes independently.

uellue:
> Another possibility: change the `make_local` and `connect` classmethods to create sync/async executor based on an argument, a bit like dask itself. So you would call `DaskJobExectuor.make_local(asynchronous=True)` and get back an async executor (actually, a coroutine that 'resolves' to an `AsyncAdapter` instance, which wraps the `DaskJobExecutor`). That way we can get rid of the `AsyncDaskJobExecutor`. What do you think?

To be honest, I'd have to understand the use case a bit better. From my gut feeling, I'd say doing it like Dask sounds like a good plan. Anyway, not a real issue at the moment. I'd say we can sit on that one for a while?

sk1p:
:+1:

PR #302, opened 2019-03-20
Add variants for local and remote tests with Digital Micrograph

uellue:
These variants allow me to run the tests for various dask configurations
without having to modify files each time


sk1p:
I guess it is not possible to make this more DRY? Just putting most code into a module and importing it from the variants would break the coding workflow in DM?

uellue:
Hm, I know what you mean. Probably one should develop this into a proper test suite. I'll think about it and update it later.

sk1p:
Thanks, looks good!

PR #308, opened 2019-03-26
stddev basic testing

jaewshin:
I tried to implement a testing for computing mean/standard deviation in a batch of images by adding test_minibatch function on LiberTEM/tests/udf/test_simple_udf.py. Also changed the name of the variables from the original stddev.py file just for a bit more clarity.


tests/udf/test_simple_udf.py
```diff
@@ -1,9 +1,16 @@
 import numpy as np
-
 from libertem.common.buffers import BufferWrapper
+from libertem.api import Context
+from libertem.udf.stddev import minibatch, merge, part
+import collections 
 
-from utils import MemoryDataSet, _mk_random
-
+if __name__ == '__main__':
```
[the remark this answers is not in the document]
jaewshin: Ah sorry I should have looked into the documentation on running tests. Deleted it now.

tests/udf/test_simple_udf.py
```diff
@@ -53,3 +60,20 @@ def my_frame_fn(frame, pixelsum):
...
+    compute_batch = minibatch(batch)
+
+    sum_var, sum_im, N = compute_batch.sum_var, compute_batch.sum_im, compute_batch.N
+
+    assert N == len(batch) # check number of observations
+    assert np.allclose([sum_im], [sum(batch)]) # check sum of image pixels 
+
+    # check sum of variances 
+    mean = sum(batch)/len(batch)
+    var = sum([np.square(mean - im) for im in batch])
```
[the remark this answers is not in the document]
jaewshin: I changed it as recommended

jaewshin:
the main problem I'm having is with testing the sum of variances, which is the last assert statement on test_minibatch. I've run tests with some small set of numpy arrays to test merge function from udf/stddev.py, which is called inside test_minibatch, and it didn't return any errors. So I'm suspecting that the error most likely comes from how I'm computing the sum of variances of the given MemoryDataSet (but it could also be something else). If you could please have a look at it, I'd very much appreciate it!

tests/udf/test_simple_udf.py
```diff
@@ -53,3 +55,26 @@ def my_frame_fn(frame, pixelsum):
     assert 'pixelsum' in res
     print(data.shape, res['pixelsum'].data.shape)
     assert np.allclose(res['pixelsum'].data, np.sum(data, axis=(1, 2)))
+
+
+def test_minibatch(lt_ctx):
```
sk1p: You can move this function to its own file, for example `tests/udf/test_stddev.py`. It should also have a more descriptive name (it is no longer testing `minibatch`, right?)

In addition to this test with random data, a simple test with "hand-made" data would be useful, where someone could calculate the results by hand and can easily convince themselves that the test case and expected results is correct.

tests/udf/test_simple_udf.py
```diff
@@ -53,3 +55,26 @@ def my_frame_fn(frame, pixelsum):
...
+    data = _mk_random(size=(16, 16, 16, 16), dtype="float32")
+    dataset = MemoryDataSet(data=data, tileshape=(1, 1, 16, 16),
+                            partition_shape=(4, 4, 16, 16), sig_dims=2)
+
+    res = lt_ctx.run_udf(
```
sk1p: You should also have a test case that calls your top-level interface function (that is currently called `run_analysis`, see comment above). You can keep a test with `run_udf` if you want to check any intermediate results that may not be returned from the `run_analysis` function.

src/libertem/udf/stddev.py
```diff
@@ -0,0 +1,171 @@
+# import functools
+import collections
+
+import numba
+import numpy as np
+
+from libertem.common.buffers import BufferWrapper
+
+# from libertem.udf import ResultBuffer, map_frames
```
sk1p: Commented-out code can be removed

src/libertem/udf/stddev.py
```diff
@@ -0,0 +1,171 @@
+# import functools
```
sk1p: Commented-out code can be removed

src/libertem/udf/stddev.py
```diff
@@ -0,0 +1,171 @@
...
+    p0 = VariancePart(sum_var = dest['batch'][:, :, 0], sum_im = dest['batch'][:, :, 1], N = dest['batch'][:, :, 2][0][0])
+    p1 = VariancePart(sum_var = src['batch'][:, :, 0], sum_im = src['batch'][:, :, 1], N = src['batch'][:, :, 2][0][0])
+    compute_merge = merge(p0, p1)
+
+    sum_var, sum_im, N = compute_merge.sum_var, compute_merge.sum_im, compute_merge.N
+    dest['batch'][:, :, 0] = sum_var
+    dest['batch'][:, :, 1] = sum_im
+    dest['batch'][:, :, 2] = N
+
+def minibatch(partition):
```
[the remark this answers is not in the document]
jaewshin: For now, I've removed this function and "part" function. I think the main problem with current implementation is that run_udf gives us partitions of the frames but we can't access a partition at its entirety; for instance, we need to use the passed function (my_frame_fn_batch) on run_udf to perform "merge" for individual frame at the partition level and then perform another "merge" with other partitions. If we can access a partition at a time (i.e., view the whole frames within the partition at once), then this would be much cleaner process in terms of code and perhaps in terms of speed as well.

src/libertem/udf/stddev.py
```diff
@@ -0,0 +1,171 @@
...
+        N += 1
+
+    mean = sum_im / N
+
+    for frame in partition.get_tiles():
+        sum_var += np.square(mean - frame.data)
+
+    return VariancePart(sum_var = sum_var, sum_im = sum_im, N = N)
+
+def whole(dataset, batchsize = 32):
```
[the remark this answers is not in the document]
jaewshin: It's also removed!

src/libertem/udf/stddev.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+def my_buffer_batch():
+    return {
+        'batch' : BufferWrapper(
+            kind = 'sig', extra_shape = (3,), dtype = 'float32'
+            )
+    }
+
+def my_frame_fn_batch(frame, batch):
```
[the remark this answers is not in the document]
jaewshin: Will make sure of that when naming functions

src/libertem/udf/stddev.py
```diff
@@ -0,0 +1,171 @@
...
+
+    #compute sum of variances
+    sum_var = 0
+    sum_var = np.square(frame - mean)
+    for im in batch:
+        sum_var += np.square(im.data - mean)
+
+    return VariancePart(sum_var = sum_var, sum_im = sum_im, N = size)
+
+def merge(p0, p1):
```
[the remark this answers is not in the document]
jaewshin: I have. It gave me an error related to VariancePart. Maybe collections.namedtuple is not supported in numba; together with your other comment about replacing VariancePart, I will try to come up with other ways than using VariancePart

src/libertem/udf/stddev.py
```diff
@@ -0,0 +1,171 @@
...
+    sum_im_AB = p0.sum_im + p1.sum_im
+
+    # compute sum of variances for joint samples
+    delta_P = mean_B - mean
+    sum_var_AB = p0.sum_var + p1.sum_var + (p1.N * delta * delta_P)
+
+    return VariancePart(sum_var=sum_var_AB, sum_im=sum_im_AB, N=N)
+
+# @numba.njit(parallel=True)
+def part(dataset):
```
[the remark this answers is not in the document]
jaewshin: Also removed

src/libertem/udf/stddev.py
```diff
@@ -0,0 +1,171 @@
...
+    #     if current is None:
+    #         current = res
+
+    #     else:
+    #         current = merge(current, res)
+
+    return current
+
+
+def myvar(p):
```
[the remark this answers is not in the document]
jaewshin: Also removed

src/libertem/udf/stddev.py
jaewshin: done

src/libertem/udf/stddev.py
jaewshin: done

src/libertem/udf/stddev.py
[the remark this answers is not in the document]
sk1p: We thought about providing a per-partition interface, too, and we may add one in the future. We want to start smaller, though, and with https://github.com/LiberTEM/LiberTEM/issues/310 we plan to introduce an interface that gives you a tile (think small list of frames). This should then allow to implement a minibatch-like function, and improve both readability and performance. I'd be curious by how much; let's benchmark this once #310 is done.

I also propose a different by-frame interface in [a comment on #310](https://github.com/LiberTEM/LiberTEM/issues/310#issuecomment-478175692); do you think this would help with writing cleaner code? I don't think it would improve performance, though.

The reason we don't give access to the whole partition as a big buffer is, that it is easy to write inefficient code that way, which does multiple passes over a partition, thus easily being limited by memory bandwidth. By only operating on smaller tiles or single frames we make sure the current working set at least fits into the L3 cache of the CPU.

src/libertem/udf/stddev.py
```diff
@@ -0,0 +1,171 @@
...
+from libertem.common.buffers import BufferWrapper
+
+# from libertem.udf import ResultBuffer, map_frames
+
+VariancePart = collections.namedtuple('VariancePart', ['sum_var', 'sum_im', 'N'])
+
+
+def my_buffer_batch():
+    return {
+        'batch' : BufferWrapper(
```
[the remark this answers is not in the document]
jaewshin: I made a buffer function batch_buffer and within it contains 'stddev,'  'sum_frame,' and 'num_frame.' Also, N part is now type 'single'

src/libertem/udf/stddev.py
```diff
@@ -0,0 +1,171 @@
...
+def my_frame_fn_batch(frame, batch):
+
+    if batch[:, :, 2][0][0] == 0:
+        batch[:, :, 0] = 0
+
+    else:
+        
+        p0 = VariancePart(sum_var = batch[:, :, 0], sum_im = batch[:, :, 1], N = batch[:, :, 2][0][0])
+        p1 = VariancePart(sum_var = 0, sum_im = frame, N = 1)
+        compute_merge = merge(p0, p1)
```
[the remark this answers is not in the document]
jaewshin: As of right now, I'm still using VariancePart, but I will think more about this to write it more efficiently

src/libertem/udf/stddev.py
```diff
@@ -0,0 +1,171 @@
...
+        sum_var, sum_im, N = compute_merge.sum_var, compute_merge.sum_im, compute_merge.N
+
+        batch[:, :, 0] = sum_var
+
+    batch[:, :, 1] += frame
+    batch[:, :, 2] += 1
+
+def stddev_merge(dest, src):
+
+    p0 = VariancePart(sum_var = dest['batch'][:, :, 0], sum_im = dest['batch'][:, :, 1], N = dest['batch'][:, :, 2][0][0])
```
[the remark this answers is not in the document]
jaewshin: Yes, I separated buffers as suggested!

src/libertem/udf/stddev.py
```diff
@@ -0,0 +1,171 @@
...
+
+    return current
+
+
+def myvar(p):
+    return p.XX / p.N
+
+
+@numba.njit 
+def run_analysis(ctx, dataset, parameters):
```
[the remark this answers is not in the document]
jaewshin: As for your last point about returning mean and stddev, I don't actually need these values for computation and these can be added once we compute variance and sum of pixels. However, I'm not sure how to return the output in a "clean" way. For instance, one can retrieve the desired statistic from the output from run_stdddev(which is like run_fem the top level interface function) by calling the buffer that stores the desired value (e.g. res['sum_frame']).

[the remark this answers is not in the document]
jaewshin: For now, I just didn't do anything to the buffers that I don't need until the very end. Do you think it'd be okay to do that? (It's under run_stddev function)

tests/udf/test_simple_udf.py
jaewshin: For now, there is not much of a difference between run_udf and run_stddev so I just removed the test on run_udf

src/libertem/udf/stddev.py
```diff
@@ -0,0 +1,189 @@
...
+
+        var[:] = compute_merge.var
+
+    sum_frame[:] += frame
+    num_frame[:] += 1
+
+
+def batch_merge(dest, src):
+    """
+    Given two buffers that contain sum of variances, sum of frames,
```
sk1p: `dest` and `src` aren't buffers' I'd rather describe them as partial results (same below in the Parameters part)

src/libertem/udf/stddev.py
```diff
@@ -0,0 +1,189 @@
...
+        Contains information about the first partition, including
+        sum of variances, sum of pixels, and number of frames used
+
+    p1
+        Contains information about the second partition, including
+        sum of variances, sum of pixels, and number of frames used
+
+    Returns
+    -------
+    collections.namedtuple object
```
sk1p: `collections.namedtuple` is rather generic, the specific returned object is a `VariancePart`

src/libertem/udf/stddev.py
```diff
@@ -0,0 +1,189 @@
...
+
+    """
+    pass_results = ctx.run_udf(
+        dataset=dataset,
+        fn=compute_batch,
+        make_buffers=batch_buffer,
+        merge=batch_merge,
+    )
+
+    pass_results['var'] = pass_results['var'].data/pass_results['num_frame'].data
```
sk1p: Your return type is a bit inconsistent: for `var`, `std`, and `mean` you return `ndarray` instances; `num_frame` and `sum_frame`  are `BufferWrapper`s. You should decide on one here; I guess `ndarray` for all makes sense?

I discussed with @kruzaeva before that having a separate return type would be cleaner than returning `BufferWrapper`s; when we get there we can change the return type here.

tests/udf/test_simple_udf.py
```diff
@@ -2,11 +2,20 @@
 import pytest
 
 from libertem.common.buffers import BufferWrapper
-
+from libertem.api import Context
```
sk1p: I think this is unused?

src/libertem/udf/stddev.py
```diff
@@ -0,0 +1,189 @@
...
+        sum of variances, sum of pixels, and number of frames used
+
+    Returns
+    -------
+    collections.namedtuple object
+        Contains information about the merged partitions, including
+        sum of variances, sum of pixels, and number of frames used
+    """
+    if p0.N == 0:
+        return VariancePart(var=p1.var, sum_im=p1.sum_im, N=p1.N)
```
sk1p: Same as `return p1`?

src/libertem/udf/stddev.py
```diff
@@ -0,0 +1,189 @@
...
+        fn=compute_batch,
+        make_buffers=batch_buffer,
+        merge=batch_merge,
+    )
+
+    pass_results['var'] = pass_results['var'].data/pass_results['num_frame'].data
+    pass_results['std'] = np.sqrt(pass_results['var'].data)
+    pass_results['mean'] = pass_results['sum_frame'].data/pass_results['num_frame'].data
+
+    return (pass_results)
```
sk1p: Nitpick: parentheses not necessary here

src/libertem/udf/stddev.py
```diff
@@ -0,0 +1,189 @@
...
+    # compute sum of variances for joint samples
+    delta_P = mean_B - mean
+    var_AB = p0.var + p1.var + (p1.N * delta * delta_P)
+
+    return VariancePart(var=var_AB, sum_im=sum_im_AB, N=N)
+
+
+def run_stddev(ctx, dataset):
+    """
+    Compute sum of variances and sum of pixels from the given dataset
```
sk1p: One more thing: the paper used as a reference should be cited here

jaewshin: Is there a proper way to cite a paper in the code? Or can I just copy and paste the url for the pdf and mention the title of the paper?

src/libertem/udf/stddev.py
jaewshin: totally

src/libertem/udf/stddev.py
jaewshin: done

src/libertem/udf/stddev.py
jaewshin: fixed it!

src/libertem/udf/stddev.py
sk1p: Just DOI + title should be fine

tests/udf/test_simple_udf.py
jaewshin: deleted now

src/libertem/udf/stddev.py
jaewshin: I've made all of them into into narray instances now

sk1p:
Merged, thank you!

PR #309, opened 2019-03-26
Fixing the simple libertem API.ipynb

irahulcse:
I had fixed the Libertem API.ipynb but while doing it I had to replace the original file with my file, so pls check it while merging. By the way, I had fully tested it except the location for the raw file and some minor changes other things are as it is.


sk1p:
Thanks, this looks good! 

>  I had to replace the original file with my file, so pls check it while merging

Yes, this should be fine. I used [nbdime](https://nbdime.readthedocs.io/en/latest/) to compare the two versions, and it looks fine!

> By the way, I had fully tested it except the location for the raw file and some minor changes other things are as it is.

I also ran the notebook locally, and with your changes, it runs without errors. I'm going to go ahead and merge this now.

irahulcse:
Thanks, @sk1p

Issue #310, opened 2019-03-26
UDF: by-tile functions


sk1p:
Part of #286.

For some cases, this gives a very nice performance boost over doing things frame-by-frame; as observed by @uellue (and our initial research/prototyping).

This is mostly done, but may need some interface additions, see latest comments below.

<details>
  <summary>Outdated discussion, click to expand</summary>
Complication: with #307, we cannot easily create a view into the result buffer, as the frames in a tile don't have to be contiguous in navigation dimension any more. So we have multiple options here:

1) instead of `buf[:] = something`, we create a method `buf.set(something)` that fancy-indexes into the result buffer using a list of nav indices
2) implement the slicing interface for `buf[:]` and implement the indirection there. I don't know yet how this should be done in detail; need to experiment. Also, should we support indexing other than the "bulk assign" using `buf[:] = ...`? If the user decides to assign to `buf[i]` for some reason... 
3) don't use a view into the result buffer, make a temporary buffer for the frame result and copy into full result buffer using fancy indexing. may have significant overhead if you have a) many result buffers or b) larger results per navigation position

For consistency, I'm currently leaning towards either 2), or 1) with corresponding change to the by-frame UDFs.

If this is complete, we may want to port over our current analyses to this interface. This of course also requires integration into the Web API.
</details>

sk1p:
> If the user decides to assign to `buf[i]` for some reason...

I think this is actually possible to solve: we index into the coodinate list, and then use the resulting slice of coordinates as fancy index to do the actual assignment operation. We "just" need to make sure this works generally; operations other than assignment need to work, too.

sk1p:
I just had an idea for an alternative interface: why limit ourselves to "by tile", why not provide a generic "iterator of frames" interface? This may allow to express some problems more clearly, and keep state in local variables without doing buffer management. Basically, this would move parts of the innermost loop of `UDFTask` into the udf itself.

The object returned by the iterator could include a view into the result buffers as well (in this case, that object is just a tuple, but it doesn't have to be):

```
some_variable = 42
for frame, results in frame_iterator:
    some_variable += frame.reshape((-1,))[random.randint(0, frame.size)]
    results['some_navbuffer'][:] = some_variable
```
(excuse the stupid example)

This is just thinking out loud... are we losing any optimization opportunities by not providing contiguous tiles at all? Yes, because then we can't call `numpy` operations on whole tiles. So there is value in a by-tile interface, too.

[...]

PR #311, opened 2019-03-27
Example for performance effect of tiling numpy operations

uellue:
The difference is a result of balancing overheads from
fine-grained loops with gains from working on data that fits
the L3 cache.

Fancy indexing with the boolean mask creates a copy. If the
copy fits the L3 cache, the subsequent calculation of std()
can use the data from the L3 cache, which is much faster.

This example shows how we should implement per-tile UDFs #310 in
order to reap the benefits of this, in particular for operations
that are limited by memory transfer. std() is such an operation.

### DIFF
--- a/prototypes/map-reduce-comparison/tiled-performance.ipynb
+++ b/prototypes/map-reduce-comparison/tiled-performance.ipynb
@@ -1,0 +1,159 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import numpy as np"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "s = 64\n",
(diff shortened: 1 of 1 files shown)


sk1p:
Hm, what should we do about the link checker? In the current form, it is not that useful. Should we make this a manual process? :disappointed:

uellue:
> Hm, what should we do about the link checker? In the current form, it is not that useful. Should we make this a manual process? 😞

Good point. So far it generated more false positives than finding real broken links. That disturbs the CI process for code. Can we somehow only run it if there was a change in the docs folder?

I'm wondering what made it trip this time. Did GitHub include user agent filtering?

sk1p:
Maybe also some kind of trigger-happy DoS protection, no real way to find out. Only doing link checks on doc changes may work. Something like this:

```
if git diff $TRAVIS_COMMIT_RANGE --name-only | grep -q '^docs/'; then tox -e docs-linkcheck; fi
```

We still need to build docs on all CI runs, as changes to doc strings in the code also need to be reflected in the documentation.

Issue #47, opened 2018-07-16
Support for additional mask shapes


sk1p:
Currently we support ring and disk shaped masks and selecting single points. I'd like to collect ideas for additional mask shapes in this ticket, so feel free to comment here!

uellue:
by Ian MacLaren: Select set of spots in zone axis diffraction pattern.

sk1p:
@uellue that would be the bitwise AND of a number of disks, if I understand correctly? Or does it also involve some kind of automatism for determination of center and radius?

uellue:
@sk1p In the most simple case that is just a bunch of disks. But since zone axis diffraction patterns are quite regular, one could implement helpers to create and position them. I'm imagining sth like a snapping grid for vector graphics that you select based on the crystal symmetry, and then stretch and rotate to match the diffraction pattern. Then functions to adjust the size of several disks at one and/or populate some sets of grid points with copies/clones?

In order to not get into feature creep, one should probably start by just allow adding an arbitrary number of disks (resp. other shapes that are added) for now. Then we see what users do with that. :-) By the way, that intersects with Karina's work because her project is to detect shifts of whatever is at these points.

uellue:
#49 for a mask suggested by Rafal

uellue:
#89 implements an example to select sets of spots and separate them from diffuse background with the Python API

uellue:
Having at least four channels ("colors") available for masks would be useful. That way one could draw masks for quadrant detectors and look what they look like, or compare sets of reflections.

uellue:
Polygon and spline shapes should be possible. As an example, draw the quadrants of a four-quadrant detector, i.e. ring sections, or other irregular masks that select specific features.

uellue:
Assign positive or negative weights to regions. Ideally, allow all values between -1 and 1.

sk1p:
This was again requested by a user, maybe a simple version of this could be implemented in 0.7

uellue:
Closing, dated.

PR #318, opened 2019-03-30
Trying to create a new mask named as Square.

irahulcse:
Square Mask is the new kind of mask which could be used to as a new mask for #47. This mask would have the same attribute like the disk mask but we will generate the image considering the side of the mask.


sk1p:
Sorry the response took so long. While this is certainly a good exercise in adding a new `Analysis`, I'm afraid we can't accept this PR. Having a square mask shape as a stand-alone `Analysis`, sadly, is not all that useful. A square would be a special case of a polygon, which will be part of the mask editor project you selected for GSOC. In the future, it would be best to discuss changes like this (either in gitter or in an issue) **before** opening a PR. Working in isolation will mean you waste your time on code that may ultimately be rejected. The general structure of your `Analysis` implementation is good, though!

Nevertheless, I can give you feedback on your changes:

- Please don't include unrelated changes in your PR. In this case, you include changes in the example notebook (without adding a square example, as far as I can see), and some stylistic changes in `ring.py` (which actually partly violate our style guides, see below). This makes it harder to review for us project maintainers
- Include test cases. Tests are small, mostly self-contained, functions that exercise your code and try to verify that it actually does what you expect it to do. See [this file](https://github.com/LiberTEM/LiberTEM/blob/9aa3086fd7bcdec8e9fa054a851dbf6b502a7b06/tests/test_analysis_shapes.py#L17) for examples, and [our contribution docs](https://libertem.github.io/LiberTEM/contributing.html) on how to run tests locally. Tests are also automatically run with each pull request in a process called [continuous integration](https://en.wikipedia.org/wiki/Continuous_integration).
- Also as part of the continuous integration, the code style is checked. See also the section in code style in [our contributing docs](https://libertem.github.io/LiberTEM/contributing.html#code-style); you can locally run the style checker. The results of the CI run can be checked on this pull request page (look for "checks"). Here is a direct link to the style check results: https://travis-ci.org/LiberTEM/LiberTEM/jobs/513438450 - as you can see, it has some "opinions" on your code :grin: 
- one of the messages there is `src/libertem/analysis/square.py:39:20: F821 undefined name 'mask'` - this means, you are trying to access a variable that you did not define before. That also means, you didn't even manually test your code. **Please** only submit PRs with changes that are tested, if not by automated testing, **at least** run your code locally with a small test notebook, or interactively in a Python REPL. If you need help running your code, **ask**!
- in the just-mentioned part of the code, you are calling `mask.square` — I guess you meant `masks.square`. This function [does not exist yet](https://github.com/LiberTEM/LiberTEM/blob/master/src/libertem/masks.py); maybe you forgot to `git add` your changes in that file?

Much of this is covered by [our contributing documentation](https://libertem.github.io/LiberTEM/contributing.html#contributing) - **please** read it again, and let me know if any part of it is unclear.

PR #319, opened 2019-04-03
Prototype for binning of the data

woozey:
Prototype showing how to do binning in both navigation and diffraction spaces.

Issue remains for tiles (sizes) which do not fully fit inside the dataset, see section 1.3.

Binning assumes arbitrary size of the dataset and equal binning factors in both directions. Tested for 2d scans (4d) only and for data with a single partition.


sk1p:
Thanks, merged!

> Issue remains for tiles (sizes) which do not fully fit inside the dataset, see section 1.3.

This is a bug in the mib reader, which should be fixed in the [ds-refactor](https://github.com/LiberTEM/LiberTEM/tree/ds-refactor) branch.

PR #321, opened 2019-04-04
Documentation for FEM UDF was added, closes #317

### LINKED ISSUE #317: FEM udf should have some documentation
At least the `run_fem` function should have a docstring which describes the individual parameters and their types. Maybe also add a simple example how the function should be called.

@kruzaeva could you take care of this?

See [this documentation](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#google-vs-numpy) for how the docstring should be formatted, specifically we are using the numpy style for docstrings.

kruzaeva:


sk1p:
Thanks, merged!

Issue #227, opened 2018-12-21
Numerical precision for high-resolution detectors


uellue:
The test for numerical precision added in #226 uncovers that float32 may not be enough for 4096x4096 detectors to give sufficient numerical precision with standard reduce operations. We might have to use better algorithms like [Kahan summation](https://en.wikipedia.org/wiki/Kahan_summation_algorithm) or calculate with float64. 

For now, the test is "cheated" by using 1.0 as value, which doesn't give numerical errors.

sk1p:
Relevant links about error boundaries in BLAS:

https://www.netlib.org/lapack/lug/node108.html
https://software.intel.com/en-us/mkl-developer-reference-c-error-analysis

ju-bar:
I propose to use butterfly summation. My tests show that it provides accuracy as well as better speed compared to straight summation.
Examples: fdncs2 (preserves array data) or fdncs2m (modifies array data)
in https://github.com/ju-bar/JMultiSliceLib/blob/master/src/JMultiSlice.cpp

ju-bar:
I've used a small test program 
https://github.com/ju-bar/TestSum
for checking different summation algorithms concerning accuracy and speed. Here is the result:
https://github.com/ju-bar/TestSum/blob/master/TestSum/output.txt

uellue:
Hm, interesting! Thank you for testing this @ju-bar.

Since we use the summation only indirectly as part of the numpy or torch implementation of the dot product, it would have to be changed there. I'm a bit curious how that is implemented there and if the BLAS implementation or torch could actually be improved with a better sum.

Discussions with @sk1p: The most pragmatic solution for LiberTEM is probably passing an option to the dataset to convert the values to float32 or float64 (resp. complex64 or complex128) for users to decide between speed and accuracy. The dataset is the best place because, for example, K2IS has to be unpacked anyway and we can unpack right into the desired type, eliminating a copy step. Going to float64 rather than changing the sum algorithm will improve the accuracy of all calculations in all libraries, not only the sum in specific places.

sk1p:
> The most pragmatic solution for LiberTEM is probably passing an option to the dataset to convert the values to float32 or float64 (resp. complex64 or complex128) for users to decide between speed and accuracy. 

[...]

PR #322, opened 2019-04-08
Dataset refactor

### LINKED ISSUE #245: Binary reader: better partitioning
Currently, support for nD datasets in the raw reader is not optimal: to use it for a line scan of 800 frames, for example, you have to specify a shape like `(1, 800, 128, 128)`. This does not play well with the simple partitioning scheme currently implemented, which only splits in the y direction. In effect, the whole line will end up in one partition and no parallelism will be achieved at all.

This issue goes hand-in-hand with #207, as removing the EMPAD-specific hacks will allow us to simplify the raw dataset a bit.

### LINKED ISSUE #252: Cleanup DataSet implementations
- [x] abstract out some helper functions or methods in the base DataSet or helper classes.
- [x] all `Partition`s and `DataTile`s should have flat navigation (signal can still be ND).
- [x] `get_tiles` needs to support some more parameters (`mmap`, `full_frames`, `dest_dtype`). 
- [x] tests for the new EMPAD reader
- [x] better partition sizing (needs total number of cores from executor!)
- [x] finish I/O capability system
  - [x] make capabilities available to `DataSet`, not only `FileSet3D` impls (maybe put them into `DataSetMeta` and make it a required part of `FileSet3D`?)
  - [x] capabilities can be dynamic (for example, they can depend on parameters or format details like pixel types (MIB 1bit can't be mapped as bool ndarray, other types can!))

Status: all `DataSet` impls should be on feature-parity now, tests pass.

### LINKED ISSUE #260: MIB: support for 1-bit dtype
Related: #129

### LINKED ISSUE #270: Bad performance of MIB datasets reading
Processing time for MIB dataset (8 GB, path="/cachedata/users/ruzaeva/34/01_ms1_10K.mib"), which is not split into small datasets, takes too long.
The reason is, probably, that data is not divided into partitions.

sk1p:
This is the current status of the fix for #252. Many readers now share a lot of code, and are in general more uniform. The new code is not yet fully featured, see #252 for details (most important IMHO: support for reading parts of frames, either for cropping or for tiling).

Fixes #252, fixes #270, fixes #260, fixes #245 

Some changes may require further discussion. 

~~For example, I included a new `EMPADDataSet` that sets all parameters automatically (from the .xml file). Currently the XML file is required; maybe this should be changed to allow manual entry of parameters if there is no XML file~~ This is now done, including support in the user interface.

I merged the raw and raw_direct readers; the new reader can be set into direct reading mode or "normal" mode. This new reader no longer supports cropping the detector, as this was a special case (and the cropping was very limited, for example it had no way to change the origin, it always cropped from 0).

@uellue for some reason, the `test_numerics` testcase started failing in development, now appears to be fine again. Can you think of any reason for instability there?


uellue:
@sk1p the `test_numerics` test only passes because the chosen value `VAL = 1.0` doesn't accumulate errors. As far as I understand the matter, the float representation is exact and doesn't use many bits. If you run it with other values, it starts building up numerical errors and fails.

I'd suspect that this test might be sensitive to changes in the exact way how `numpy` or `pytorch` execute the operations. Perhaps it can produce spurious errors for that reason?

I've written it as a prototype to verify that a solution to #227 works, once we have that implemented. See also #226! As soon as we have a "high precision mode", we can verify that it works with other values for `VAL`.

PR #332, opened 2019-05-07
 Add strain mapping with affine transformation of peak positions to reference

uellue:
Inspired by Giulio Guzzinati
https://arxiv.org/abs/1902.06979


uellue:
As a side note, there are vast differences between the various optimizers available in `scipy.optimize.minimize()`. The default one seems to give the most accurate and robust results, but takes comparatively many iterations to converge. Further tests might be worthwhile.

PR #334, opened 2019-05-13
Move full frame refinement code to prototypes folder

uellue:
Large deviations are usually wrong for strain mapping, that means we
should constrain the optimizer tightly.


uellue:
Regarding test coverage, this would work best with real data.

uellue:
Since this code didn't show a clear advantage over the correlation-based blobfinder, but could be interesting for future developments, I propose to move it to the prototypes folder for now.

sk1p:
Thanks, merged!

Issue #197, opened 2018-12-11
Use sparse.pydata.org instead of scipy.sparse


uellue:
Currently, LiberTEM uses [scipy.sparse](https://docs.scipy.org/doc/scipy/reference/sparse.html) for sparse matrix multiplication. 

However, the alternative implementations from [torch](https://pytorch.org/docs/stable/sparse.html) and [pydata](http://sparse.pydata.org/en/latest/) are available as well.

We should compare the implementations with regards to performance, stability and ease of use, and decide if we should stick with scipy.sparse, pick something else or even support several implementations like we do with dense matrices already.

See also #182

uellue:
https://sparse.pydata.org/ is coming out ahead by a large margin in most cases:

```python
import numpy as np
import scipy.sparse as sp
import sparse
import torch

# 1 MiB tile size
size = 1024*1024*1
​
(fy, fx) = (512, 512)
​
l = size // (fy*fx)
​
print(l)
# 4
​
data = np.random.random((l, fy*fx))
​
n_masks = 100
n_entries = 25
​
masks = np.zeros((n_masks, (fy*fx)))
​
mask_indices = []
frame_indices = []
values = []
​
for i in range(100):
    for y in np.random.choice(range(fy), size=n_entries, replace=False):
        for x in np.random.choice(range(fx), size=n_entries, replace=False):
            index = y*fx + x
            mask_indices.append(i)
            frame_indices.append(index)
            masks[i, index] = 1
            values.append(1)

np.dot(data, masks.T)

%timeit np.dot(data, masks.T)
# 51.6 ms ± 963 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

t_data = torch.from_numpy(data)
t_masks = torch.from_numpy(masks.T)
torch.mm(t_data, t_masks)

%timeit torch.mm(t_data, t_masks)
# 39.2 ms ± 1.95 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

sp_masks = sp.csr_matrix(masks)
sp_masks.dot(data.T)

%timeit sp_masks.dot(data.T)
# 13.4 ms ± 419 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

i = torch.LongTensor([mask_indices, frame_indices])
v = torch.DoubleTensor(values)
t_sp_mask = torch.sparse.DoubleTensor(i, v, torch.Size([n_masks, fy*fx]))
[...]

PR #336, opened 2019-05-14
Use sparse.pydata.org

uellue:
See #197 for discussion.

The test in the notebook confirms the advantage even if we first generate dense masks and then convert to sparse.


uellue:
@sk1p From my point of view this should be mergeable now. What do you think?

uellue:
Closing and reopening since CI seems to be stuck.

sk1p:
Thanks, merged! I wonder WTF was going on with CI there, there were some of these `CancelledError` exceptions again. Maybe because of the new `distributed` version?

PR #339, opened 2019-05-16
Allow creating a stack of masks with one factory function

### LINKED ISSUE #338: Generate a deep mask stack with one function instead of list of functions
Currently, mask stacks are assembled one by one. This works well for scenarios with few masks, such as shape or COM analysis.

For other purposes, such as radial integration, non-trivial binning or template matching, a deep stack of masks with thousands or tens of thousand masks could be useful. Tests for #197 have shown that this can be efficient if the matrices are sparse.

For that reason, we should accept functions that generate the whole stack at once instead of slices.

Possibilities for implementation:
1. Accept both list of functions or single function for factories argument. Advantage: minimal impact. Disadvantage: Magic polymorphism, not obvious
2. Two kwargs, factories or factory, of which one is set. Advantage: Explicitly clear which is what. Disadvantage: Two invalid situations -- bot set or both not set
3. New subclass of mask analysis. Advantage: Clear and explicit. Disadvantage: Significant code duplication and growing API surface.

I am tending towards the first one. What do you think?

uellue:
Still to do:
- [x] Update perforamnce test notebook from sparse
- [x] Update or implement helper functions in libertem.masks

Closes #338


sk1p:
Hm, CI is stuck again, can you maybe try to pin the dask distributed version to 0.28, or maybe 0.27.1 to see if it works more reliable?

uellue:
Hard to reproduce, it seems. :-/ @sk1p Should I leave it unpinned or would you like to run more tests?

uellue:
I've reset the branch to exclude the CI tinkering commits. CI passed with dask.distributed<=1.27.1, <=1.28 and without restriction.

src/libertem/job/masks.py
```diff
@@ -116,6 +136,16 @@ def _compute_masks(self):
         # If it is None, use sparse only if all masks are sparse
         # and set the use_sparse property accordingly
 
+        if not isinstance(self.mask_factories, Iterable):
```
sk1p: Can we somehow not repeat the conditionals here? I guess the only real difference is the last case, where we may have "mixed-sparsity" in the list of masks?

sk1p:
Looks good; merged, thanks!

PR #342, opened 2019-05-21
add extra_shape param to single type buffer

jaewshin:
Added extra_shape parameter for kind="single" buffer so that users can arbitrarily set the dimension of the buffer beyond 1 dimension.


sk1p:
Looks good! If you are ready, click the "ready for review" button and I'll merge it.

sk1p:
Thanks, merged!

PR #345, opened 2019-05-23
Issue 80 - Fix to allow container deployment

ozej8y:
This fix allows the user to set --local_directory to a folder outside the container.
This then allows Dask to write files for a container deployment.

e.g.
libertem-server --local_directory=$home/dask-user-space

### DIFF
--- a/src/libertem/web/cli.py
+++ b/src/libertem/web/cli.py
@@ -5,7 +5,7 @@ import os
 @click.command()
 @click.option('--port', help='port on which the server should listen on',
               default=9000, type=int)
-@click.option('--local_directory', help='local directory to manage dask-worker-space files',
+@click.option('--local-directory', help='local directory to manage dask-worker-space files',
               default='dask-worker-space', type=str)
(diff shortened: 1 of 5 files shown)


src/libertem/web/cli.py
```diff
@@ -5,17 +5,19 @@
 @click.command()
 @click.option('--port', help='port on which the server should listen on',
               default=9000, type=int)
+@click.option('--local_directory', help='local directory to manage dask-worker-space files',
```
sk1p: Nitpick: I would prefer a dash here instead of the underscore, following the usual unix conventions.

uellue:
Hi @ozej8y, thank you for contributing and welcome to LiberTEM! 👍  :-)

src/libertem/web/cli.py
ozej8y: Updated code to replace the dash.

sk1p:
Thanks, merged!

PR #347, opened 2019-05-29
pca skeleton

jaewshin:
- Uses incremental PCA algorithm (Set it to produce deterministic output. However, the result might not be the same as using original PCA on the entire data)
- Uses class-based UDF approach


src/libertem/udf/pca.py
```diff
@@ -0,0 +1,217 @@
...
+import numpy as np
+
+from libertem.udf import UDF
+from libertem.udf.stddev import merge
+
+
+pca = collections.namedtuple('PCA', ['sum_im', 'var', 'N',  'n_components', 'singular_vals'])
+
+
+def flip_svd(U, V):
```
sk1p: This function is adapted from `sklearn`, right? Should be properly credited if this is the case (it is completely fine, as long as we comply with [their license](https://github.com/scikit-learn/scikit-learn/blob/master/COPYING)). Also, first part of the docstring in sklearn is kind of important: `"Sign correction to ensure deterministic output from SVD"` - it helps to understand the "why" of the function.

src/libertem/udf/pca.py
```diff
@@ -0,0 +1,217 @@
...
+            N=self.results.num_frame,
+            n_components=self.results.n_components,
+            singular_values=self.results.singular_vals
+            )
+
+        sig_dim = self.results.singular_vals.shape
+        num_frame = tile.shape[0]
+
+        new = pca(
+            sum_im=np.sum(tile),
```
sk1p: You may need to specify `axis=0` here, right? Otherwise, the whole tile is reduced to a scalar value

src/libertem/udf/pca.py
```diff
@@ -0,0 +1,217 @@
...
+    """
+    max_abs_rows = np.argmax(np.abs(V), axis=1)
+    signs = np.sign(V[range(V.shape[0]), max_abs_rows])
+    U *= signs
+    V *= signs[:, np.newaxis]
+
+    return U, V
+
+
+def IncrementalPCA(prev_result, data):
```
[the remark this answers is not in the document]
jaewshin: Ah sorry for forgetting this again. Will make sure of this next time.

src/libertem/udf/pca.py
```diff
@@ -0,0 +1,217 @@
...
+            'n_components': self.buffer(
+                kind='single', dtype='int32'
+                ),
+            'mean': self.buffer(
+                kind='sig', dtype='float32'
+                ),
+            'var': self.buffer(
+                kind='sig', dtype='float32'
+                ),
+            'num_frame': self.buffer(
```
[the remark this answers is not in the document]
jaewshin: Sorry I'm not sure if I correctly understood your comment here. Do you mean that when you set full_frame=False, which are "stacks of cropped regions," we need independent values for kind='single'? If so, why is that necessary? Cropping dimension would still be the same under full_frame=False and we would have the same cropping dimension for all partial frames in the tile right? Also, if we were to use kind='single', then we would have to specify the dimension through extra_shape parameter, correct? If so, how do we access the cropping dimension from bufferwrapper object?

src/libertem/udf/pca.py
```diff
@@ -0,0 +1,217 @@
...
+            the merged partitions, including mean, variances,
+            number of frames used, singular values for partitions, and
+            the number of components
+        """
+        prev = pca(
+                    var=dest['var'][:],
+                    mean=dest['mean'][:],
+                    num_frame=dest['num_frame'][:],
+                    singular_val=dest['singular_vals'][:],
+                    n_components=dest['n_compoennts'][:]
```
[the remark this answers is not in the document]
jaewshin: Sorry fixed it

src/libertem/udf/pca.py
```diff
@@ -0,0 +1,217 @@
+import collections
+
+import numpy as np
+
+from libertem.udf import UDF
+from libertem.udf.stddev import merge
+
+
+pca = collections.namedtuple('PCA', ['sum_im', 'var', 'N',  'n_components', 'singular_vals'])
```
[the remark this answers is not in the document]
jaewshin: So for namedtuple, the main issue as I understood it is the performance, right? (I've read that using namedtuple is slower than simple class) I have yet to find a clear, concise solution to this (most had multiple lines of code). If inheritance is what you care about, I think class would be the easiest option. There was a way to mix up namedtuple and class properties using typing.namedTuple(python 3.6 +), which allows multiple inheritances etc as well as using it the regular namedtuple way: https://stackoverflow.com/questions/44320382/subclassing-python-namedtuple. But then again, if performance is what you care about, I don't think this fundamentally resolves the issue. Let me know what you think about this!

src/libertem/udf/pca.py
```diff
@@ -0,0 +1,217 @@
+import collections
+
+import numpy as np
+
+from libertem.udf import UDF
+from libertem.udf.stddev import merge
```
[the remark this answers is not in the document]
jaewshin: Fixed it!

src/libertem/udf/pca.py
```diff
@@ -0,0 +1,217 @@
...
+        ----------
+        dest
+            Contains information about the first partition, including
+            sum of variances, sum of pixels, and number of frames used
+
+        src
+            Contains information about the second partition, including
+            sum of variances, sum of pixels, and number of frames used
+
+        pca
```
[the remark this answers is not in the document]
jaewshin: I'm actually not sure why I included it in here.. I've taken it out

src/libertem/udf/pca.py
jaewshin: Indeed yes. For citation, there is no "official" way of citing a code directly, other than in the form of a bibliography. Also, I couldn't find doi for scikit learn anywhere... Would providing a link to the specific code file (on github) be suffice?

src/libertem/udf/pca.py
jaewshin: Yes totally

jaewshin:
Thanks a lot for your review! As for your general comment about incremental PCA, I've added the citations for papers I looked into. As for testing, I've been struggling in the past few days on constructing a test case. I will let you know if I succeed in making one or if I get too stuck and not make much progress.

src/libertem/udf/pca.py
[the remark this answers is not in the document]
jaewshin: Furthermore, if full_frame=False is indeed taking partial frames, I'm not sure if we can use this method for dimensionality reduction. Computing some statistic (related to dimension reduction) on a subset of the image and then merging it doesn't quite make sense to me...

src/libertem/udf/pca.py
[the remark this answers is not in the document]
sk1p: Performance should be fine, as there should only be O(1) calls to namedtuple constructors per frame, and not O(number_of_pixels) or something like that.

Actually, my concern was more readability, as there are some parts that mostly just pack things into a namedtuple to pass it to a function. If this is already the most readable, I'm fine with it :+1:

src/libertem/udf/pca.py
sk1p: > For citation, there is no "official" way of citing a code directly
> [...] Would providing a link to the specific code file (on github) be suffice?

I _personally_ think a link should be fine (one could argue about the "threshold of originality" for copying 4 lines of code, but I'm not sure we want to go there :grin:) but the license says:

```
[...]
Redistribution and use in source and binary forms, with or without
modification, are permitted provided that the following conditions are met:

  a. Redistributions of source code must retain the above copyright notice,
     this list of conditions and the following disclaimer.
[...]
```

So, following the license to the letter, we need to reproduce the license in full somewhere. A good example of this, actually in sklearn: https://github.com/scikit-learn/scikit-learn/blob/a50c03f975c4bbf306c5af1826f46ba5dc5230a0/sklearn/utils/_pprint.py - in that case, the whole file is from an external project. Maybe it makes sense to keep code you borrow/adapt from sklearn into a separate file, which has [the whole license](https://github.com/scikit-learn/scikit-learn/blob/master/COPYING) as a comment at the top. What do you think?

jaewshin: Yeah that shouldn't be a problem at all. I've been writing code for NNMF on the side, and I was looking into sklearn code for reference as well (sklearn contained some cython and extra library stuff for nnmf so I didn't copy any yet except for maybe the general idea). If in the future such an issue of copying arises again in the future, having a separate file would ease the burden of citation

jaewshin:
Initial literature review on dimension reduction is referenced on Issue #363 
+ Implementation for PCA
+ Testing for PCA
+ Documentation for PCA
+ Jupyter notebook for testing PCA
For now, removed codes on NMF.

docs/source/app/pca.rst
```diff
@@ -0,0 +1,30 @@
...
+
+How PCA works
+-------------
+
+.. image:: ./images/pca_diagram.png
+
+As mentioned above, PCA heavily exploits the distributed nature of the UDF architecutre. First, using the partitioned data from the partitioning stage of UDF, we perform standard PCA on each of the partitions. Note that the current implementation utilizes :meth:`~libertem.udf.UDFFrameMixin.process_partition` instead of :meth:`~libertem.udf.UDFPartitionMixin.process_frame` to provide additional boost in the performance of the algorithm. Then, for each partition, we store the values for the loading matrix (matrix multiplication of the left singular matrix with the singular values), and the component matrix which is of our primary interest since the component matrix serves as a projection matrix into the lower-dimensional space and thereby achieving dimension reduction.
```
uellue: I believe process_frame and process_partition are mixed up?

docs/source/app/pca.rst
```diff
@@ -0,0 +1,30 @@
...
+
+Once PCA is independently applied on all partitions, we proceed to `merge` these results of independent PCA results, which is done rather naturally by implementing the :meth:`~libertem.udf.UDF.merge` method in UDF class. More specifically, given the results of the PCA on each partition, we first reconstruct the original matrix using the loading and the component matrices, and then reducing the dimension of the data matrix via `hyperbox` method. `Hyperbox` method is a way to reduce the dimension of the loading matrix by finding the maximal `hyperbox` that contains the span of the vectors in the loading matrix. In doing so, we can find a simpler description for the loading matrix, which then reduces the dimension of the data, thereby leading to an increase in the overall performance. At each iteration, once `hyperbox` method is performed for the partition, we use Incremental PCA algorithm (IPCA) to incrementally update the component matrix by fitting the post-`hyperbox` data matrix. One major advantage of IPCA is that it can easily handle large-size data that standard PCA algorithm cannot, and UDF architecture served as a basis for implementing Incremental PCA.
```
uellue: Isn't it the other way around, first we reduce and then we reconstruct a reduced matrix? In general, you should make clear that we are looking for the *components* here, i.e. we are not creating a description for the loading matrix, but a matrix that is much smaller but yields the same components when doing PCA.

Nice use of incremental PCA for the merging, btw! :-)

For me it would be very important to describe precisely how the hyperbox method is validated and what its limits are. Did you work on that?

jaewshin: Yes you are totally right. I got mixed up in the wording of this. I will fix it and make it more clear that the component matrix is what we are looking for. As for validating the hyperbox method, I did try few things, which is on pca.ipynb under example directory, but couldn't really find a case where the hyperbox method failed, which worries me since we don't know when and where this method fails... If you have any additional ideas on how to verify this, please let me know

PR #348, opened 2019-05-31
UDF for crystallinity map calculation

kruzaeva:
Returns crystallinity map.
Calculates coefficients for each frame: low values for frames (only zero order peak) which belong to amorphous regions, high values for crystalline frames (zero order peak+periodic non zero order peaks)

### DIFF
--- a/src/libertem/udf/crystallinity.py
+++ b/src/libertem/udf/crystallinity.py
@@ -38,10 +38,11 @@ def init_fft(partition, rad_in, rad_out, real_center, real_rad):
 
 
 def fft(frame, real_mask, fourier_mask, intensity):
-    if not (real_mask is None):
-        intensity[:] = np.sum(np.fft.fftshift(abs(np.fft.fft2(frame*real_mask)))*fourier_mask)
+    if real_mask is not None:
+        maskedframe = frame*real_mask
     else:
-        intensity[:] = np.sum(np.fft.fftshift(abs(np.fft.fft2(frame)))*fourier_mask)
+        maskedframe = 1*frame
+    intensity[:] = np.sum(np.fft.fftshift(abs(np.fft.fft2(maskedframe)))*fourier_mask)
     return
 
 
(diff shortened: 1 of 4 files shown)


src/libertem/udf/crystallinity.py
```diff
@@ -0,0 +1,84 @@
...
+        Inner radius of a ring mask for the integration in Fourier space
+
+    rad_out: int
+        Outer radius of a ring mask for the integration in Fourier space
+
+    real_center: tuple
+        (x,y) - coordinates of a center of a circle for a masking out zero-order peak in real space
+
+    real_rad: int
+        (x,y) - radius of circle for a masking out zero-order peak in real space
```
sk1p: I think `(x,y)` for `real_rad` is not quite right...

Also, is `(x,y)` correct for `real_center`? In the `init_fft` function above, `_make_circular_mask` is called with `real_center[1], real_center[0]`. Compare this with the [signature of `_make_circular_mask`](https://github.com/LiberTEM/LiberTEM/blob/a3f54b1e531a248d74f0fe381b6df0068531c398/src/libertem/masks.py#L6) which has the parameters ordered as `centerX, centerY` (for historical reasons, really...).

src/libertem/udf/crystallinity.py
```diff
@@ -0,0 +1,84 @@
...
+
+    kwargs = {
+        'real_mask': real_mask,
+        'fourier_mask': fourier_mask,
+    }
+    return kwargs
+
+
+def fft(frame, real_mask, fourier_mask, intensity):
+    intensity[:] = np.sum(np.fft.fftshift(abs(np.fft.fft2(frame*real_mask)))*fourier_mask)
```
sk1p: What do you think about using `pyfftw` here, if available?

As a note, it almost looks like we need to provide our own fft "abstraction" (very thin layer...), for example as `libertem.fft` and centralize the choice of fft backend. That is kind of stupid, but saves us from cluttering each FFT using module with the "optional import" pattern [as seen in the blobfinder UDF](https://github.com/LiberTEM/LiberTEM/blob/a3f54b1e531a248d74f0fe381b6df0068531c398/src/libertem/udf/blobfinder.py#L15-L28)...

Either that, or we use a single FFT impl. throughout. Anyways, the decision should not hold up merging this PR.

/cc @uellue

sk1p:
Thanks, merged!

PR #349, opened 2019-06-03
dtype support

### LINKED ISSUE #227: Numerical precision for high-resolution detectors

uellue:
Tools and automatisms to control the dtype in which data is read and calculations are performed. All dtype combinations that support numpy.dot() should work now, including integer and complex number types for masks and/or data.

Closes #227 by allow switching to float64 or complex128.

Keep sparse.pydata.org for most of the sparse mask handling, but switch back to scipy.sparse for processing with the correct data layout, which is faster.

Drive-by change: Additional mask generation code and moving utility functions around.

### DIFF
--- a/tests/udf/test_blobfinder.py
+++ b/tests/udf/test_blobfinder.py
@@ -1,4 +1,5 @@
 import numpy as np
+import pytest
 
 import libertem.udf.blobfinder as blobfinder
 
@@ -30,10 +31,12 @@ def test_refinenemt():
     assert (rx < x) and (rx > (x - 1))
 
 
+@pytest.mark.xfail(reason="FIXME update blobfinder UDF to new UDF interface")
 def test_smoke(lt_ctx):
     """
     just check if the analysis runs without throwing exceptions:
     """
+    pytest.sk
     data = _mk_random(size=(16 * 16, 16, 16), dtype="float32")
     dataset = MemoryDataSet(data=data, tileshape=(1, 16, 16),
                             num_partitions=2, sig_dims=2)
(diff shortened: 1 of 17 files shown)


tests/udf/test_blobfinder.py
```diff
@@ -30,10 +31,12 @@ def test_refinenemt():
     assert (rx < x) and (rx > (x - 1))
 
 
+@pytest.mark.xfail(reason="FIXME update blobfinder UDF to new UDF interface")
```
sk1p: This needs to be removed; it is the cause for the decrease in coverage. I think you can just revert the whole file to the state it is in `master` and things should be fine.

sk1p:
Other than the stray changes in the blobfinder test, this looks good!

tests/udf/test_blobfinder.py
uellue: Done! :-)

sk1p:
Could you rebase to current master? I'm now pinning `dask.distributed` to <1.28 to see if this helps with our CI hangs. Locally I couldn't reproduce the hangs with `1.27.1`, where it hung up quite consistently on `1.28`.

uellue:
> Could you rebase to current master? I'm now pinning `dask.distributed` to <1.28 to see if this helps with our CI hangs. Locally I couldn't reproduce the hangs with `1.27.1`, where it hung up quite consistently on `1.28`.

Done!

sk1p:
Looks good, thanks!

PR #352, opened 2019-06-05
Port the blobfinder and logsum to the new UDF interface

uellue:
* Porting
* Add docstrings
* Add more tests
* Use multiple inheritance for UDF composition in blobfinder


sk1p:
Looking good, thanks!

PR #353, opened 2019-06-05
Fix EMPAD bug; resolve performance regression

uellue:
1. Fix integer overflow in EMPAD reader

2. Switching the sparse matrix back-end has degraded performance for shallow mask stacks, like for example for the various shape analyses. Since sparse matrices are critical for deep mask stacks where the tests showed good performance (to be checked again, perhaps), disable sparse matrices for the default analyses for now and rely on dense matrices that show good, predictable performance in this scenario.


sk1p:
Thanks, merged!

PR #354, opened 2019-06-06
Postprocessing interface for UDFs and sparse matrix based correlation

uellue:
This allows to perform postprocessing for completed frame data
in combination with process_tile()

This is targeted at sparse matrix based correlation that is expected to benefit immensely from
processing tiles instead of frames in combination with refinement.


sk1p:
Looks good, thanks!

PR #356, opened 2019-06-06
Radial Fourier analysis

uellue:
This is based on PR #354 since it relies on switching sparse back-ends.

* Back-end and API
* Web GUI -- thx @sk1p!
* Optimizations for large frames -- depending on the parameters, sparse or dense can be better. This is relevant for the GUI where it was easily possible to enter parameters that will eat up a lot of RAM and CPU time.


uellue:
The coverage information seems to be in disarray...

client/src/messages.ts
```diff
@@ -304,8 +315,13 @@ export interface PickFrameDetails {
     parameters: PickFrameParams,
 }
 
+export interface RadialFourierDetails {
+    type: AnalysisTypes.RADIAL_FOURIER,
+    parameters: RadialFourierParams,
+}
+
 export type AnalysisParameters = MaskDefRing | MaskDefDisk | CenterOfMassParams | PointDef | PickFrameParams;
```
sk1p: Hmm, we may need `RadialFourierParams` as part of `AnalysisParameters`, too. Not sure why it is not complaining, because the `updateParameters` action uses this type...

uellue:
+1 for documentation! Same goes for the Blobfinder. I'll work on it tomorrow. :-)

sk1p:
Going to merge this, despite WIP label, as it is easier to collaborate on documentation this way.

PR #358, opened 2019-06-10
added more for udf documentation

jaewshin:
Added some explanation for buffers and try to explain udf from beginner's perspective


docs/source/udf.rst
```diff
@@ -7,6 +7,42 @@ of LiberTEM allows users to run their own reduction functions easily, without ha
...
+How UDF works in layman's terms
+-----------------------------------
+
+.. image:: ./images/diagram.svg
+
+The UDF interface of LiberTEM is heavily utilizing the existing LiberTEM architecture. First,
+data is partitioned into several `"partitions"` and distributed across workers. Then, each partition,
+which can be viewed as a collection of frames, are processed by the user-defined `process_frame` function.
+Here, frames are fetched in an iterative manner, and one could view `process_frame` as `merging` the previous
+computations (i.e., the result computed on previously considered set of frames) and the newly added frame.
```
sk1p: Hmm, it kind of depends on the operation. If you only have `kind="nav"` buffers, you can view `process_frame` as working independently on different frames. For `kind="sig"`, this merge-inside-`process_frame` description is spot on, though!

Also, this "merging" inside of `process_frame` needs to be differentiated from the `merge` method described below: `process_frame` only has access to the results from inside the current partition, whereas `merge` always merges the results from a single partition into the full result. (maybe  "previously considered set of frames of a partition" or something like this?)

docs/source/udf.rst
```diff
@@ -7,6 +7,42 @@ of LiberTEM allows users to run their own reduction functions easily, without ha
...
+After all the frames in a partition are processed, LiberTEM `merges` the results from each worker, which is called
+by the parameter `merge` in UDF class. Note that in both `process_frame` and `merge`, buffers store the intermediate
```
sk1p: > the parameter `merge` 

Do you mean "method" here?

docs/source/images/diagram.svg
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
sk1p: Over/under "Output 1" there is a stray "z" somehow

FYI this SVG only seems to work fine in browsers; it doesn't load for example in inkscape

docs/source/udf.rst
```diff
@@ -7,6 +7,42 @@ of LiberTEM allows users to run their own reduction functions easily, without ha
...
+outcomes of the user-defined operations.
+
+Initializing Buffers
+--------------------
+In the UDF interface of LiberTEM, buffers are the tools to save and pass on the
```
sk1p: Maybe this should also mention somewhere the concrete case: that when accessing a buffer via `self.results.buffername`, you get back a view into a numpy array, and that this view will correspond to the current unit of data you are working on (frame/tile/partition/...)

docs/source/udf/buffer_types.py
```diff
@@ -0,0 +1,38 @@
+from libertem.udf import UDF
+
+class test_buffer(UDF):
```
sk1p: Style: class names should be `CamelCase`

docs/source/udf/buffer_types.py
```diff
@@ -0,0 +1,38 @@
...
+
+	def get_results_buffer(self):
```
sk1p: typo: `get_result_buffers`

docs/source/udf/buffer_types.py
```diff
@@ -0,0 +1,38 @@
...
+		"""
+		Initialize a buffer of different types
```
sk1p: Plural: buffers?

docs/source/udf/buffer_types.py
```diff
@@ -0,0 +1,38 @@
...
+
+		The first buffer "single_buffer" specifies a buffer of
+		dimension (16, 16) with entries of type float32
+
+		The second buffer "sig_buffer" specifies a buffer whose
+		first component dimension is increased by 2 from that of
+		signal dimension
+
+		The third buffer "nav_buffer" specifies a buffer whose
```
sk1p: I think for these three buffers, it would help to give example shapes. If for example my dataset has shape `(14, 14, 32, 32)` with two signal dimensions (`nav` would be `(14, 14)`, sig `(32, 32)`), how would the concrete shapes represented by each buffer be like?

Also, what exactly do you mean by "increased" here? It may be misunderstood to mean the existing dimensions are somehow changed, where in reality, new dimensions are introduced.

docs/source/udf/buffer_types.py
```diff
@@ -0,0 +1,38 @@
+from libertem.udf import UDF
```
sk1p: Is it intentional that the file is indented by tabs? Usually, python code is indented by 4 spaces, but you can leave it as-is if there is a reason.

docs/source/udf/buffer_types.py
```diff
@@ -1,22 +1,30 @@
...
 
 		The third buffer "nav_buffer" specifies a buffer whose
-		first component dimension is increased by 1 and the second
-		component dimension is increased by 2 from that of navigation
-		dimension
+		dimension is initialized to be the same as the navigation dimension.
+		With 'extra_shape = (1, 2)', one additional dimension is added to the
+		first dimension of the navigation dimension and two additional
+		dimensions are added to the second dimension of the navigation dimension,
+		so the resulting dimension for this buffer will be (15, 16)
```
sk1p: There is still a misunderstanding here - if you specify `extra_shape`, extra dimensions are introduced! In this example, the full shape would be `(14, 14, 1, 2)` - so you can store two `float32` numbers as a result for each frame. The **size** of the existing navigation dimensions stays the same.

(Same applies to the `sig_buffer` paragraph above)

jaewshin: Oh got it. Sorry for the misunderstanding! I'll fix it immemdiately

sk1p:
Thanks again for the PR, I'm going to merge it into master now. If you want to continue the work on the docs, feel free to just open a new PR.

sk1p:
Oh, the target was still set to `udf_v2_wip`. Let me try to fix that.

sk1p:
Fixed by cherry picking the commits into the master branch

PR #362, opened 2019-06-14
Streamline code around RadialFourierAnalysis

### LINKED ISSUE #361: Confirm selecting the correct matrix back-end in RadialFourierAnalysis
It seems to have switched to dense matrix for a rather deep mask stack with many bins and orders. That exceeded the memory budget of the workers. Enforcing use_sparse='scipy.sparse' helped, which means that the automatic switching logic likely failed.

Code is currently in udf_wip_v2.

uellue:
* Optionally include raw results with analysis
* Corrected and improved back-end selection, closes #361 
* Allow overriding the sparse back-end
* More control over radial_bins() to generate it in the desired format right from the start and avoid conversions
* Visualization and AnalysisResult list optionally lazy to avoid unnecessary overheads when using RadialFourierAnalysis within a script. The savings are very significant in some cases and remove a bottleneck in the head node.
* Add visualization of dominant order, display as default

### DIFF
--- a/src/libertem/analysis/base.py
+++ b/src/libertem/analysis/base.py
@@ -9,7 +9,7 @@ class AnalysisResult(object):
     """
     def __init__(self, raw_data, visualized, title, desc, key):
         self.raw_data = raw_data
-        self.visualized = visualized
+        self._visualized = visualized
         self.title = title
         self.desc = desc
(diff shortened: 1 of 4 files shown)


sk1p:
Thanks, merged!

PR #365, opened 2019-06-16
Meta as argument for get_result_buffers()

uellue:
Include dtype in meta

This is required for setting dtypes of result buffers to match the result type
with a dataset.

Required to implement UDFs with equivalent functionality and behavior of MaskJob

FIXME continue porting dtype switching from MaksJob to all Jobs and UDFs

### DIFF
--- a/docs/source/udf/max.py
+++ b/docs/source/udf/max.py
@@ -4,7 +4,7 @@ from libertem.udf import UDF
 
 
 class MaxUDF(UDF):
-    def get_result_buffers(self, meta):
+    def get_result_buffers(self):
         """
         Describe the buffers we need to store our results:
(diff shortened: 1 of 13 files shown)


sk1p:
Offline discussion: adding a `meta` parameter to `get_result_buffers` may not be the best option; instead, we could make `meta` always available on `UDF` instances. This could be implemented using a `set_meta(...)` method on `UDFBase` which is then called in both `run_on_dataset` and `run_on_partition` contexts.

uellue:
> This could be implemented using a `set_meta(...)` method on `UDFBase` which is then called in both `run_on_dataset` and `run_on_partition` contexts.

Done! :-)

sk1p:
Looks good, thanks!

PR #367, opened 2019-06-17
Ameliorate the effect of network backpressure

uellue:
Sending the messages at the end of the loop can block because of network backpressure if the connection between client and server is slow or unreliable.

Previously, this could severely stall the loop because it would enforce an update for each calculation result, acerbating the problem.

With this change, the delay from sending messages is not counted for the debouncing, which allows the server to process a larger number of results until sending the next message over the network. Additionally, this skips a lot of visualization steps in combination with the lazy calculation implemented in PR #362.

Implement the same logic for UDFs.

Thx @sk1p for the help! :-)


sk1p:
Looks good, merged!

Issue #134, opened 2018-10-04
GUI rework/rethink


sk1p:
There are several drawbacks to the way the current GUI works:

 * only works well for square-ish scan regions and square-ish frames. tall or very wide scans are scaled down without a way to zoom in
 * does not work well with high-resolution images
 * does not use screen real estate well
 * not flexible enough for complex workflows

For 0.2, I'd like to refactor the UI to make the following possible:

 * zoom into/out of both picked frames and results
 * fullscreen view of images
 * more complex workflows, like: make HAADF image, then pick a representing frame from the HAADF image, then compute something based on the selected frame (example: distance of all frames from selected frame according to some metric)

The following could be of interest:

 * http://golden-layout.com/tutorials/getting-started-react.html - resizable/reorderable tabs/split-views/windows (see https://godbolt.org/ for a live example)

When we have decided what kind of workflows we want to support in 0.2 and in the future, this issue should be updated accordingly.

uellue:
We should have a clean concept on how to show context information for an analysis. Right now the info under frame and result window is a bit unstructured, just concatenating various text bits and elements, and adding a new item therefore needs tinkering. This made moving the pick coordinate display #164 more difficult than necessary, for example. I was wrestling with separator characters and `&emsp;` ...

sk1p:
There are several related issues in the [GUI project](https://github.com/LiberTEM/LiberTEM/projects/2), #235 restating the problems with large frames.

sk1p:
For visualization changes like colormap selection (#166), value range, log scale etc (#128). we will have the same problem as in #235 - image encoding will limit the performance and hence the interactive response time. We may want to rethink the visualization aspect, and maybe move it to the client.

For colormapped visualization in 2D: https://github.com/santilland/plotty https://av.tib.eu/media/20373

For possible 3D visualization: http://deck.gl/#/examples/custom-layers/3d-surface-explorer

For a more fully featured framework: https://gr-framework.org/tutorials/using_gr_in_javascript.html

Deck.gl and plotty.js rely on WebGL (plotty actually has a fallback to plain old canvas) - we have to think if our users always have it available. On the other hand, WebGL is a natural fit for the problem, and allows to offload visualization to the GPU. Implementing a zoomable interface should also work well with WebGL. Deck.gl already implements a zoomable interface by integrating with mapbox-gl-js, but that's not really a good fit for our application, I think. I mean, one could create a 'projection' that fits the physical units, and position the raster data on that "map"...

[...]

PR #369, opened 2019-06-18
Adaptive rate for GUI update

uellue:
Adapt the integration window for results to match the postprocessing and communication time.

This strongly reduces the update rate and increases the integration window if the network creates
backpressure or if calculating the result images is slow. That improves the overall performance and
user experience. We are not getting the results to the GUI anyway -- at least the merging is not
stalled and the next update will show more data when it finally comes.

### DIFF
--- a/src/libertem/web/jobs.py
+++ b/src/libertem/web/jobs.py
@@ -105,15 +105,15 @@ class JobDetailHandler(CORSMixin, tornado.web.RequestHandler):
 
         t = time.time()
         post_t = time.time()
-        prev = 0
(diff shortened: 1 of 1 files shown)


uellue:
The solution is not ideal yet, but significantly better than before.

In a future version we should make sure that:

* Updates are smaller in size
* Buffers and queues between back-end and client are minimized and the back-end process "feels" back pressure immediately

sk1p:
Looks good, merging! The future part can then be fixed as part of #134 I think.

PR #377, opened 2019-06-22
Auto UDF to allow trivial UDFs without boilerplate

uellue:
That can wrap a function that accepts a frame as a single argument. Together with functools.partial, this allows for a lot of flexibility.

### DIFF
--- a/docs/source/udf.rst
+++ b/docs/source/udf.rst
@@ -2,8 +2,9 @@ User-defined functions
 ======================
 
 A common case for analysing big EM data sets is running a reduction operation
-on each frame of the data set. The user-defined functions (UDF) interface
-of LiberTEM allows users to run their own reduction functions easily, without having
+on small subsets of the data set, for example individual detector frames.
+The user-defined functions (UDF) interface of LiberTEM allows users to run their
+own reduction functions easily, without having
 to worry about parallelizing, I/O, the details of buffer management and so on. 
 
 
@@ -14,8 +15,8 @@ How UDF works in layman's terms
 
 The UDF interface of LiberTEM is heavily utilizing the existing LiberTEM architecture. First,
 data is partitioned into several `"partitions"` and distributed across workers. Then, each partition,
-which can be viewed as a collection of frames, are processed by the user-defined `process_frame` method.
-Here, frames are fetched in an iterative manner, and `process_frame` method performs user-defined operations
+which can be viewed as a collection of frames, are processed by the user-defined `process_frame` or `process_tile` method. Tiles are described in more detail in TODO confirm link `Tiled processing`_ and are useful to improve the performance of some numerical operations. In the following, the simpler `process_frame` interface is described first.
+Here, frames or tiles are fetched in an iterative manner, and the `process_frame` or `process_tile` method performs user-defined operations
 on the frames. After all the frames in a partition are processed, LiberTEM iteratively `merges` the results from each worker, which is called
 by the `merge` method in UDF class. To summarize, the UDF interface of LiberTEM performs operations at two levels: `process_frame`, which performs user-defined operations
 on frames within each partition, and `merge`, which merges the output of `process_frame` from each partition. Note that in both `process_frame` and `merge`, buffers store the intermediate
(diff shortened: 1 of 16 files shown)


docs/source/udf.rst
```diff
@@ -64,16 +66,30 @@ to a single value:
...
+Auto UDF
+--------
 
+The :code:`AutoUDF` class and :code:`run_auto()` function allow to run simple functions that accept a frame as the only parameter with an auto-generated :code:`kind="nav"` result buffer over a dataset ad-hoc without defining an UDF class. The :code:`AutoUDF` class determines the output shape and type by calling the function with a mock-up frame of the same type and shape as a real detector frame and converting the return value to a numpy array. The :code:`extra_shape` and :code:`dtype` parameters for the result buffer are derived automatically from this numpy array. Additional constant parameters can be passed to the function via :code:`functools.partial`, for example.
+
+Example: Calculate sum over the last signal axis.
+
+.. code-block:: python
+
+      result = libertem.udf.run_auto(
```
sk1p: Hmm, looking at this example, maybe we could make it even easier:

```
ctx.map(dataset=dataset, f=functools.partial(np.sum, axis=-1), roi=roi)
```

As this is more of a 'facility' than a concrete UDF, I would argue it has its place in the `Context`. 

We should also have a general discussion about the `Context` API, what to add vs. what not to add there etc. - this also comes into play when thinking about breaking LiberTEM into more coarse modules that are separately useful. Maybe something for Monday? cc @kruzaeva

docs/source/udf.rst
```diff
@@ -64,16 +66,31 @@ to a single value:
...
+
+The :class:`AutoUDF` class and :meth:`Context.map()` method allow to run simple functions that accept a frame as the only parameter with an auto-generated :code:`kind="nav"` result buffer over a dataset ad-hoc without defining an UDF class. For more advanced processing, such as custom merge functions, postprocessing or performance optimization through tiled processing, defining an UDF class is required.
+
+The :class:`AutoUDF` class determines the output shape and type by calling the function with a mock-up frame of the same type and shape as a real detector frame and converting the return value to a numpy array. The :code:`extra_shape` and :code:`dtype` parameters for the result buffer are derived automatically from this numpy array. Additional constant parameters can be passed to the function via :meth:`functools.partial`, for example. The return value should be much smaller than the input size for this to work efficiently.
 
+Example: Calculate sum over the last signal axis.
+
+.. code-block:: python
+
+      result = ctx.map(
```
sk1p: Maybe the equivalent `AutoUDF(f=f)` plus `run_udf` call should also be documented here? Otherwise it makes little sense to even talk about `AutoUDF`, which then becomes only an implementation detail.

docs/source/udf.rst
```diff
@@ -14,46 +15,47 @@ How UDF works in layman's terms
 
 The UDF interface of LiberTEM is heavily utilizing the existing LiberTEM architecture. First,
 data is partitioned into several `"partitions"` and distributed across workers. Then, each partition,
-which can be viewed as a collection of frames, are processed by the user-defined `process_frame` method.
-Here, frames are fetched in an iterative manner, and `process_frame` method performs user-defined operations
+which can be viewed as a collection of frames, are processed by the user-defined :meth:`process_frame` or :meth:`process_tile` method. Tiles are described in more detail in `tiled processing`_ and are useful to improve the performance of some numerical operations. In the following, the simpler :meth:`process_frame` interface is described first.
```
sk1p: These are still not linked to the right places in the reference documentation; the proper way appears to be to mention the fully-qualified name and prefix with a tilde, i.e. 

```
:meth:`~libertem.udf.UDFFrameMixin.process_frame`
```

That should properly link to the reference, if one exists.

uellue:
> There are still some general structure issues I'd like to improve in the documentation, but I think the main point of this PR is addition of `Context.map` and `AutoUDF`, and documentation for these. And that looks already good.

Thanks for the review, nice catches! :-)

Yes, we should definitely work on the structure. More PRs to follow. Next on my list is Blobfinder. :-)

sk1p:
Looks good, thanks!

Issue #286, opened 2019-03-07
User-defined functions


sk1p:
UDF Design goals:

- easy to understand, small API surface
- easy to use 
  - do index calculations etc. for the user where possible
  - fail fast and loudly in case of errors
- efficient
  - low overhead
  - should perform at least as well as current Analyses
- composable
  - allow chaining and execution of independent UDFs in a single pass over the data

This is an overview issue for user-defined functions. As a first step, we want to support mapping functions over frames. ~This is already partially done in the `phase-correlation-proto` branch, where we developed this for supporting the analysis developed by @kruzaeva.~ Now merged in master! :rocket:  We still have some work to do to develop a stable and user-friendly interface, and to support some more use-cases:

- [x] Low-level functionality for mapping functions over frames
- [x] Support for auxiliary data, for example parameters for each frame #287
- [x] Integration into the Web API, for showing partial results. See #216 
- [ ] Type-checking for better developer experience: #262 
- [ ] Support for running functions in the cluster without reading an input dataset #288
- [ ] Support for creating a new dataset, for example by running a distributed simulation #289
- [x] By-tile operations: #310
- [x] Documentation: #306

Future work:

- [ ] Experiment with auto-jit-ing via numba (related: #350)
- [x] Port over our existing analyses

sk1p:
UDFs have landed in `master`, at least the low-level functionality. The interface is quite nice already, I think, but may change in the near future, as we implement some more functionality with this.

uellue:
Using `BufferWrapper` with `dtype="object"` is confirmed to work. That means we can use pretty much any data type(s) we like as return values for UDFs. Performance TBD!

sk1p:
See also the interface discussions in #310 - I'm going to experiment a bit with the interfaces and see if I can come up with something elegant that matches the constraints discussed here and in related tickets.

sk1p:
For reference, here is the class-based interface proposed in #310:

```python
class MyUDF_1:
    wish = "floats"
    
    def __init__(self):
        self.local_variable = 42
        
    
    def process(self, frame):
        print(self.local_variable, "UDF 1: ", frame)
        

class MyUDF_2:
    wish = "ints"
    
    def __init__(self):
        self.local_variable = 23
    
    def process(self, frame):
        print(self.local_variable, "UDF 2: ", frame)
        
uu1 = MyUDF_1()
uu2 = MyUDF_2()

print(uu1.wish)
print(uu2.wish)

[...]

PR #383, opened 2019-06-28
Create dask arrays from LiberTEM DataSet; UDF process_partition. Based on PR #377

### LINKED ISSUE #376: Create Dask arrays from LiberTEM DataSets
It should be possible to write a wrapper for our dataset readers that construct a distributed dask array using our partitions as chunks. That would probably be a very useful feature for dask since it would make it much easier to construct good dask arrays from the file formats we support. That way, our performance optimizations would be available in Hyperspy and other software that can build on dask arrays.

Edit: Running UDFs on Dask arrays is now #384.

uellue:
Both are based on creating a tile that covers an entire partition.

FIXMEs:
* Documentation
* Tests for each supported file type -- may require special measures as encountered in K2IS support

Based on PR #377, mostly for practical testing through the new Context.map() function.

Closes #376
See also #286

### DIFF
--- a/docs/source/udf.rst
+++ b/docs/source/udf.rst
@@ -81,6 +81,8 @@ Auto UDF
 
 The :class:`~libertem.udf.AutoUDF` class and :meth:`~libertem.api.Context.map` method allow to run simple functions that accept a frame as the only parameter with an auto-generated :code:`kind="nav"` result buffer over a dataset ad-hoc without defining an UDF class. For more advanced processing, such as custom merge functions, post-processing or performance optimization through tiled processing, defining an UDF class is required.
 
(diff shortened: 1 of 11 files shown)


uellue:
The AppVeyor fail seems strange to me. Does it have anything to do with the changes or is it a spurious error?

docs/source/udf.rst
```diff
@@ -81,6 +81,8 @@ Auto UDF
 
 The :class:`~libertem.udf.AutoUDF` class and :meth:`~libertem.api.Context.map` method allow to run simple functions that accept a frame as the only parameter with an auto-generated :code:`kind="nav"` result buffer over a dataset ad-hoc without defining an UDF class. For more advanced processing, such as custom merge functions, post-processing or performance optimization through tiled processing, defining an UDF class is required.
 
+As an alternative to Auto UDF, you can use the :meth:`~libertem.io.dataset.base.DataSet.get_dask_array` method of a :class:`~libertem.io.dataset.base.DataSet` and use a `dask.array <https://docs.dask.org/en/latest/array.html>`_ to perform your calculations. See :ref:`Integration with Dask arrays<daskarray>` for more details.
```
sk1p: Needs to be updated to point to function instead of method, looks good otherwise!

uellue: Thx, nice catch, fixed!

PR #385, opened 2019-07-01
Fluctuation EM GUI

kruzaeva:
Calculating of SD over a ring mask


sk1p:
The CI failure is not your fault, it is a change in the new version of pytest, see also https://github.com/pytest-dev/pytest/issues/5412

Going to fix this in a minute...

sk1p:
If you now fetch from master and push the result, CI should pass.

src/libertem/analysis/fem.py
```diff
@@ -0,0 +1,25 @@
+from libertem.viz import visualize_simple
+
+from .base import BaseAnalysis, AnalysisResult, AnalysisResultSet
+
+import libertem.udf.FEM as FEM
+
+
+class ApplyFEM(BaseAnalysis):
```
sk1p: What do you think about keeping in the naming scheme of `SomethingAnalysis` here? That's currently used by all other analyses

sk1p:
Hi, to get this merged, the conflicts need to be resolved. If you pull from upstream/master, resolve conflicts, commit and push it should work.

sk1p:
Thanks, merged!

PR #394, opened 2019-07-06
Custom templates for blobfinder

### LINKED ISSUE #392: Support user-defined templates for blobfinder
Right now, the blobfinder supports two templates, radial gradient and background subtraction, which each have radial symmetry.

In reality, the CBED disks are not always round disks.

For that reason, the blobfinder should support user-defined templates. In particular it should be easily possible to generate good templates from real data, for example by binarizing with a threshold and multiplying with a radial gradient.

uellue:
Include utility functions to generate advanced templates

The combination of background subtraction, radial gradient, antialiased border and elliptical distortion to match the CBED disk as closely as possible gives outstanding results with CBED strain maps at higher convergence angles recorded with fine detector resolution.

Fix a typo in function name. This might break examples or downstream code, should be included in release 0.2.

Closes #392

### DIFF
--- a/docs/source/reference.rst
+++ b/docs/source/reference.rst
@@ -15,6 +15,11 @@ Python API
 Internal API
 ------------
 
+.. automodule:: libertem.masks
+   :members:
+   :undoc-members:
+   :special-members: __init__
+
 .. automodule:: libertem.common.shape
    :members:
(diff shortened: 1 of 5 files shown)


sk1p:
Thanks!

PR #401, opened 2019-07-22
Gui region clustering

kruzaeva:
GUI for region Clustering (a new type of analysis)
+Rectangular masks of ROI for Average and SD default frame mode.
+GUI for region clustering based on non-zero order peaks position:
SD calculation->
Peaks finding->
Feature vector forming->
Clustering


src/libertem/udf/feature_vector_maker.py
```diff
@@ -73,9 +74,9 @@ def make_feature_vec(ctx, dataset, num, delta, center=None, rad_in=None, rad_out
         Returns array of coordinates of possible peaks positions
 
     """
-    res_stat = run_stddev(ctx, dataset)
+    res_stat = run_stddev(ctx, dataset, roi)
     savg = res_stat['mean']
-    sstd = res_stat['std']
+    sstd = res_stat['var']
```
sk1p: Maybe `sstd` should be renamed to `svar`, if it actually is the variance?

src/libertem/udf/feature_vector_maker.py
```diff
@@ -84,10 +85,13 @@ def make_feature_vec(ctx, dataset, num, delta, center=None, rad_in=None, rad_out
         masked_sstd = sstd*mask
     else:
         masked_sstd = sstd
-    coordinates = peak_local_max(masked_sstd, num_peaks=num, min_distance=0)
+    if not (min_dist is None):
```
sk1p: Better to read: `if min_dist is not None:`. Another possibility would be to have only a single `peak_local_max` call, and set `min_dist` to 1 if it is `None`.

src/libertem/udf/sumsigudf.py
```diff
@@ -0,0 +1,21 @@
+import numpy as np
+
+from libertem.udf import UDF
+
+
+class sumsigUDF(UDF):
```
sk1p: Classes should be in `CapitalCamelCase`, so `SumsigUDF` or `SumSigUDF` here (matching the `SumSigAnalysis`).

src/libertem/web/jobs.py
```diff
@@ -95,6 +98,14 @@ def get_analysis_by_type(self, type_):
         self.finish()
         self.event_registry.broadcast_event(msg)
 
+        if hasattr(analysis, 'controller'):
+            return await analysis.controller(
+                cancel_id=uuid, executor=executor,
+                job_is_cancelled=lambda: self.data.job_is_cancelled(uuid),
+                send_results=lambda results, finished: self.send_results(results, uuid,
+                 finished=finished)
```
sk1p: Nitpick: `finished` should be aligned with the other params (yeah, sorry, my OCD is kicking in :grin:)

client/src/analysis/components/Clustering.tsx
```diff
@@ -0,0 +1,212 @@
...
+                    width={scanWidth} height={scanHeight}
+                    extraHandles={resultHandles}
+                    extraWidgets={resultWidgets}
+                />
+            </>}
+            toolbar={toolbar}
+            clustparams= {clustparams}
+
+            title1="Frame view. Choose ring parameters for peak finding"
+            title2="Bright field image (sum over frame). Choose ROI for SD map calculation"
```
sk1p: `title2`: Maybe we should write this more abstract here. I think important here is that the user is choosing a region for peak detection, right?

client/src/analysis/components/Clustering.tsx
```diff
@@ -0,0 +1,212 @@
...
+                    jobIndex={2} analysis={analysis.id}
+                    width={scanWidth} height={scanHeight}
+                    extraHandles={resultHandles}
+                    extraWidgets={resultWidgets}
+                />
+            </>}
+            toolbar={toolbar}
+            clustparams= {clustparams}
+
+            title1="Frame view. Choose ring parameters for peak finding"
```
sk1p: `title1`: should describe that only peaks inside the ring are included in the feature vector

client/src/analysis/components/Clustering.tsx
```diff
@@ -0,0 +1,212 @@
...
+    const riHandle = {
+        x: cx - ri,
+        y: cy,
+    }
+    const roHandle = {
+        x: cx - ro,
+        y: cy,
+    }
+
+    const [delta, setDelta] = React.useState(0);
```
sk1p: Is 0 the correct default value for `delta`? That makes the peak detection quite sensitive, right?

client/src/analysis/components/DefaultFrameView.tsx
```diff
@@ -15,68 +28,144 @@ const useDefaultFrameView = ({
...
+            value: DefaultModes.SUM,
+        },
+        {
+            text: "SD",
+            value: DefaultModes.SD,
         },
         {
-            text: "Pick",
-            value: AnalysisTypes.PICK_FRAME,
+            text: "PICK",
```
sk1p: Please, change back to `Pick`, `PICK` is a bit too screamy for my taste :grinning:

client/src/analysis/components/RoiPicker.tsx
```diff
@@ -1,88 +1,54 @@
...
-
-    const handleRChange = (newR: number) => {
-        setR(newR);
-    };
 
-    const rHandle = {
-        x: cx - r,
-        y: cy,
-    }
+        React.useEffect(() => {
```
sk1p: Indentation needs to be fixed here.

client/src/analysis/components/RoiPicker.tsx
```diff
@@ -1,88 +1,54 @@
...
-                    type: AnalysisTypes.SUM_FRAMES,
+                    type: AnalysisTypes.SD_FRAMES,
                     parameters: roiParameters,
                 }))
             }
         }, 100);
 
         return () => clearTimeout(handle);
-    }, [analysisId, enabled, jobIndex, cx, cy, r]);
+    }, [analysisId, enabled, jobIndex, JSON.stringify(roiParameters)]);
```
sk1p: As you are using `analys` inside of the effect function, you also need to add it to the list of dependencies here. Same for `useEffect` below.

client/src/analysis/components/DefaultFrameView.tsx
```diff
@@ -15,68 +28,144 @@ const useDefaultFrameView = ({
...
         cx, cy, setCx, setCy
     });
 
-    const { sumRoiHandles, sumRoiWidgets } = useRoiPicker({
-        enabled: frameMode === AnalysisTypes.SUM_FRAMES_ROI,
+
+    const { RectRoiHandles, RectRoiWidgets, RectroiParameters }  = useRectROI({ scanHeight, scanWidth })
+    const { diskRoiHandles, diskRoiWidgets, diskroiParameters}  = useDiskROI({ scanHeight, scanWidth })
+
+    useRoiPicker({
```
sk1p: Hmm. There are now so many different `useRoiPicker` calls in here, when in reality, we basically only have a) two different analyses (sum, sd) and three or four types of ROI (whole dataset, disk, rect, "point" in the case of pick). Maybe we can work more on decoupling ROI selection and running the analysis?

client/src/analysis/components/Clustering.tsx
```diff
@@ -0,0 +1,212 @@
...
+    const clustparams =
+    <Accordion>
+    <Accordion.Title active={paramsVisible} index={0} onClick={handleClick}>
+          <Icon name='dropdown' />
+          Parameters
+    </Accordion.Title>
+    <Accordion.Content active={paramsVisible}>
+    <Form>
+        <Form.Field> 
+                <label> Delta. Relative intensity difference for decision making for feature vector value (delta = (x-ref)/ref, so, normally, value should be in range [0,1]) </label><input type="number" value={delta} step="0.01" min="0" max="2" onChange={deltaChange}/>
```
sk1p: Nitpick/usability: clicking on a `label` should activate the associated `input`

client/src/analysis/components/Clustering.tsx
```diff
@@ -0,0 +1,212 @@
...
+        </Form.Field>
+        <Form.Field>
+            <label> Number of clusters </label> <input type="number" value={n_clust}  step="1" min="2" max="100" onChange={clustChange}/> 
+        </Form.Field>    
+        <Form.Field>
+            <label>  Maximal number of possible peak positions to detect (better put higher value,
+        the output is limited to the number of peaks the algorithm could find) </label> <input type="number" value={n_peaks}  step="1" min="5" max="200" onChange={peakChange}/>    
+        </Form.Field>
+        <Form.Field>
+            <label>  Minimal distance between peaks </label> <input type="number" value={min_dist}  step="1" min="0" max="100" onChange={min_distChange}/>
```
sk1p: distance in pixels, right?

sk1p:
One more thing I noticed while reviewing this PR: the [description in the docstring of `make_feature_vec`](https://github.com/LiberTEM/LiberTEM/blob/ee61c84880c33c1725bc2fbbc34e9728578ecc24/src/libertem/udf/feature_vector_maker.py#L28) needs to be fixed, it describes the crystallinity analysis, not feature vector construction.

Issue #399, opened 2019-07-18
Update the GMS Python live demo prototype


uellue:
The LiberTEM internals that the interactive live processing demo prototype in GMS Python has relied on have changed. For that reason it currently doesn't work.

Since scripting in combination with live data stream and visualization is a nice application and it shows the power of GMS + Python at the same time, we should update the demo in due time.

Discussion with @sk1p: A good time could be when the older "Job" interface for mask application is replaced with an interface based on UDFs.

uellue:
Related to #675

uellue:
Adressed in #1011, closing!

PR #402, opened 2019-07-23
Updating DM examples

uellue:
* copy() because of DM API changes
* Use EMPAD reader
* FIXME async prototype is currently broken, see Issue #399


sk1p:
Thanks, merged!

PR #404, opened 2019-07-26
Documentation update

uellue:
* Review and update of existing documentation
* Added Fluctuation EM
* Added basic benchmark results


uellue:
@sk1p Could you please have a look at the latest Travis failure when you have time? This looks weird to me, apparently a change in a dependency that broke stuff.

sk1p:
Yeah, looks like `sparse` needs updates for the recent `numba` release, see: https://github.com/pydata/sparse/issues/257. For now (0.2), we can pin `numba` to the previous release; it is a bit unfortunate, but if we keep with our plan of releasing more often, should not be much trouble (We can do the next release directly when `sparse` is updated to work well with the new `numba` release .)

uellue:
Actually, it is `numpy<=1.16.4` that has to be pinned. A bit unfortunate, hopefully this will be fixed soon.

README.rst
```diff
@@ -57,9 +57,9 @@ or by `joining our Gitter chat <https://gitter.im/LiberTEM/Lobby>`_.
 LiberTEM is suitable as a high-performance processing backend for other applications, including live data streams. `Contact us <https://gitter.im/LiberTEM/Lobby>`_ if you are interested! 
 
 
-LiberTEM is evolving rapidly and prioritizes features following user demand and contributions. In the future we'd like to implement live acquisition, and more analysis methods for all applications of pixelated STEM.
+LiberTEM is evolving rapidly and prioritizes features following user demand and contributions. In the future we'd like to implement live acquisition, and more analysis methods for all applications of pixelated STEM and other large-scale detector data.
 If you like to influence the direction this
-project is taking, or if you'd like to contribute, please join our `gitter chat <https://gitter.im/LiberTEM/Lobby>`_,
+project is taking, or if you'd like to :doc:`contribute <contributing>`, please join our `gitter chat <https://gitter.im/LiberTEM/Lobby>`_,
```
sk1p: Using roles like `:doc:` in the README.rst may be problematic, as they may not be supported by the pypi renderer.

See also https://github.com/LiberTEM/LiberTEM/commit/f01df221383970bc9edad438a88d83ad231ca43b - if you rebase to current master you should get a validation error in CI.

sk1p:
I went ahead and merged it as-is, we can fix the things I commented later. See also #410

PR #406, opened 2019-08-06
Compatibility with numpy 1.17

uellue:


uellue:
Hm, something is strange about the coverage numbers. I am sure that this change doesn't decrease coverage by 18.21 % @sk1p 🤔?

sk1p:
That was a glitch in the previous build, actually:

Here we see a +18% delta for a commit that only touches untracked files: https://codecov.io/gh/LiberTEM/LiberTEM/commit/8ce088b

This is because the coverage upload for the node-based client failed (see very bottom of [the build log](https://travis-ci.org/LiberTEM/LiberTEM/jobs/565436403))

And in the next build it is reversed, as the node build successfully uploads coverage data: https://codecov.io/gh/LiberTEM/LiberTEM/commit/ebe2a4e19d83f955ebcd1d320c5de8472e7a08be

tests/utils.py
```diff
@@ -187,7 +187,7 @@ def _naive_mask_apply(masks, data):
     for mask in masks:
         assert mask.shape == data.shape[2:], "mask doesn't fit frame size"
 
-    dtype = np.result_type(*masks, data)
+    dtype = np.result_type(*[m.dtype for m in masks], data.dtype)
```
sk1p: Is this needed for sparse arrays?

sk1p:
Merged, thanks for taking care of this!

tests/utils.py
uellue: Yes, for some reason result_type doesn't accept sparse arrays anymore with numpy 1.17

PR #426, opened 2019-09-12
Fix aux data in combination with `roi` and `postprocess`

sk1p:
Supersedes #417 - _should_ fix the issues described there. Didn't yet re-run the notebook, but tests pass.


sk1p:
Build failure is because of flaky test, going to merge as-is.

PR #446, opened 2019-10-07
Fixed a glitch where negative peak elevations were possible

### LINKED ISSUE #283: Spurious test failure
See https://travis-ci.org/LiberTEM/LiberTEM/jobs/501225151 for details

uellue:
Edit: No peak_local_max involved, root cause not sure -- numerical error?

> Peak_local_max didn't give the maximum of the corrmap in some border cases,
> likely in a result that is near zero.

Include more fuzzing and better reporting in test

Closes #283


sk1p:
Fix looks good, even though it is puzzling what the root cause is. Having a look at the CI log, the test case takes about 5 seconds to run... I've opened a separate issue for this: #447

PR #448, opened 2019-10-09
Improve slow tests, instructions for profiling, rewrap

### LINKED ISSUE #447: Audit test suite for run time
We have some test cases that take several seconds to run. For example, from [a recent CI run](https://travis-ci.org/LiberTEM/LiberTEM/jobs/594576315):

```
5.82s call     tests/test_analysis_masks.py::test_numerics_fail
5.36s call     tests/udf/test_blobfinder.py::test_run_refine_affinematch
4.99s call     tests/udf/test_blobfinder.py::test_run_refine_sparse
4.21s call     tests/test_analysis_masks.py::test_numerics_succeed
3.45s call     tests/server/test_job.py::test_run_job_1_sum
```

That's like 30% of time in the top 5 test cases... :frowning_face: 

This is not only an issue in CI, but also when running the tests locally. We should strive to improve performance here, to encourage devs (us!) to run the tests before committing. Possible to improve CI time: run slower tests in different job in parallel. But that would not improve local test runs, and also has quite some set-up overhead for the extra job. We could opt to have "quick" tests for running locally, which exercise the whole system but omit slower test for special cases.

In my opinion, CI time is still okay-ish, but we should look out to not have it spiral out of control.

Thoughts?

uellue:
Closes #447

### DIFF
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -3,59 +3,92 @@
 Contributing
 ============
 
-LiberTEM is intended and designed as a collaboratively developed platform for data analysis. That means all our development is coordinated openly, mostly on our `GitHub repository <https://github.com/LiberTEM/LiberTEM/>`_ where our code is hosted. Any suggestions, Issues, bug reports, discussions and code contributions are highly appreciated! Please let us know if you think we can improve on something, be it code, communication or other aspects.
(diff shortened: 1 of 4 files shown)


sk1p:
Thanks, much appreciated!

PR #449, opened 2019-10-10
Trigger a timeout when guessing parameters for HDF5 takes too long

### LINKED ISSUE #440: Opening a specific HDF5-based Matlab 7.3 file takes a lot of resources
When opening a specific Matlab .mat file in format 7.3, which is based on HDF5, one worker is using the CPU for several minutes. The task manager shows moderate file IO activity at the same time. During that time, LiberTEM is not responsive and there's no apparent way to terminate the process except restarting the LiberTEM server.

Update: This only happens with particular files.

image

Opening the same file with HDFView 3.1.0 gives a similar behavior.

Expected behavior: LiberTEM is responsive even if such a long-running operation is underway. The user can terminate it.

uellue:
TODO better error handling. Right now, the opening dialogue defaults to RAW
if the timeout is triggered.

TODO There seems to be code duplication. between _get_datasets() and detect_params().
@sk1p is there a specific reason to do it like that? I have to admit that I don't fully
understand the consequences of changing this.

TODO: Choose a well-balanced timeout value in case traversing a legitimate file
takes longer than that, for example over network.

Closes #440

### DIFF
--- a/src/libertem/io/dataset/hdf5.py
+++ b/src/libertem/io/dataset/hdf5.py
@@ -71,12 +71,7 @@ def _get_datasets(path):
 
     with h5py.File(path, 'r') as f:
         f.visititems(_make_list)
-        for name, size, shape, dtype in sorted(datasets, key=lambda i: i[0]):
-            yield {"name": name, "value": [
-                {"name": "Size", "value": str(size)},
-                {"name": "Shape", "value": str(shape)},
-                {"name": "Datatype", "value": str(dtype)},
-            ]}
+    return datasets
 
 
 class H5Reader(object):
@@ -165,20 +160,8 @@ class H5DataSet(DataSet):
             return False
 
         # try to guess the hdf5 dataset path:
-        datasets = []
-
-        timeout = 3
-        t0 = time.time()
-
-        def _make_list(name, obj):
-            if time.time() - t0 > timeout:
-                raise TimeoutError
-            if hasattr(obj, 'size') and hasattr(obj, 'shape'):
-                datasets.append((name, obj.size, obj.shape, obj.dtype))
-
-        with h5py.File(path, 'r') as f:
-            f.visititems(_make_list)
         try:
+            datasets = _get_datasets(path)
             largest_ds = sorted(datasets, key=lambda i: i[1], reverse=True)[0]
             name, size, shape, dtype = largest_ds
         except (IndexError, TimeoutError):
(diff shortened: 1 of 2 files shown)


uellue:
I've looked into creating an artificially slowed down HDF5 file to get test coverage for the new code. I tried to inject a modified version of `h5py.File` with a slowed-down function. However, that didn't work as expected. Can we leave this code path without coverage?

sk1p:
Based on your patch, I removed the duplication and added test cases.

>TODO better error handling. Right now, the opening dialogue defaults to RAW
if the timeout is triggered.

This is actually because the `TimeoutError` is thrown outside of the `try` block, so it propagates up and HDF5 is skipped completely as format. The rationale here is that different readers are tried on different files, and may throw exceptions as they are handed something that may look like garbage to them. Maybe we should at least log the exception in these cases (Linux does it like this IIRC; you may get log messages from different file systems when mounting without a specified file system.)

> I've looked into creating an artificially slowed down HDF5 file to get test coverage for the new code. I tried to inject a modified version of h5py.File with a slowed-down function. However, that didn't work as expected. Can we leave this code path without coverage?

What works, and without introducing a slow sleep somewhere, is just making `visititems` throw the `TimeoutError`. Or, to also cover the visitor function completely, one can make `time.time` return a value from a sequence of pre-defined timestamps on each call. I've included both approaches for now.

To get this right, it is crucial to patch it in the right namespace, in this case directly in `h5py.File` (this depends on how the module is imported, for example, if `libertem.io.dataset.hdf5` had `from h5py import File`, we would have to patch `libertem.io.dataset.hdf5.File` instead, IIRC.) There is a handy tool called `mock` in the stdlib that handles many of the special cases for us.

> TODO: Choose a well-balanced timeout value in case traversing a legitimate file
takes longer than that, for example over network.

Don't know about a good value, maybe let's change this once we hit it :grin:

uellue:
`mock` looks neat! Next time I'll play around with that. :-)

@sk1p thx for the improvements and guidance!

sk1p:
Could you re-test (after your weekend!) with your strange HDF5 file, and let me know if this also works in practice? Then I'll merge. Thanks!

uellue:
> Could you re-test (after your weekend!) with your strange HDF5 file, and let me know if this also works in practice? Then I'll merge. Thanks!

Tested and confirmed to work! 👍

sk1p:
Thanks, merged!

PR #451, opened 2019-10-14
Changelog in preparation to v.0.2.2

### LINKED ISSUE #443: Release notes?
We should add a step "write release notes" to the release process, and think about if/how we want to automate putting it into the [GitHub release](https://github.com/LiberTEM/LiberTEM/releases) and other places (documentation?)

Possible tool to use: https://pypi.org/project/towncrier/

uellue:
Closes #443

### DIFF
--- a/docs/source/app/phasechange.rst
+++ b/docs/source/app/phasechange.rst
@@ -7,6 +7,8 @@ Detection of the positions of additional peaks for all frames of crystalline reg
 
 Next, methods to distinguish crystalline and amorphous regions in phase-change materials will be described.
 
+.. _`crystallinity map`:
+
 Crystallinity map
 ~~~~~~~~~~~~~~~~~
 First, two different types of the regions, crystalline (low resistance) and amorphous (high resistance), should be distinguished.
@@ -58,7 +60,7 @@ Amorphous frame:
 
 .. figure:: ./images/phasechange/pickam.PNG
 
-
+.. _clustering:
 
 Clustering
 ~~~~~~~~~~
--- a/docs/source/app/phasechange.rst
+++ b/docs/source/app/phasechange.rst
@@ -64,9 +64,15 @@ Amorphous frame:
 
 Clustering
 ~~~~~~~~~~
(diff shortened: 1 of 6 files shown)


docs/source/changelog.rst
```diff
@@ -0,0 +1,136 @@
...
+#####
+
+Point release to fix a number of minor issues, most notably PR :pr:`439` that
+should have been merged for version 0.2.
+
+Bug fixes
+---------
+
+* Trigger a timeout when guessing parameters for HDF5 takes too long (:issue:`440` , :pr:`449`)
+* Improved error and validation handling when opening files with GUI (:issue:`433,442`)
```
sk1p: This is actually not part of the point release, as at least the GUI related changes are a bit entangled with TypeScript updates etc., added a new HTTP endpoint etc.

What do you think about just backporting [this commit](https://github.com/LiberTEM/LiberTEM/commit/ec74c1346d93eff58d9e2201a7ead5af7aa7cf44) and changing the log entry to something like "Slightly improved error [...]"?

uellue: Ok! I've updated the file accordingly.

sk1p:
Thanks, merged!

docs/source/app/phasechange.rst
```diff
@@ -58,13 +60,19 @@ Amorphous frame:
...
-
+.. _clustering:
 
 Clustering
 ~~~~~~~~~~
-To further categorize the crystalline regions according to their lattice orientation, clustering, based on non-zero diffraction peaks positions, can be used.
-The scripting interface allows to cluster membrane, amorphous and crystalline regions with different lattice orientation in a more efficient way, taking into account regions of interests.
-To look at full analysis `follow this link to a Jupyter notebook <pcmclustering.ipynb>`_.
+
+*This feature was added in version 0.3.0.dev0*
```
sk1p: I missed this when reviewing the PR, but: there are sphinx built-in directives for this, like [`:: versionadded`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded). We should use them in the future :grin:

PR #459, opened 2019-10-30
Update creators and contributors

uellue:
* Add ORCID to a number of entries
* Include ORCID and GitHub handle in output


sk1p:
Looks good, thanks!

PR #465, opened 2019-11-05
Use folder structure for changelog of continuous release

uellue:
This helps to avoid merge conflicts and keeps changes nicely separated.

For now, manual merging should do fine. As soon as it becomes annoying we can automate it.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -7,32 +7,10 @@ Changelog
 0.3.0.dev0 (continuous)
 #######################
 
-Features
---------
+.. toctree::
+   :glob:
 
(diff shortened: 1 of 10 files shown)


sk1p:
Looks good, merging! The testperformance snippet has a different headline marker, but it seems sphinx doesn't care in this case... I guess it is only important if the headline appears in the navigation tree?

uellue:
> The testperformance snippet has a different headline marker, but it seems sphinx doesn't care in this case... I guess it is only important if the headline appears in the navigation tree?

As far as I can see, it decides per file which marker is which headline level. We can work on keeping it consistent. For now, what works, works! :-)

PR #467, opened 2019-11-05
Round peak positions and convert to int

uellue:
Giving peak positions as floats is a common mistake and annoyance because
the wrong type leads to an opaque error message downstream. Converting
automatically to int gives the expected result and I can't imagine a scenario
where it wouldn't work.

Crystallizing PR #458


sk1p:
Removing annoyances sounds good, merging :grin:

PR #468, opened 2019-11-05
Blobfinder performance boost in particular for small peaks

uellue:
This is faster when correlating many, many small peaks on smaller frames

Extracted from PR #458, crystallizing


sk1p:
I would also be interested in both benchmark numbers comparing the methods, and profiling for the innermost loop.

Just as a remark, the file is getting quite large; we may want to think about restructuring. We could for example make `blobfinder` a package, with a module for shared functions and base classes, and a module for each correlation method. Maybe also a module for the match patterns. We can always keep backwards-compatibility imports from the `__init__` module. What do you think?

uellue:
> I would also be interested in both benchmark numbers comparing the methods, and profiling for the innermost loop.

The profile showed that the evaluation of correlation patterns is by far the bottleneck for small peaks. I am working on a @numba.njit version of the complete per-frame loop.

> Just as a remark, the file is getting quite large; we may want to think about restructuring. We could for example make `blobfinder` a package, with a module for shared functions and base classes, and a module for each correlation method. Maybe also a module for the match patterns. We can always keep backwards-compatibility imports from the `__init__` module. What do you think?

Ok, good point, I'll work on it!

uellue:
After optimization, `FastCorrelationUDF` is usually the fastest. `FullFrameCorrelationUDF` can be a bit faster (6.5 s vs. 10 s) for very many peaks (383 peaks on a 256x256 frame).

The changes slightly accelerated processing of K2 frames in the `strainmap-SiGe.ipynb` example with large peaks, too -- 2min 53s vs 3min 5s. The efficient processing pipeline seems to compensate for the additional memory use for the buffer to bundle the Fourier transforms, which could already get tight in terms of L3 cache use for this example.

uellue:
By the way, the coverage tools don't recognize when a Numba function calls other Numba functions. That's why they are not covered. There are three ways to handle this:

* Accept dropping coverage when porting code to Numba
* Mark code with `# pragma: no cover`
  * No added value beyond making the numbers look happy IMHO
* Try if marking code with `@numba.extending.register_jitable` allows to run them as normal Python functions in tests
  * That could be counterproductive since it would introduce a code path only for testing, while the real world uses the `@numba.njit` version.

@sk1p what do you think?

sk1p:
Sorry, by merging the `Circular` stuff this PR now has a conflict. Could you rebase to current master please?

> By the way, the coverage tools don't recognize [numba]

I think it is even worse, they don't record numba functions at all.

Another possibility:

1) Mark all tests that use numba functions with some pytest marker, for example `with_numba`

2) Run _all_ tests with `NUMBA_DISABLE_JIT=1` to gain coverage information (also this will run the numba kernels with loop bounds checks, which IIRC are disabled for njit compiled code)

3) Re-run the `with_numba`-marked tests without setting the environment variable. 

Pro: best test coverage
Con: "slightly" slower tests, depending on the test case input sizes

sk1p:
> Sorry, by merging the Circular stuff this PR now has a conflict. Could you rebase to current master please?

Other than that, the PR looks good! :+1: :+1: I would merge this, and we can fix the numba coverage issue separately.

uellue:
> we can fix the numba coverage issue separately.

Ok, sounds good! :-)

PR #469, opened 2019-11-05
Plain circular match pattern

uellue:
This is useful to construct feature vectors, or for comparison.

Crystallizing PR #458


sk1p:
Looks good, thanks!

PR #471, opened 2019-11-07
ClusterDataSet and CachedDataSet

sk1p:
This is the low-level part of cluster datasets and the caching layer. These are not supposed to be used by the user directly, but rather via more high-level methods that still need to be implemented.

It also includes some structural changes, as distributed datasets by design need to have access to the executor, so they get a synchronous executor passed to `initialize`.


sk1p:
Remaining CI failures appear to be because of an too-old sqlite3 version bundled with Python. Currently re-writing using transactions and explicit checks (`UPSERT` is so nice, and you only realize it when you don't have it available :grin:).

sk1p:
Hmm, Windows is still not too happy about the `rename()`, even though the mmap and file handle should be closed at that point. Need to dig deeper...

uellue:
Wow, this looks very clean and compact. 👍 Maybe we can see together about Windows next week.

uellue:
The PermissionError is triggered if a Dask.distributed client has been started before the tests are executed.

```
import dask.distributed as dd

def test_start_client():
    cluster = dd.LocalCluster()
    client = dd.Client(cluster)  # This triggers the subsequent error
```

Starting the cluster without attaching a client, starting a multiprocessing pool or starting an asyncio event loop doesn't trigger it. The value of the `set_as_default kwarg` of `dd.Client` doesn't change the behavior.

uellue:
Looking deeper into what `dd.Client()` does: This triggers *one PermissionError in the next test*:

```
def test_start_client():
    cluster = dd.LocalCluster()
    raise Exception("outta here")
```

That means if I comment out `test_simple`, it is `test_roi` that fails.

uellue:
...while this one *doesn't* trigger the error:

```
def test_start_client():
    with dd.LocalCluster() as cluster:
        pass
    raise Exception("outta here")
```

... and this one *does trigger the error for both subsequent tests*:

```
def test_start_client():
    with dd.LocalCluster() as cluster:
        client = dd.Client(cluster)
```

sk1p:
See also: https://github.com/numpy/numpy/issues/13510

sk1p:
Merging, thanks @uellue for your help on the Windows-related issue!

PR #476, opened 2019-11-11
Switch to current pywin32 package

uellue:
pypiwin32 is deprecated

This might affect some recent CI failures


sk1p:
Thanks, merging!

PR #477, opened 2019-11-11
Include @woozey in creators because of binning prototype commit

uellue:


uellue:
@woozey Thx for the fix! This is a good opportunity to improve the process to guide people towards editing the JSON files and using those as the only source of truth, see the last commit.

sk1p:
Hm, maybe the `.gitignore` file could actually ignore the `.rst` files, so they don't get accidentally added to the repo. I guess right now it only serves the purpose of keeping the directory in git? 

Other than that it looks good!

uellue:
> Hm, maybe the `.gitignore` file could actually ignore the `.rst` files, so they don't get accidentally added to the repo.

Done! :-)

sk1p:
Perfect, thanks!

PR #481, opened 2019-11-12
Reduce likelyhood of jumps to wrong position

uellue:
Reduce radius to increase clearance between peaks. This fixes occasional issues in CI.

Note: This tries all correlation patterns, that means including the ones that are not so good
at separating close disks.

## Contributor Checklist:


sk1p:
Thanks, applied! Let's see how this fares in practice. If we still get random failures after this, we may want to think about manually selecting a "snapshot" of parameters that are known to work well with the chosen match templates and their parameters...

PR #482, opened 2019-11-12
CI update, Python 3.8 support

sk1p:
Run travis and appveyor tests on Python 3.8 and fix some linting warnings via https://config.travis-ci.com/explore

**Status**: Currently waiting for binary wheels of dependencies

 * scikit-image: their build depends on `numpy`, but they [removed their pyproject.toml](https://github.com/scikit-image/scikit-image/pull/3914) and with it their declaration of build deps, meaning pip doesn't know to install numpy before building skimage. Also, using tox, I didn't find out yet if it is possible at all to install numpy before the other packages. Work-around: build in separate build environment before running tox.
 * llvmlite: needs special attention for building (custom llvm build etc.)
 * h5py, psutil, scikit-learn: are built from source

Plan: merge this with allowing failures. Disable the work around so we see always the real status.

See also: https://github.com/scikit-image/scikit-image/issues/4261

## Contributor Checklist:

* [ ] I have added/updated documentation for all user-facing changes - may want to update installation docs to include Python 3.8 once the tests are running there.


sk1p:
I just re-ran the CI pipelines and, while [the linux build succeeds](https://travis-ci.org/LiberTEM/LiberTEM/jobs/611439042?utm_medium=notification&utm_source=github_status), the Windows build [fails with a `NotImplementedError`](https://ci.appveyor.com/project/sk1p/libertem/builds/29949367/job/m23ywi9e7lk5ymx7) somewhere deep in tornado... so, we'll need to wait a bit longer for things to stabilize to fully support Python 3.8 in LiberTEM.

uellue:
Somehow the way how DLLs are loaded seems to have changed. `python Scripts\pywin32_postinstall.py -install` is required to run pywin32 with Python 3.8 on Windows: https://github.com/mhammond/pywin32/issues/1431#issuecomment-548584385

uellue:
The `NotImplementedError` still persists.

uellue:
Here's sth on the NotImplementedError: https://stackoverflow.com/questions/58422817/jupyter-notebook-with-python-3-8-notimplementederror

uellue:
What would be the best place to put the event loop tweaking snippet? The error is triggered both when starting the web server and when starting a cluster.

How is this handled in the case of embedding? Apparently this should be executed early before an event loop is started. Should this be in "Tips and tricks"?

sk1p:
Thanks for digging out the issue!

> What would be the best place to put the event loop tweaking snippet? The error is triggered both when starting the web server and when starting a cluster.

Hmm. Maybe we should have a function `setup_eventloop` which is called in the relevant places - which, in case of windows, sets this policy? I thought about the win_tweaks module, but maybe we can find a less hacky solution.

> How is this handled in the case of embedding? Apparently this should be executed early before an event loop is started. Should this be in "Tips and tricks"?

Tips and tricks sounds good for now - maybe later we can have a dedicated page on embedding LiberTEM.

uellue:
Hm, it seems that dask.distributed is suffering from the same problem. Starting `dask-scheduler` on the command line fails in my Python 3.8 environment on Windows.

uellue:
Working on this, I just got one layer deeper. The fix doesn't work for dask.distributed. Here a little reproducer and its result:

```python
import asyncio
import sys
import dask.distributed as dd

if __name__ == '__main__':
    if sys.platform == 'win32':
        asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())

    cli = dd.Client()
``` 
[distributed38.log](https://github.com/LiberTEM/LiberTEM/files/4128032/distributed38.log)

Not setting the event loop policy just lets it crash earlier, apparently.

uellue:
Ok, got one layer further:

```python
import asyncio
import sys

import numpy as np
import dask.distributed as dd
import dask.array as da


if sys.platform == 'win32':
    asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())

if __name__ == '__main__':
    # For context see https://github.com/tornadoweb/tornado/issues/2608

    cli = dd.Client()

    a = da.asarray(np.zeros(3))

    print(a.sum().compute())

    cli.close()
```

Having the event loop configuration outside of `if __name__ == '__main__':` lets spawning worker execute that code block as well, apparently before doing work with tornado that would trigger the error.

I got stuck while getting this to work in LiberTEM or together with ipython. I got the web server to start, but it trips with a different error when starting a local cluster:

```
(libertem38) C:\Users\weber\Documents\LiberTEM\LiberTEM>libertem-server
[2020-01-29 14:22:13,872] INFO [libertem.web.server.main:112] listening on localhost:9000
[2020-01-29 14:22:13,882] DEBUG [asyncio.__init__:59] Using selector: SelectSelector
[2020-01-29 14:22:18,212] INFO [libertem.web.config.get:21] ConfigHandler.get
[2020-01-29 14:22:18,213] INFO [libertem.web.base.log_message:28] message: CONFIG
[2020-01-29 14:22:18,314] INFO [libertem.web.connect.get:20] ConnectHandler.get
[2020-01-29 14:22:18,320] INFO [libertem.web.config.get:21] ConfigHandler.get
[2020-01-29 14:22:18,330] INFO [libertem.web.base.log_message:28] message: CONFIG
tornado.application - ERROR - Uncaught exception PUT /api/config/connection/ (::1)
HTTPServerRequest(protocol='http', host='localhost:9000', method='PUT', uri='/api/config/connection/', version='HTTP/1.1', remote_ip='::1')
  File "C:\Users\weber\.conda\envs\libertem38\lib\site-packages\tornado\web.py", line 1699, in _execute
    result = await result
  File "c:\users\weber\documents\libertem\libertem\src\libertem\web\connect.py", line 51, in put
    sync_executor = await sync_to_async(
  File "c:\users\weber\documents\libertem\libertem\src\libertem\executor\base.py", line 172, in sync_to_async
    return await loop.run_in_executor(pool, fn)
  File "C:\Users\weber\.conda\envs\libertem38\lib\concurrent\futures\thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "c:\users\weber\documents\libertem\libertem\src\libertem\executor\dask.py", line 219, in make_local
    cluster = dd.LocalCluster(**(cluster_kwargs or {}))
  File "C:\Users\weber\.conda\envs\libertem38\lib\site-packages\distributed\deploy\local.py", line 203, in __init__
    super(LocalCluster, self).__init__(
  File "C:\Users\weber\.conda\envs\libertem38\lib\site-packages\distributed\deploy\spec.py", line 246, in __init__
    self._loop_runner = LoopRunner(loop=loop, asynchronous=asynchronous)
  File "C:\Users\weber\.conda\envs\libertem38\lib\site-packages\distributed\utils.py", line 372, in __init__
    current = IOLoop.current()
  File "C:\Users\weber\.conda\envs\libertem38\lib\site-packages\tornado\ioloop.py", line 265, in current
    loop = asyncio.get_event_loop()
  File "C:\Users\weber\.conda\envs\libertem38\lib\asyncio\events.py", line 639, in get_event_loop
    raise RuntimeError('There is no current event loop in thread %r.'
RuntimeError: There is no current event loop in thread 'ThreadPoolExecutor-0_0'.
```

sk1p:
> I got the web server to start, but it trips with a different error when starting a local cluster: [...]

The event loop also needs to be set in the `AsyncAdapter`. So you need to call your event loop setup code after creating the thread, for example in the constructor of `AsyncAdapter`. Something like this could work:

```python
fut = self._pool.submit(your_setup_function)
fut.result()  # may raise an exception, if `your_setup_function` raises one
```

sk1p:
Is it possible that the event loop needs to be set even earlier, before importing dd? I guess we could also wait for https://github.com/dask/distributed/pull/3249 to be merged...

uellue:
> Is it possible that the event loop needs to be set even earlier, before importing dd? I guess we could also wait for [dask/distributed#3249](https://github.com/dask/distributed/pull/3249) to be merged...

Thx for the reference! Yes, we should wait since this will hopefully resolve the issues with Dask. Then we see what is left to do.

sk1p:
We'll need something like this in our dependencies once `distributed` 2.11 is released:

```
distributed >= 2.2.0;python_version<'3.8'
distributed >= 2.11.0;python_version>='3.8'
```

sk1p:
Distributed 2.11 is now released, let's see if this works

sk1p:
Retried this PR to check for #695. No travis-related checks appearing in this PR. Somehow the webhooks seem to be broken?

uellue:
@sk1p What is the status and future plan here? That's a lot of CI failures... 😅 How does it relate to #565? Regarding the 0.5 release, do you think we can include experimental support for Python 3.8, or do we postpone that?

sk1p:
> Regarding the 0.5 release, do you think we can include experimental support for Python 3.8, or do we postpone that?

I think we can advertise experimental Python 3.8 support - I'm now using it in development and it works fine. Not sure how the situation is on Windows, though.

Regarding this PR, I don't think we'll have to merge it - for 0.6 or 0.7 I'd like to finish the move to Azure Pipelines, which mostly means adjusting the release script and no longer using `travis-sphinx`.

uellue:
Hm, we have `"distributed>=2.2.0,<2.9"` because of the weird errors that newer versions of distributed caused. I've just retested with the brand new distributed 2.14, and it still locks up. Technically we can still work with the local context. I am testing on 3.8 now to see if the error exists there, too.

uellue:
Confirmed: Distributed 2.14 locks up with `prototypes/reproducers/test_dask_remote.ipynb` on Python 3.8.

uellue:
> Regarding this PR, I don't think we'll have to merge it

So should we close it, then? :-)

sk1p:
> So should we close it, then? :-)

If we commit to switching to Azure pipelines at the beginning of 0.6, I'd say go ahead and close, otherwise it is useful to have for ad-hoc py38 testing.

uellue:
> If we commit to switching to Azure pipelines at the beginning of 0.6, I'd say go ahead and close, otherwise it is useful to have for ad-hoc py38 testing.

Ok, let's leave it open. I've removed milestone.

sk1p:
Closed in favor of #565

PR #487, opened 2019-11-15
pywin32 venv installation bug was fixed upstream, relax pinning

sk1p:
I left in a `!=226` constraint to prevent accidental installation of this known-broken version.


sk1p:
Looks like it works, merging

PR #495, opened 2019-11-29
Calculate size as int64 to avoid overflows

### LINKED ISSUE #493: Integer overflow for calculating mmap size
When opening large RAW files, the following error is triggered:

image

This doesn't occur with smaller RAW files. This is likely an integer overflow when calculating the size.

Thx @unsimulatable for the report!

uellue:
Include efficient test case for large file using a sparse file

TODO confirm sparsity on Linux

Fixes #493

## Contributor Checklist:

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -129,6 +129,8 @@ def large_raw(tmpdir_factory):
     else:
         with open(filename, 'wb') as f:
             f.truncate(size)
+        stat = os.stat(filename)
+        assert stat.st_blocks == 0
     ds = RawFileDataSet(
         path=str(filename),
         scan_size=shape[:2],
--- a/conftest.py
+++ b/conftest.py
@@ -1,5 +1,6 @@
 import os
 import importlib.util
+import platform
 
 import numpy as np
 import pytest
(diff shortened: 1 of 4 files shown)


sk1p:
> TODO confirm sparsity on Linux

Can be done using either command line utilities:

```
alex@one:/tmp$ truncate --size 1G is_it_sparse
alex@one:/tmp$ du -sh is_it_sparse 
0	is_it_sparse
alex@one:/tmp$ ls -lah is_it_sparse 
-rw-r--r-- 1 alex alex 1.0G Nov 29 12:35 is_it_sparse
alex@one:/tmp$ 
```

Or comparing the `st_size` and `st_blocks` stat fields:

```
>>> import os
>>> os.stat('is_it_sparse').st_blocks
0
>>> os.stat('is_it_sparse').st_size
1073741824
>>> 
```

(`st_blocks * blocksize` should be much smaller than `st_size`)

uellue:
@sk1p Thx! Sparsity is confirmed on Linux.

sk1p:
Looks good, thanks. Should someone run this on a file system with different semantics regarding sparse files, It should still eventually succeed, if only slowly.

Issue #485, opened 2019-11-14
Debug hangs of `Client.close` at the end of test runs


sk1p:
At the end of test runs, we now had quite a few hanging CI jobs, and I could also locally reproduce this just using plain old `pytest tests/`. I dug into this and found out it hangs at the `atexit` function that closes the global client. One thing that came to light is that `Client.close()` of `distributed` doesn't pass on the `timeout` in the synchronous case - if I patch this in, I get the following traceback:

```
Error in atexit._run_exitfuncs:
Traceback (most recent call last):
  File "/home/clausen/.virtualenvs/libertem-debug/lib/python3.7/site-packages/distributed/client.py", line 4556, in _close_global_client
    c.close(timeout=2)
  File "/home/clausen/.virtualenvs/libertem-debug/lib/python3.7/site-packages/distributed/client.py", line 1346, in close
    sync(self.loop, self._close, fast=True, callback_timeout=timeout)
  File "/home/clausen/.virtualenvs/libertem-debug/lib/python3.7/site-packages/distributed/utils.py", line 328, in sync
    raise gen.TimeoutError("timed out after %s s." % (callback_timeout,))
tornado.util.TimeoutError: timed out after 2 s.
```

But now there are a few questions to answer:

 * Why is there a global client at all, where is it created? Did we miss a cleanup somewhere?
   * Idea: breakpoint: when the global `Client` is registered (or just raise an Exception there)
 * Is the timeout-ignoring of `.close()` a feature or a bug?
 * Why is the close function hanging at all? What is it waiting for?
 * Debugging with gdb showed quite a lot of threads, are we leaking somewhere?

sk1p:
client.coroutines contains the following:

```
[<Task finished coro=<Client._handle_report() done, defined at /home/clausen/.virtualenvs/libertem-debug/lib/python3.7/site-packages/distributed/client.py:1151> exception=RuntimeError('generator raised StopIteration')>]
```

sk1p:
Maybe these issues are also caused by our mocking of `time.time`? If there is some timeout functionality that relies on monotony of time, which is running in the background while we do our testing,  we may be breaking that...

sk1p:
A dump of threads running when the hang occurs:

```
[<_MainThread(MainThread, stopped 140206564341568)>, <Thread(PyFFTWCacheThread, started daemon 140206306023168)>,
[...]

PR #496, opened 2019-11-29
Improve API documentation

### LINKED ISSUE #494: Document the return type of analysis classes better
Right now, dealing with the results of an Analysis can be difficult because the contents are "abstracted away". The information to access the contents from an Analysis should be accessible from the Context API documentation.

uellue:
Improve API reference

* Include information on the keys of an analysis result
* Correct formatting issues and a few incorrect return types
    
Closes #494

## Contributor Checklist:

### DIFF
--- a/docs/source/reference/api.rst
+++ b/docs/source/reference/api.rst
@@ -34,3 +34,35 @@ PickFrameJob
 .. automodule:: libertem.job.raw
    :members:
    :special-members: __init__
+
+Analysis API
+~~~~~~~~~~~~
+
+.. automodule:: libertem.analysis.base
+   :members:
+   :undoc-members:
+   :special-members: __init__
+
+.. automodule:: libertem.analysis.com
+   :members:
+   :special-members: __init__
+
+.. automodule:: libertem.analysis.disk
+   :members:
+   :special-members: __init__
+
+.. automodule:: libertem.analysis.ring
+   :members:
+   :special-members: __init__
+
+.. automodule:: libertem.analysis.point
+   :members:
+   :special-members: __init__
+
+.. automodule:: libertem.analysis.sum
+   :members:
+   :special-members: __init__
+
+.. automodule:: libertem.analysis.raw
+   :members:
+   :special-members: __init__
--- a/docs/source/reference/api.rst
+++ b/docs/source/reference/api.rst
@@ -14,7 +14,7 @@ Job API
 ~~~~~~~
 
 .. automodule:: libertem.job.base
-   :members:
+   :members: Job
    :special-members: __init__
 
 MaskContainer and ApplyMaskjob
@@ -25,44 +25,28 @@ highly efficient mask application operations, such as virtual detector, center o
 feature vector calculations. In the future, they might be migrated to the :ref:`user-defined functions` API.
 
 .. automodule:: libertem.job.masks
-   :members:
-   :special-members: __init__
-
-PickFrameJob
-............
-
-.. automodule:: libertem.job.raw
-   :members:
+   :members: MaskContainer
    :special-members: __init__
 
 Analysis API
 ~~~~~~~~~~~~
 
 .. automodule:: libertem.analysis.base
-   :members:
-   :undoc-members:
+   :members: Analysis, AnalysisResult, AnalysisResultSet
+   :special-members: __init__
+
+.. automodule:: libertem.analysis.masks
+   :members: MasksResultSet, SingleMaskResultSet
    :special-members: __init__
 
 .. automodule:: libertem.analysis.com
-   :members:
-   :special-members: __init__
-
-.. automodule:: libertem.analysis.disk
-   :members:
-   :special-members: __init__
-
-.. automodule:: libertem.analysis.ring
-   :members:
-   :special-members: __init__
-
-.. automodule:: libertem.analysis.point
-   :members:
+   :members: COMResultSet
    :special-members: __init__
 
 .. automodule:: libertem.analysis.sum
-   :members:
+   :members: SumResultSet
    :special-members: __init__
 
 .. automodule:: libertem.analysis.raw
-   :members:
+   :members: PickResultSet
    :special-members: __init__
--- a/docs/source/reference/api.rst
+++ b/docs/source/reference/api.rst
@@ -3,6 +3,34 @@ LiberTEM Context API
 
 See :ref:`api documentation` for introduction and complete examples.
 
+Context
+~~~~~~~
+
 .. automodule:: libertem.api
+   :members:
+   :special-members: __init__
+
+Job API
+~~~~~~~
+
+.. automodule:: libertem.job.base
+   :members:
+   :special-members: __init__
+
+MaskContainer and ApplyMaskjob
+..............................
+
+:class:`libertem.job.masks.MaskContainer` and :class:`libertem.job.masks.ApplyMasksJob` allow to implement
+highly efficient mask application operations, such as virtual detector, center of mass or
+feature vector calculations. In the future, they might be migrated to the :ref:`user-defined functions` API.
+
+.. automodule:: libertem.job.masks
+   :members:
+   :special-members: __init__
+
+PickFrameJob
+............
+
+.. automodule:: libertem.job.raw
    :members:
    :special-members: __init__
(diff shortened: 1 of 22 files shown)


docs/source/reference/api.rst
```diff
@@ -3,6 +3,66 @@ LiberTEM Context API
...
    :special-members: __init__
+
+Job API
+~~~~~~~
+
+.. automodule:: libertem.job.base
+   :members:
+   :special-members: __init__
+
+MaskContainer and ApplyMaskjob
```
sk1p: Typo (was already there in old version): `ApplyMasksJob`.

There should be a note that this part (`Job` and `Analysis`) of the API is actually internal, and subject to change. That was the idea of having the `Context`, to allow changes of the implementation without impacting users.

src/libertem/api.py
```diff
@@ -517,8 +618,9 @@ def map(self, dataset, f, roi=None):
         Returns
         -------
 
-        BufferWrapper:
-            The result of the UDF. Access the underlying numpy array using the `data` attribute.
+        BufferWrapper : libertem.common.buffers.BufferWrapper
+            The result of the UDF. Access the underlying numpy array using the
+            :attr:`~libertem.common.buffers.BufferWrapper.data` property.
```
sk1p: ... maybe we should also note that a `BufferWrapper` can just be passed to `numpy` functions in place of any array (also applies to other `BufferWrapper` mentions)

uellue:
Hi @sk1p, actually, the Analysis and Job classes are exposed to the user, in a way. The methods of the Context create objects of that type, which the user then receive and pass to the Context for execution. My reason for restructuring was to have the documentation of the return type and of the return type when running the analysis in one place.

Perhaps one could resolve this by splitting the documentation into public part which only contains the name and a description of the Analysis, and an internal part with more details? The Analysis class could be an Interface with zero functionality, except of being an input type for run()?

src/libertem/api.py
```diff
@@ -451,6 +543,12 @@ def run(self, job: Union[Job, BaseAnalysis], roi=None):
         ----------
         job
             the job or analysis to run
+
+        Returns
+        -------
+        result : numpy.ndarray or libertem.analysis.base.AnalysisResultSet
+            Running a Job returns a :class:`numpy.ndarray`, running
+            an Analysis returns a :class:`libertem.analysis.base.AnalysisResultSet`
```
sk1p: Maybe also add: "see the matching `create_*_analysis` function for documentation of the specific `AnalysisResultSet` fields"

sk1p:
> My reason for restructuring was to have the documentation of the return type and of the return type when running the analysis in one place.

Okay, sounds good.

> the Analysis and Job classes are exposed to the user, in a way. The methods of the Context create objects of that type, which the user then receive and pass to the Context for execution.

Yes, but they are (supposed to be!) completely opaque: the user doesn't have to (and shouldn't) know anything about them. The only behavior that needs to be documented is that whatever is returned from `create_*_{job,analysis}` should be passed on to `run`, and how the result of calling `run` will look like. The best comparison would be a `pthread_t`, which is the thread id of POSIX threads. The POSIX standard allows the implementation to choose any type for `pthread_t` - so it can be an integer handle, or a `struct` with any amount of fields. The user doesn't have to care - they just pass it into functions as a handle.

> Perhaps one could resolve this by splitting the documentation into public part which only contains the name and a description of the Analysis, and an internal part with more details?

Sounds reasonable. How is this done using Sphinx? Isn't there a disambiguation problem if we have two API docs for a singe class?

> The Analysis class could be an Interface with zero functionality, except of being an input type for run()?

Ok, so we would have:

```python
class Analysis:
    pass

class BaseAnalysis(Analysis):
    # ... actual base functionality here ...

class FunkyAnalysis(BaseAnalysis):
    # implementation of whatever methods are needed

class Context:
    def run(self, job: Union[Job, Analysis], ...):
        # ...
```

Right?

Maybe we are focusing on the wrong class here, though: IMHO the most important thing here is that the user understands the structure of the concrete `AnalysisResultSet` - the analysis itself is from a user's perspective just a handle.

Maybe we could, for documentation purposes, add new subclasses of `AnalysisResultSet` which document their fields and example usage. Something like:

```python
class MaskResultSet(AnalysisResultSet):
    """
    Results of mask application. For each input mask `i`, it contains a channel (or attribute) "mask_{i}"
    which has the shape of the navigation dimension of the input dataset.
    [...]

    TODO: also document case of complex numbers with `_real`, `_imag`, ... suffixes
    TODO: example(s)
    TODO: also document dict- and list-like indexing (res[0] is the 0th mask etc.) 
    """
    pass  # if needed, can have convenience methods here for the specific result type
```

We can then change the `create_*_analysis` docstring to something like: "use `Context.run` on the resulting object to execute the analysis. `Context.run` will then return a `MaskResultSet` [...]", with a reference to the docs for `MaskResultSet`. Maybe also having a short version in the `create_*_analysis` docstring (in this case something like your patch already adds).

uellue:
Thx @sk1p for the comprehensive feedback and ideas! This sounds like a good solution. I'll be working on it. :-)

docs/source/reference/api.rst
uellue: Ok, the Job is now not included anymore. The MaskContainer is already in use in the SparseCorrelationUDF and it seems like a generally useful component. Do you think we'll have to change much and warn the user about it or can we publish it as part of the API?

uellue: After consideration, I've moved it back to the Internal part since it is not used in the API. Is there a reason to have the ApplyMasksJob documented at all?

sk1p: I would need to have a closer look at `MaskContainer` again, but for now I would prefer to have it internal. Maybe we should revisit as a general tool to use, once the parameters are properly scattered and shared in shm?

I agree that having `Job` impls documented is not really useful, also because the interface is very simple.

src/libertem/analysis/masks.py
```diff
@@ -41,6 +41,75 @@ def get_preset_dtype(self):
         return self.parameters.get('dtype', None)
 
 
+class MasksResultSet(AnalysisResultSet):
+    """
+    Running a :class:`MasksAnalysis` vial :meth:`libertem.api.Context.run` on a dataset
```
sk1p: Typo: via

src/libertem/analysis/masks.py
```diff
@@ -41,6 +41,75 @@ def get_preset_dtype(self):
...
+
+    If any of the masks or the dataset contain complex numbers, the regular mask results
+    attributes carry the absolute value of the results, and additional attributes with real
+    part, imaginary part, phase and full complex result are available.
+
+    .. versionadded:: 0.3.0.dev0
+
+    Attributes
+    ----------
+    mask_0, mask_1, ..., mask_<n> : libertem.analysis.base.AnalysisResult
```
sk1p: Maybe also document `__getitem__`: `result[0] == mask_0` etc. - maybe it should also be augmented by an iteration interface for looping over mask results, something like this (in the base class):

```
def __iter__(self):
    return iter(self.results)
```

I would be fine with doing this in the next iteration of the docs.

src/libertem/api.py
```diff
@@ -225,9 +265,9 @@ def create_radial_fourier_analysis(self, dataset, cx: float = None, cy: float =
             ri: float = None, ro: float = None, n_bins: int = None, max_order: int = None,
             use_sparse: bool = None):
         """
-        Calculate the Fourier transform of rings around the center.
+        Create an Analysis that calculates the Fourier transform of rings around the center.
```
sk1p: Good catch on the language use, here and the other instances!

docs/source/reference/api.rst
uellue: Ok, then I'll leave it as in the last commit where I made `MaskContainer` internal! From `Job` only the abstract base class remains in the API documentation because it is the return type of some `Context` methods.

src/libertem/analysis/masks.py
uellue: Done! Implementing the iterator method was the least of the issue...

src/libertem/api.py
uellue: 😃

sk1p:
The CI failure was caused by #485 which should be fixed in current master. Will merge now, thanks a lot!

PR #497, opened 2019-12-04
Support for stacks of DM3/DM4 files

### LINKED ISSUE #291: Support for DM3/DM4
Support DM3 files in order to open this dataset: https://zenodo.org/record/2566137

See also #254 
The simulated data is in npz format #222 

CC:
@joverbee, thank you for publishing the data on Zenodo!

sk1p:
Closes #291. 

* [x] I have added/updated test cases (with caveat that they don't run in CI because we don't host test data yet)


sk1p:
@uellue let me know if you'd like to review - otherwise I would say this is ready to merge. The test coverage drop is because we are still lacking infrastructure for testing with real input data (and IMHO everything else is snake oil)

src/libertem/io/dataset/dm.py
```diff
@@ -0,0 +1,272 @@
...
+    This DataSet is not supported in the GUI yet, as the file dialog needs to be
+    updated to properly handle opening series.
+
+    Note
+    ----
+    Single-file 4D DM files are not yet supported. The use-case would be
+    to read DM4 files from the conversion of K2 data, but those data sets
+    are actually transposed (nav/sig are swapped), so it is very slow to
+    read single diffraction patterns / frames. Please let us know if you have
+    a use-case for single-file 4D DM files and we may add support!
```
uellue: That means the data would have to be transposed back into the usual shape, which is slow, or algorithms would have to be adapted to work directly on transposed data. As an example, applying a mask in the conventional layout corresponds to calculating a weighted sum frame along the navigation dimension in the transposed layout.

Since the transposed layout corresponds to a TEM tilt series, support for transposed 4D STEM data could have more general applications beyond supporting 4D DM4 files. Please contact us if you have a use-case for single-file 4D DM files or other applications that process stacks of TEM files, and we may add support!

sk1p:
We discussed if we should support 4D single-file formats; I would prefer to do this in a separate step. Supporting both in a single DataSet can get a bit messy, as the details are a bit different.

Issue #503, opened 2019-12-09
Improve API documentation for included UDF impls


sk1p:
This issue is about the generated API docs for our UDFs: https://libertem.github.io/LiberTEM/reference/application.html - we haven't been strict enough for previously merged UDFs with regards to documentation, so there is some room for improvement.

All UDFs should properly document their params - the FEM UDF is almost there, it just needs to have proper docstring formatting applied. Even idf the UDF have "wrapper functions", the UDFs themselves should be documented.

The UDFs should properly document the structure of their results - I'm not sure what the best way is here, but at minimum there should be a textual description in the `__init__` docstring. It could also reference the `get_result_buffers` method, if it has a good docstring. `blobfinder.run_refine` is an example of an attempt - but the large code-formatted box currently causes some bad formatting issues.

The API docs should only include methods that are actually interesting for the user - those just inherited from the base class mostly are not, and `get_task_data` etc. should be only included if there is something interesting to document - for example, if the docstring actually contains a description of data that can be shared between `process_frame` calls.

WIP PR #507, still TODO:

 * [x] I removed `inherited-members` from `libertem.udf.blobfinder`, as it means that many of the internal methods of the base class. Sphinx can do exclude methods via `exclude-members`, but that would have to be a very long list, and needs to be updated when UDF-Internals are updated. Can this be converted into a whitelist instead, somehow?
 * [x] The applications reference page has gotten quite long now, should maybe broken into subsections? The formatting also doesn't make it easy to scan IMHO
 * [x] Need to fix formatting for `run_refine` function parameters - describe the result keys in a different way
 * [x] Make sure all UDFs / applications are documented in the reference section!
   * [x] Crystallinity map
   * [x] Clustering UDF (experimental)
 * [x] Maybe we should put the links from the application "topic" pages to the reference to the top, instead at the bottom, to make navigation easier

sk1p:
I'm now more happy with the UDF docs than before, I'm going to close this now. I think we can do some more incremental improvements before the 0.4 release

PR #507, opened 2019-12-11
Improve documentation for our UDF implementations, refs #503

sk1p:
Add some `__init__` methods with docstrings, which dramatically improves
use from an interactive shell (`help()` or `MyUDF?`, tab-completion in
jupyter, ...) and IDEs. Fix some types in docstrings, add examples, ...

See #503 for items that still need to be done

## Contributor Checklist:

### DIFF
--- a/src/libertem/udf/blobfinder/correlation.py
+++ b/src/libertem/udf/blobfinder/correlation.py
@@ -223,6 +223,11 @@ class CorrelationUDF(UDF):
         super().__init__(peaks=np.round(peaks).astype(int), *args, **kwargs)
 
     def get_result_buffers(self):
+        '''
+        The common buffers for all correlation methods are :code:`centers`,
+        :code:`refineds`, :code:`peak_values`, :code:`peak_elevations`. See
+        source code for details of the buffer declaration.
+        '''
         num_disks = len(self.params.peaks)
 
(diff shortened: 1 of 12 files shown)


sk1p:
@uellue I pushed a proposed structure change of the reference docs (done for amorphous stuff as an example). This makes the application reference pages separate per topic, but still under the main reference section. 

http://localhost:8008/reference/application.html is then a TOC page that links to the separate pages.

What do you think?

src/libertem/udf/blobfinder/correlation.py
```diff
@@ -495,12 +519,34 @@ def postprocess(self):
...
+    Returns
+    -------
+    sum_result : numpy.ndarray
+        Log-scaled sum frame of the dataset
+    centers, refineds, peak_values, peak_elevations : libertem.common.buffers.BufferWrapper
+        See :meth:`CorrelationUDF.get_result_buffers` for details.
+    peaks : numpy.ndarray
+        List of found peaks
+    """
+    # FIXME use ROI for sum!
```
sk1p: Noticed this FIXME: you can now pass the `roi` to `Context.run` as the sum is implemented as a UDF! Also, if possible, the parameters should be documented

Issue #370, opened 2019-06-18
TimeoutError in CI resolved by nannynanny()?


uellue:
The CI TimeoutError phenomenon hasn't occurred since we merged the PR #360 with nannynanny() waking up the loop regularly. Should we perhaps undo all the other changes, namely pinning dask.distributed to a lower version:
https://github.com/LiberTEM/LiberTEM/blob/8a93f7bc451637506a25b2466ad5ae51a08d5abb/setup.py#L136

and setting the thread count to one
https://github.com/LiberTEM/LiberTEM/blob/8a93f7bc451637506a25b2466ad5ae51a08d5abb/src/libertem/executor/base.py#L120

and see where that takes us?

sk1p:
I would prefer to do that in 0.3 (or 0.2.1 or whatever we decide to release next)

sk1p:
CI has been running smoothly for the last time, with exception of the time mocking anomaly, which should now be fixed.

uellue:
What about the ThreadPoolExecutor? I'll do a PR to change it back, to see what that does.

Issue #186, opened 2018-12-05
Scheduling of background and high-performance tasks


uellue:
LiberTEM can open many data sets in parallel. Some of the data might be in the file system cache and/or on fast local storage. Other data might be in other locations with far less bandwidth. Even a 10 GbE NAS or a regular SSD can be regarded as very slow from a LiberTEM perspective.

Right now, tasks are processed on a "first come, first serve" basis by one worker process per CPU core. This can be inefficient in such a heterogeneous scenario because the IO-bound slow process is waiting for a long list of high-performance tasks or vice-versa, although they can be efficiently processed in parallel by the operating system. 

Proposed solution (discussion with @sk1p): Start a swarm of workers with one worker per core individually for each data set. Concurrency between data sets is then handled by the operating system. Disadvantage: Two high-performance data sets would interfere with each other.

Perhaps distinguish by data set source, i.e. a swarm of workers per file system, not data set?

uellue:
Maybe include checks if a block is in the file system cache? http://man7.org/linux/man-pages/man1/fincore.1.html For example, have dedicated workers for cached files that will not load IO, but heavily load memory and CPU.

For Windows I couldn't find an equivalent to fincore yet.

sk1p:
We can use the `dask.distributed` priority system, instead of having more workers. This already works well for file browsing etc. 

The question then is: can we extend that to more "fuzzy" priorities? Like you say, we can determine if a dataset is in fs cache, or how much of it is in fs cache, but can we somehow guarantee fairness for datasets? Like: one dataset is in cache, it has higher priority, but another one is doing slower I/O. The job for the slow dataset should have at least one worker running on each node.

Maybe we should defer this one until after we have a real cluster running, anything else is just a thought experiment or guessing...

uellue:
@sk1p 👍 for testing! Actually, the point here is that the IO-heavy and CPU-heavy tasks can run efficiently in _parallel_., using hyperthreading and the scheduling of the operating system. What about one worker per CPU for cached data, plus a single one for each slow data source? One or two workers should even saturate a 10GbE network connection or a SATA SSD.

sk1p:
[...]

PR #513, opened 2019-12-12
Undo 55c8160305b1a42f154151d1a8e2a5bda562380e to see if this is still…

uellue:
… necessary

See #370

## Contributor Checklist:

### DIFF
--- a/src/libertem/executor/base.py
+++ b/src/libertem/executor/base.py
@@ -203,7 +203,7 @@ class AsyncAdapter(AsyncJobExecutor):
         converted to async and executed in a separate thread.
         """
         self._wrapped = wrapped
-        self._pool = concurrent.futures.ThreadPoolExecutor(1)
+        self._pool = concurrent.futures.ThreadPoolExecutor(4)
 
     def ensure_sync(self):
         return self._wrapped


sk1p:
Hmm, we need to be careful with this change - if I remember correctly, having more than one thread means that some resources will be created per thread by dask.distributed - need to look out for debug output in a stress test, for example when running multiple analyses at the same time via the web API (try for example picking while another, or multiple other, analyses are running).

uellue:
Yes, it should not go into 0.3. We can perhaps merge it after the release and stress-test it a bit in the real world? At least on my system, running parallel analyses is no issue. The file browsing behavior https://github.com/LiberTEM/LiberTEM/issues/186#issuecomment-534518019 seems to be improved by this change, at least on my system.

By the way, is there a reason to limit this to four and not leave it at the default? I couldn't notice much of a difference on my system when using it in an IO-limited scenario. It is really hard to judge and non-deterministic.

uellue:
Ok, CI answered the question about the default thread count.

sk1p:
> Ok, CI answered the question about the default thread count.

:+1: 

For reference: https://travis-ci.org/LiberTEM/LiberTEM/jobs/624223065 - my guess would be that dask distributed creates a new client, or at least a new connection, per thread. So it uses more fds than usually, and trips the respective test.

> I couldn't notice much of a difference on my system when using it in an IO-limited scenario. It is really hard to judge and non-deterministic.

Because the threading in this case is not introduced for performance reasons, but rather to isolate the asyncio loop.

sk1p:
> Because the threading in this case is not introduced for performance reasons, but rather to isolate the asyncio loop.

Closing the PR for this reason, having only 1 thread here makes sense.

PR #548, opened 2020-01-14
Blobfinder fixes; changelog update

uellue:
* Allow running blobfinder tests with NUMBA_JIT_DISABLED
* Fixes and test cases for the bugs that were uncovered in the process

* Managing changelog entries as a minor drive-by change

@sk1p it was really worth working on the NUMBA_JIT_DISABLED test capability.  Numba doesn't do bounds checks just as C, and testing with JIT disabled catches bugs that access the wrong memory.

## Contributor Checklist:

### DIFF
--- a/tests/udf/test_blobfinder.py
+++ b/tests/udf/test_blobfinder.py
@@ -40,8 +40,7 @@ def test_refinement():
     assert (rx < x) and (rx > (x - 1))
 
 
-# FIXME: numba coverage disabled because of to_fixed_tuple issue:
-# @pytest.mark.with_numba
+@pytest.mark.with_numba
 def test_smoke(lt_ctx):
     """
     just check if the analysis runs without throwing exceptions:
@@ -124,6 +123,9 @@ def test_run_refine_fastmatch(lt_ctx):
     b = np.array([0., 29.19]) + np.random.uniform(-1, 1, size=2)
     indices = np.mgrid[-2:3, -2:3]
     indices = np.concatenate(indices.T)
+
+    drop = np.random.choice([True, False], size=len(indices), p=[0.9, 0.1])
+    indices = indices[drop]
 
     radius = 10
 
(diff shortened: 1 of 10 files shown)


uellue:
@skip About the "too many parameters" codeclimate issue: Do you have an idea how to restructure this or should we accept it as-is? This function is mostly used for generating test data.

sk1p:
> Do you have an idea how to restructure this

Hm, a possibility is the "builder" pattern, where build up the parameters over a bunch of method calls, with a final "build" step. It is mostly useful if you have many parameters, many of them being optional. An example usage would be:

```
frame = CbedFrameBuilder(128, 128)\
    .setVectors(zero=zero, a=a, b=b)\
    .setIndices(indices)\
    .setRadius(5).get()
```

In Python it can be a bit awkward, I know this mostly from Java APIs. A downside is also that you then have to write this builder class... As an alternative, one could aggregate parameters that are always/mostly used together into objects, for example using named tuples:

```
MyVectorsType = namedtuple('MyVectorsType', ['zero', 'a', 'b'])

def cbed_frame(shape, vectors, indices, radius, all_equal):
    ...
    frame_peaks(zero=vectors.zero, a=vectors.a, ...)
    ...
```

The downside being that you need to create and import all these auxiliary types...

uellue:
@sk1p  Thx for the suggestions! All things considered, 9 parameters is the simpler solution here IMO. The builder pattern looks good if this was part of a major API. :-)

tests/udf/test_blobfinder.py
```diff
@@ -322,6 +324,57 @@ def test_run_refine_fullframe(lt_ctx):
...
+    template = match_pattern.get_template((2*crop_size, 2*crop_size))
+    nbytes = template.nbytes
+
+    udf_cs = (
+        blobfinder.FastCorrelationUDF,
+        blobfinder.FullFrameCorrelationUDF,
+    )
+
+    for udf_c in udf_cs:
+        for limit in (
```
sk1p: Generally, I would prefer to use pytest's parametrization instead of for-loops, because it allows to test the parameters independently and not abort after the first failure. That may give you a better picture what works and what is actually failing in the test case.

It may be difficult in this case, as the limit depends on local variables, but at least `udf_cs` could be [parametrize'd](https://docs.pytest.org/en/latest/parametrize.html).

Other than this, the PR looks good and ready to merge!

uellue: Ok, done! :-)

Issue #549, opened 2020-01-16
Remove deprecated Job functionality


uellue:
All currently implemented Job functionality can, in principle, be expressed through the UDF API. By only having one interface we can reduce the complexity of LiberTEM. Furthermore, the ApplyMasksJob functionality can be useful as a base class and example for other UDFs, and dtype support should be added to UDFs in any case.

Steps:

- [x] Implement basic ApplyMasksJob functionality through an UDF PR #550 
- [x] Implement basic PickFrameJob functionality through an UDF #550
- [x] Port dtype support to UDF interface PR #550 
- [x] Copy and adapt all Job-related tests for the new UDFs #550
        --> This should be covered by the existing UDF tests, tests for the new UDFs and tests through the Analyses
- [ ] Benchmarking, profiling and optimization to make sure the new system is at least on par performance-wise.
     --> This requires more systematic tests under controlled conditions. In particular the impact of mmap should be tested on madmax nodes since they are more sensitive to memory transfer bottlenecks than moellenstedt.
- [x] Specific performance regression: Pick mode on HDF5. Discussion with @sk1p: The ROI implementation of HDF5 datasets should be improved. Fix committed in #567 
- [x] Port MasksAnalysis to new UDF #550 
- [x] Port remaining Analysis classes to the new UDFs #550 
- [x] Investigate if there are any applications that could be implemented through the Job interface, but not through UDFs.
- [x] Update examples #576 
- [x] Update documentation, including a migration guide, depreciation warnings and a schedule for retirement of the Job API #577 
- [x] Trigger depreciation warnings when using the Job API #577 
- [ ] Remove Job API after the depreciation period, possibly with 1.0 release.

uellue:
In a quick test, the UDF version is significantly faster than the Job version. As a next step, perhaps we should re-run the big cluster benchmark to see if we still hit the previous numbers or even exceed them?

![image](https://user-images.githubusercontent.com/25689052/72590424-e9e1eb80-38fd-11ea-856c-65eb4e696de5.png)

![image](https://user-images.githubusercontent.com/25689052/72590465-05e58d00-38fe-11ea-81d8-0af508c9e7b7.png)

uellue:
Testing in the GUI, it feels as if the JOB version is somewhat faster (10-20 %) for single mask analysis for MIB and K2. This should be verified under more controlled test conditions.

uellue:
[...]

PR #550, opened 2020-01-16
Implement the functionality of ApplyMasksJob with an UDF

uellue:
Start work towards retiring the Job API #549 

## Contributor Checklist:

### DIFF
--- a/docs/source/udf/advanced.rst
+++ b/docs/source/udf/advanced.rst
@@ -331,11 +331,11 @@ dimension as it appears in processing:
 
 .. _udf dtype:
 
-Dtype support
+dtype support
 -------------
 
 UDFs can override :meth:`~libertem.udf.UDF.get_preferred_input_dtype` to
-indicate a "lowest denominator" compatible dtype. The actual input dtype is
+indicate a "lowest common denominator" compatible dtype. The actual input dtype is
 determined by combining the indicated preferred dtype with the input dataset's
 native dtype using :func:`numpy.result_type`. The default preferred dtype is
 :attr:`numpy.float32`. Returning :code:`bool` will usually switch to the
--- a/docs/source/udf/advanced.rst
+++ b/docs/source/udf/advanced.rst
@@ -335,11 +335,13 @@ dtype support
 -------------
 
 UDFs can override :meth:`~libertem.udf.UDF.get_preferred_input_dtype` to
-indicate a "lowest common denominator" compatible dtype. The actual input dtype is
-determined by combining the indicated preferred dtype with the input dataset's
-native dtype using :func:`numpy.result_type`. The default preferred dtype is
-:attr:`numpy.float32`. Returning :code:`bool` will usually switch to the
-dataset's native dtype.
+indicate a "lowest common denominator" compatible dtype. The actual input dtype
+is determined by combining the indicated preferred dtype with the input
+dataset's native dtype using :func:`numpy.result_type`. The default preferred
+dtype is :attr:`numpy.float32`. Returning :attr:`UDF.USE_NATIVE_DTYPE`, which is
+currently identical to :code:`numpy.bool`, will switch to the dataset's native
+dtype since :code:`numpy.bool` behaves as a neutral element in
+:func:`numpy.result_type`.
 
 If an UDF requires a specific dtype rather than only preferring it, it should
 override this method and additionally check the actual input type, throw an
--- a/docs/source/udf/advanced.rst
+++ b/docs/source/udf/advanced.rst
@@ -276,6 +276,7 @@ available as properties of the :attr:`libertem.udf.UDF.meta` attribute of type
 Common applications include allocating buffers with a :code:`dtype` or shape
 that matches the dataset or partition via
 :attr:`libertem.udf.UDFMeta.dataset_dtype`,
+:attr:`libertem.udf.UDFMeta.input_dtype`,
 :attr:`libertem.udf.UDFMeta.dataset_shape` and
 :attr:`libertem.udf.UDFMeta.partition_shape`.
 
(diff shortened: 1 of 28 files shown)


uellue:
Perhaps we could increase the number of allowed parameters in codeclimate? The current limit feels tight and a fix to work with less parameters is not always the better choice, the way it seems.

src/libertem/udf/base.py
```diff
@@ -489,6 +505,32 @@ def merge(self, dest: Dict[str, np.array], src: Dict[str, np.array]):
...
+        or wider input data would switch from :code:`float32` to :code:`float64`, and
+        complex data in the dataset will switch the input dtype kind to :code:`complex`,
+        following the NumPy casting rules.
+
+        In case your UDF only works with specific input dtypes, it should throw
+        an error or warning if incompatible dtypes are used, and/or implement a
+        meaningful conversion in your UDF's :code:`process_<...>` routine.
+
+        If you prefer to always use the dataset's native dtype instead of
+        floats, you can override this method to return :code:`bool`.
```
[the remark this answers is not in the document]
uellue: > Hm, is this a behavior of `np.result_type`?

Yes. :-) `bool` behaves as a neutral element for dtypes in `numpy.result_type()`. I agree that this is pretty opaque. I'll add an explicit marker value which is identical to bool.

src/libertem/api.py
```diff
@@ -195,6 +195,10 @@ def create_mask_analysis(self, factories, dataset, use_sparse=None,
         dtype : numpy.dtype, optional
             Specify the dtype to do the calculation in. Integer dtypes are possible if
             the numpy casting rules allow this for source and mask data.
+        _use_udf : bool, optional
+            This temporary inofficial parameter allows to switch between the Job and the UDF
+            implementations for testing and comparison purposes until the Job API is officially
+            deprecated and the UDF implementation is used exclusively. See :issue:`549`
```
sk1p: Best to not have this parameter in the API, as we only need it for testing and maybe performance comparison. Same can be done by manually creating and tweaking the `Analysis` instance.

uellue: Good point, thx! :-)

src/libertem/udf/masks.py
```diff
@@ -0,0 +1,148 @@
...
+        if self.params.mask_count is None:
+            return len(self.masks)
+        else:
+            return self.params.mask_count
+
+    @property
+    def masks(self):
+        if self._mask_container is None:
+            self._mask_container = self._make_mask_container()
+        return self._mask_container
```
sk1p: Why do we need the mask container as an instance attribute? Why not only have it in task data? Is it somehow used before `get_task_data` is run? I guess for getting the mask count when creating the result buffers?

I'm thinking ahead a bit, how this will fit together with putting static data into shared memory, but I think for that, the API has to be augmented anyways - `get_task_data` can be used for constructing buffers and other mutable structures, so we'll have to add a new method for static shared data.

src/libertem/udf/masks.py
```diff
@@ -0,0 +1,148 @@
...
+        p = self.params
+        return MaskContainer(
+            p.mask_factories, dtype=p.mask_dtype, use_sparse=p.use_sparse, count=p.mask_count
+        )
+
+    def get_task_data(self):
+        ''
+        if self._mask_container is None:
+            self._mask_container = self._make_mask_container()
+        mask_container = self._mask_container
```
sk1p: Could directly access `self.masks` instead, right? No need to duplicate the lazy initialization code.

src/libertem/udf/raw.py
```diff
@@ -0,0 +1,59 @@
...
+    Load raw data from ROI
+
+    This UDF is meant for frame picking with a very small ROI, usually a single frame.
+
+    .. versionadded:: 0.4.0.dev0
+    '''
+    def get_preferred_input_dtype(self):
+        ''
+        # We load the native dtype
+        return bool
```
sk1p: Should use the new constant here

src/libertem/analysis/raw.py
```diff
@@ -38,36 +39,37 @@ class PickResultSet(AnalysisResultSet):
...
-            self.parameters.get('x'),
-            self.parameters.get('y'),
+
+    def get_origin(self):
+        dims = self.dataset.shape.nav.dims
+        if dims not in (1, 2, 3):
+            raise ValueError(
+                "can only handle 1D/2D/3D nav currently, received %s dimensions" % dims
+            )
+        zyx = np.array([
```
sk1p: No need for an actual numpy array here, a simple tuple is enough

src/libertem/analysis/raw.py
```diff
@@ -38,36 +39,37 @@ class PickResultSet(AnalysisResultSet):
...
-            if z is not None:
-                raise ValueError("z must not be specified for 2D nav dataset")
-            origin = (y, x)
-        elif self.dataset.shape.nav.dims == 3:
-            if x is None or y is None or z is None:
-                raise ValueError("need x/y/z to index 3D nav datasets")
-            origin = (z, y, x)
-        else:
-            raise ValueError("cannot operate on datasets with more than 3 nav dims")
+        if (None in keep) or not all((d is None for d in drop)):
```
sk1p: Nitpick: no need for the extra parenthesis in `all`, something like `all(d is None for d in drop)` works, too.

src/libertem/udf/masks.py
uellue: Yes, I need the mask count and dtype to specify the result buffers. I need this both on the workers and on the central node.

Yes, a static MaskContainer could be very useful.

src/libertem/udf/masks.py
uellue: Ok, per our discussion I'll just use the property. :-)

src/libertem/udf/masks.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+class ApplyMasksUDF(UDF):
+    '''
+    Apply masks to signals/frames in the dataset.
+
+    .. versionadded:: 0.4.0.dev0
+    '''
+    def __init__(self, mask_factories, use_torch=True, use_sparse=None, mask_count=None,
+                mask_dtype=None, preferred_dtype=None, *args, **kwargs):
```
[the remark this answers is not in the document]
uellue: No need, removed! :-)

docs/source/udf/advanced.rst
```diff
@@ -276,6 +276,7 @@ available as properties of the :attr:`libertem.udf.UDF.meta` attribute of type
 Common applications include allocating buffers with a :code:`dtype` or shape
```
sk1p: In this file, there are still some references to the old location of the `MaskContainer`

sk1p:
Merged, thank you!

PR #551, opened 2020-01-17
Enable memory mapping for UDFs

uellue:
Is there a reason against using this?

## Contributor Checklist:


sk1p:
There shouldn't be - I think I postponed this until we have some config/control interface for UDFs. It shouldn't have any negative consequences, so I'm merging this. With `process_partition`, one might even see some improvements as computation and I/O should overlap better.

uellue:
Perhaps setting the target_size parameter appropriately could help?

uellue:
This could also be a function of UDFs, specifying their preferred tile size.

PR #558, opened 2020-01-21
Add profiling instructions to docs, fixes #552

### LINKED ISSUE #552: Document UDF profiling
First shot: simple profiling using `line_profiler`. Later add more advanced methods, for example finding the bottleneck of the whole system using `perf` and related tools.

sk1p:
Adds new documentation on how to use `line_profiler` to profile and optimize your UDF.

## Contributor Checklist:

### DIFF
--- a/docs/source/udf/profiling.rst
+++ b/docs/source/udf/profiling.rst
@@ -1,0 +1,180 @@
+.. _`udf profiling`:
+
+.. testsetup:: *
+
+    from libertem.api import Context
+
+
+Profiling UDFs
+==============
+
+To bring your UDF to production, it may be necessary to look into the performance and
+efficiency of your function. There may be some low-hanging fruit to optimize away, but
+they only really become visible once you profile your code. Just looking at code and guessing
+which parts are expensive can be misleading and cause you to spend time optimizing the wrong part
+of your code. Always measure!
+
+Profiling means instrumenting the execution of a program and finding out which parts
+of your code use how much of a given resource (for example, CPU time, or memory). 
+
+Prerequisite: inline executor
+-----------------------------
+
+By default, your UDF will be run in a multi-process or multi-threaded environment. This makes
+profiling a challenge. But as most of the time, improving the execution time of your UDF in a
+single-threaded executor also improves the respective multi-process execution time, we can run
+the UDF in a single thread to do profiling measurements.
+
+LiberTEM provides what it calls :class:`~libertem.executor.inline.InlineJobExecutor`,
+which executes your UDF in a single thread, thus allowing the use of default Python profiling
+mechanisms. 
+
+To use it, pass it to the :class:`~libertem.api.Context` on construction:
+
+.. testcode::
+   
+   from libertem.executor.inline import InlineJobExecutor
+   ctx = Context(executor=InlineJobExecutor())
+
+Then, you can continue as usual, loading data, executing your UDF, etc.
+
+Line profiling using `line_profiler`
+------------------------------------------
+
+Profiling comes in different forms, some will show you how much time each function
+of a whole program took, others can give you a line-by-line overview for functions
+you are interested in, like `line_profiler`. For most UDFs this will be sufficient
+for performance analysis.
+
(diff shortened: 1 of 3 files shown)


docs/source/udf/profiling.rst
```diff
@@ -0,0 +1,180 @@
...
+
+
+Profiling UDFs
+==============
+
+To bring your UDF to production, it may be necessary to look into the performance and
+efficiency of your function. There may be some low-hanging fruit to optimize away, but
+they only really become visible once you profile your code. Just looking at code and guessing
+which parts are expensive can be misleading and cause you to spend time optimizing the wrong part
+of your code. Always measure!
```
uellue: Amen! :-)

docs/source/udf/profiling.rst
```diff
@@ -0,0 +1,180 @@
...
+
+Profiling means instrumenting the execution of a program and finding out which parts
+of your code use how much of a given resource (for example, CPU time, or memory). 
+
+Prerequisite: inline executor
```
uellue: Inline Executor, capitalization? Not sure on this one!

docs/source/udf/profiling.rst
```diff
@@ -0,0 +1,180 @@
...
+-----------------------------
+
+By default, your UDF will be run in a multi-process or multi-threaded environment. This makes
+profiling a challenge. But as most of the time, improving the execution time of your UDF in a
```
uellue: ...since the usual profiling tools will only capture information on the main process that performs the final reduction and not the processes that do most of the work.

docs/source/udf/profiling.rst
```diff
@@ -0,0 +1,180 @@
...
+single-threaded executor also improves the respective multi-process execution time, we can run
```
uellue: This sentence should logically come after the next IMO.

docs/source/udf/profiling.rst
```diff
@@ -0,0 +1,180 @@
...
+the UDF in a single thread to do profiling measurements.
+
+LiberTEM provides what it calls :class:`~libertem.executor.inline.InlineJobExecutor`,
+which executes your UDF in a single thread, thus allowing the use of default Python profiling
+mechanisms. 
+
+To use it, pass it to the :class:`~libertem.api.Context` on construction:
```
uellue: This description of how to use the InlineJobExecutor is redundant with https://github.com/LiberTEM/LiberTEM/blob/master/docs/source/debugging.rst#debugging-udfs-or-other-python-code

Perhaps the part on debugging the Python part of LiberTEM can somehow be combined with profiling since they are conceptually very similar?

docs/source/udf/profiling.rst
uellue: What about

In order to use the default Python profiling mechanisms, all parts of the UDF should be executed in the same single thread. LiberTEM provides what it calls :class:`~libertem.executor.inline.InlineJobExecutor` for this purpose. Improving the execution time in a single-threaded executor will almost always improve the multi-process execution time. Nevertheless, you should confirm the impact of changes in a multi-processing or multi-threaded environment as well.

docs/source/udf/profiling.rst
sk1p: I'll use the full name here

docs/source/udf/profiling.rst
sk1p: I tried to merge this and the previous comment into something cohesive, let me know what you think

docs/source/udf/profiling.rst
sk1p: Hmm, I would prefer to not merge, the `InlineJobExecutor` instructions are short enough I think. Also would not want to cross-reference too much, which would require jumping around too much to understand.

uellue: Ok, makes sense!

docs/source/udf/profiling.rst
uellue: Nice, I like it! :-)

uellue:
Thx, merging! .-)

PR #556, opened 2020-01-21
Some small doc fixes

sk1p:
## Contributor Checklist:

* [x] I have added/updated documentation for all user-facing changes

PR #559, opened 2020-01-22
Use the same dtype as _naive_mask_apply

uellue:
Fix potential mismatch from performing calculations with different dtypes, refs #556 

## Contributor Checklist:


sk1p:
Looks good :+1:

PR #564, opened 2020-01-24
Test for both JOB and UDF

uellue:
Make sure that the functionality is tested with the legacy JOB back-end
and the new UDF back-end for the time being

## Contributor Checklist:

### DIFF
--- a/tests/io/test_empad.py
+++ b/tests/io/test_empad.py
@@ -109,9 +109,13 @@ def test_apply_mask_on_empad_job(default_empad, lt_ctx):
     assert np.count_nonzero(results[0]) > 0
 
 
-def test_apply_mask_analysis(default_empad, lt_ctx):
+@pytest.mark.parametrize(
+    'TYPE', ['JOB', 'UDF']
+)
+def test_apply_mask_analysis(default_empad, lt_ctx, TYPE):
     mask = np.ones((128, 128))
     analysis = lt_ctx.create_mask_analysis(factories=[lambda: mask], dataset=default_empad)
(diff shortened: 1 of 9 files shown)


tests/io/test_empad.py
```diff
@@ -124,15 +128,19 @@ def test_sum_analysis(default_empad, lt_ctx):
     assert np.count_nonzero(results[0].raw_data) > 0
 
 
-def test_pick_job(default_empad, lt_ctx):
+def test_pick_job(default_empad, lt_ctx, TYPE):
```
sk1p: Missing the parametrize mark here?

uellue: Actually, added the TYPE in error since this is using the JOB directly. No UDF here!

Nice catch, thx!

uellue:
@sk1p Did you test it with your file collection, by the way? I couldn't run many of the tests on my system.

sk1p:
> @sk1p Did you test it with your file collection, by the way? I couldn't run many of the tests on my system.

I didn't. That's actually something I want to fix - I guess I'll devise an interim solution for #86 ASAP.

uellue:
> > @sk1p Did you test it with your file collection, by the way? I couldn't run many of the tests on my system.
>
> I didn't. That's actually something I want to fix - I guess I'll devise an interim solution for #86 ASAP.

OK, then this PR should probably wait until the new infrastructure is ready.

sk1p:
I'll merge this now and then we'll deal with the fallout later. Thanks!

PR #575, opened 2020-02-04
Misc. DataSet fixes

sk1p:
Fixes test failures for DM and FRMS6

## Contributor Checklist:


sk1p:
Codecov patch coverage issue is expected, as it doesn't include dataset tests.

Issue #431, opened 2019-09-18
Content of .tar.gz asset on GitHub, PyPi and Zenodo


uellue:
The .tar.gz asset on GitHub contains a dump of the entire GitHub repository, while the assets on PyPi and Zenodo contain only the source code of the Python module, as far as I can tell.

Is this intentional? Perhaps the upload on Zenodo could contain the full repository as well since that would be required to re-create the state of the project should GitHub go under in the far, far future?

uellue:
If at all possible, we could dump Issue tracker etc as well? :-)

sk1p:
> The .tar.gz asset on GitHub contains a dump of the entire GitHub repository, while the assets on PyPi and Zenodo contain only the source code of the Python module, as far as I can tell.

PyPI: intentional, as `sdist` is a standard Python format which you can directly install.

Zenodo: agree, should be changed! It could include: wheel, sdist, repo archive. Could be created using `git archive` for example. Or we could download from the GitHub releases thingy via https://github.com/LiberTEM/LiberTEM/archive/{tag}.tar.gz (pro: same files for both, con: creates an ordering between github release and zenodo upload). Should we do this for the next feature release? 

> If at all possible, we could document Issue tracker etc as well on Zenodo? :-)

Yeah, from my experience, I always look for some GitHub badge, for example on documentation sites, to check out the issue tracker, license etc. - is it possible at all to include images in the zenodo README? I would suggest we create a small HTML preamble that includes the most important links and prepend that to the rendered README HTML.

uellue:
> Or we could download from the GitHub releases thingy via [https://github.com/LiberTEM/LiberTEM/archive/{tag}.tar.gz](https://github.com/LiberTEM/LiberTEM/archive/%7Btag%7D.tar.gz) (pro: same files for both, con: creates an ordering between github release and zenodo upload). Should we do this for the next feature release?

Downloading from GitHub sounds good enough for me. :-)

> I would suggest we create a small HTML preamble that includes the most important links and prepend that to the rendered README HTML.

See PR #455!

uellue:
> Should we do this for the next feature release?

Hm, is there a reason to not do it for the next point release?

sk1p:
> Hm, is there a reason to not do it for the next point release?

Let's try to do this for 0.4.0.

PR #582, opened 2020-02-10
zenodo_upload: support for uploading more than two files

sk1p:
First part of #431. Next part will need some restructuring of the release script, switching the ordering of GitHub release and Zenodo upload.

## Contributor Checklist:

### DIFF
--- a/scripts/release
+++ b/scripts/release
@@ -211,11 +211,11 @@ def upload_to_zenodo(verbose, parent, token, sandbox_token):
 
     out = subprocess.check_output([
         "python", join(HERE, "zenodo_upload"),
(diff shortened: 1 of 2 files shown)


sk1p:
> I assume that you tested it on your system? Looks good to me!

Don't have zenodo sandbox credentials - only tested so much that the new `click` setup seems to work fine. I guess we will see once we do the RC?

uellue:
> > I assume that you tested it on your system? Looks good to me!
> 
> Don't have zenodo sandbox credentials - only tested so much that the new `click` setup seems to work fine. I guess we will see once we do the RC?

Ok, let's do it that way! :-) 👍

PR #583, opened 2020-02-10
Upload release repo snapshot to zenodo

### LINKED ISSUE #431: Content of .tar.gz asset on GitHub, PyPi and Zenodo

sk1p:
Download repo snapshot from `/archive/{tag}.tar.gz`, upload that to zenodo and move zenodo upload after GitHub release creation. Fixes #431.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,7 +82,7 @@ jobs:
       python: 3.7
       dist: xenial
       install:
-        - pip install click pygithub requests twine html5lib docutils bibtexparser
+        - pip install -r scripts/requirements.txt
       script:
         - python scripts/release upload --pypi-test-user=libertem_bot --pypi-user=libertem_bot --no-dry-run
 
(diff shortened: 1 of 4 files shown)


sk1p:
> Looks sensible to me. :-) 

Thanks! Will merge.

> Looking at the code, I noticed a lot of similarity and some duplicate structure between the release script and the Zenodo upload script. If we do a bigger overhaul of these scripts at some point, we could perhaps consider having one master script and Python modules for the functionality?

Yes, agreed. I was tempted to do this a few times already actually — as it is, it't very hard to write tests for all the functions, and as we see it's also hard to run in dev (without access to zenodo sandbox etc.) Testing will involve some mocking, but that's better than keeping it "untested"! Btw. I had some good experience with mocking `requests` using the [responses library](https://pypi.org/project/responses/).

> The scripts will be duplicated between all LiberTEM sub-packages, right? We might be coming to a point that our CI and release infrastructure becomes a project in its own right.

Yeah - maybe we can think about extracting it after we have a few sub-packages? Then we should know the concrete requirements.

Issue #591, opened 2020-02-13
Update screenshots in usage documentation


uellue:
The screenshots in https://libertem.github.io/LiberTEM/usage.html are from an ancient release and the GUI offers more and different features now.

Edit: ... and document ROI and stddev map!

See also #588 

- [x] update screenshots of current functionality (#605)
- [ ] document ROI and stddev features

twentyse7en:
I would like to work on this.

twentyse7en:
What exactly should I do ? Update the screenshot with the same example for latest version.

sk1p:
> What exactly should I do ? Update the screenshot with the same example for latest version.

Yes, that's the first part of the issue, the second part would be to write a description for the ROI and stddev GUI features, including screenshots. The first part is easy - for the second part, you will need to really understand what is going on. Don't hesitate to ask questions!

twentyse7en:
For the first part , where can I get the data used in that tutorial ?

For the second part , I would like to know what is going on.

uellue:
Perhaps you can use the Zenodo dataset from @cophus and team instead: https://zenodo.org/record/3592520

Use the "circular", not the "bullseye" dataset. Each of the two HDF5 files is a container with several datasets. Best use "4DSTEM_experiment/data/datacubes/polyAu_4DSTEM/data" since it looks a bit more like a real sample than the other. You can use [HDF5 view](https://www.hdfgroup.org/downloads/hdfview/) to explore the files if you like.

If you mention the source of the dataset in the text, you already finished half of #588 as well. :-)

The second part is pretty straightforward, too. You'll see the additional drop-downs below the left display in the current GUI compared to the old screenshots. In the current version you can choose to view the standard deviation or the sum of the frame, and you can calculate sum and stddev for a limited region of interest on the right side. That's a nice addition to pick mode. Best to play a bit with it first, then we can explain more what it does if you like. :-)

twentyse7en:
Hey @uellue , completed the first part. I did work on those new option.This is what I understand, ROI is introduced.By default it's `all`. New options are `disk` and `rect`. Regions could be moved and resized. We can choose std and average for those region.

When I was switching between `average` and `stddev` image on left was changing. Could you explain whats actually happening when we select stddev.

sk1p:
> When I was switching between average and stddev image on left was changing. Could you explain whats actually happening when we select stddev.

[...]

PR #605, opened 2020-02-17
Update screenshots - First part

twentyse7en:
Updated screenshots in usage documentation.
completed first part.
ISSUE #591

## Contributor Checklist:


twentyse7en:
@uellue can you verify the first part.

sk1p:
The `pick_frame` one being black/empty is actually a bug that is fixed in #606 - if you retry this with LiberTEM from master, you should be able to see diffraction patterns for all possible positions.

uellue:
@twentyse7en Looks good already, many thanks!

For further improvement, perhaps you could make the browser window narrower so that it doesn't show so much blank space left and right? That way the text will hopefully appear a bit larger, too.

Next step would then be to include documentation on stddev and ROI. Was @sk1p 's explanation of stddev  in #591 clear? Essentially, it is a measure of how much a specific pixel changes between frames. A pixel that is always the same brightness will get a value of zero, and a pixel that fluctuates a lot will get a high value.

twentyse7en:
Removing the blank space would be good. I'm not familiar with 4D datasets, I should better read concepts page first and would ask questions if I don't get it.

uellue:
The screenshots look good now! 👍

sk1p:
@twentyse7en :+1: - do you want to do the second part in this PR, or is this ready for merging?

twentyse7en:
It would be better if someone else do second part, I'm having series exam on my college.

sk1p:
@twentyse7en thanks for the PR, merged! :tada:

PR #611, opened 2020-02-18
PickUDF now complains about all-zero ROI

### LINKED ISSUE #609: Handle empty ROI gracefully
To reproduce, run a UDF with `roi=np.zeros(ds.shape.nav, type=bool)`. Currently raises a weird exception in the internals...

Edit: specifically happens with the `PickUDF` - maybe the error in the UDF runner can also be improved?

sk1p:
Previously the internals complained about a weird `extra_shape` of the result buffer.

Fixes #609

## Contributor Checklist:

### DIFF
--- a/src/libertem/udf/raw.py
+++ b/src/libertem/udf/raw.py
@@ -28,6 +28,8 @@ class PickUDF(UDF):
             navsize = np.count_nonzero(self.meta.roi)
         else:
             navsize = np.prod(self.meta.dataset_shape.nav)
+        if navsize == 0:
+            raise ValueError("cannot load data with all-zeros ROI")
         warn_limit = 2**28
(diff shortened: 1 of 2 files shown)


uellue:
How is the behavior of other UDFs? Is there any situation when running an UDF with an all-zero ROI makes sense?

I could just imagine that some sort of automation could come across that with code like "roi = select_by_criterion(map)" that returns an empty set. If UDFs behave sensibly with that, users don't have to handle this special case separately.

As discussed, perhaps emulate NumPy's behavior?

```python
In [1]: import numpy as np

In [2]: a = np.ones((0, 1, 0))

In [3]: a.shape
Out[3]: (0, 1, 0)

In [4]: a
Out[4]: array([], shape=(0, 1, 0), dtype=float64)
```

PR #618, opened 2020-02-18
Improve clustering analysis

### LINKED ISSUE #596: Cluster analysis uses too much memory
IIRC, currently it uses O(N²) memory, where N is the number of frames / scan positions. In practice, this means the libertem server will be OOM killed after a while. In practice, it also depends on the length of the feature vector / max. number of peaks, but that should only be a linear factor.

This can/should be improved by implementing, for example, a connectivity matrix or similar techniques.

uellue:
Closes #596

* Reduce memory footprint and improve speed by using a connectivity matrix
* Use faster ApplyMasksUDF to generate a feature vector
* Don't use delta and binary vector, instead use normalized float values

- [x] Update screenshots and documentation
- [x] Review tests
- [x] Correct PR in changelog
- [x] Update notebook with new file and feature vector method

## Contributor Checklist:


sk1p:
Looks good, merged! :+1:

Issue #616, opened 2020-02-18
Improve usability of HDF5 dataset


sk1p:
The first step when opening a HDF5 dataset via the Python API is getting the `ds_path`. We should provide some convenience functions (like the `_get_datasets` helper in `libertem.io.dataset.hdf5`) to streamline this process. We could also try to automatically choose one ds if the parameter is missing completely, but that could be flaky...

Ideas?

PR #621, opened 2020-02-19
Add `filetype=auto` to `Context.load`

sk1p:
This adds a new method `dataset_info` to our `Context` class, which detects the dataset type and returns some information about it (same as is available via the web GUI). This is for example useful if you are opening an HDF5 file and don't know which `ds_path` to choose.

Would love to get some feedback on this!

Also contains a little fix to sort HDF5 data sets in the diagnostics by size, as that is often the most useful when trying to decide what `ds_path` to open.

If this is useful, I'll add ~tests~, documentation and changelog entry!

Closes: #616 

## Contributor Checklist:

### DIFF
--- a/docs/source/changelog/features/filetype_auto.rst
+++ b/docs/source/changelog/features/filetype_auto.rst
@@ -1,0 +1,5 @@
+[Feature] :code:`filetype="auto"` for Context.load
+==================================================
+
+* :meth:`libertem.api.Context.load` now automatically detects file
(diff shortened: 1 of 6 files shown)


uellue:
Essentially exposing the dataset type guessing logic for scripting?

In a context of scripting, it would actually be great to have some kind of "autoload" function that only requires a path since we don't have to specify the file type or any parameters in many cases. So far I was rarely interested in dataset parameters beyond shape and dtype when developing scripts or UDFs.

PR #640, opened 2020-02-22
Numba-based StdDevUDF

uellue:
This version is ~~slightly~~ about 2x faster compared to the version in #625 for the benign case of signal buffers that fit comfortably into the CPU caches.

For the malign case of signal buffers that don't fit the caches well,
it is about 10x faster since it eliminates large intermediate buffers and
calculates ~~each pixel in one pass~~ the result in cache-friendly blocks and CPU -friendly access patterns.

* Added test for numerical stability.
* Added citation in bibtex

## Contributor Checklist:

### DIFF
--- a/src/libertem/udf/stddev.py
+++ b/src/libertem/udf/stddev.py
@@ -39,34 +39,80 @@ def merge(N0, sum_im0, var0, N1, sum_im1, var1):
 
     shape = sum_im0.shape
 
-    sum_im0 = sum_im0.flatten()
-    sum_im1 = sum_im1.flatten()
-    var0 = var0.flatten()
-    var1 = var1.flatten()
+    sum_im0_ = sum_im0.flatten()
+    sum_im1_ = sum_im1.flatten()
+    sum_im_AB = np.zeros_like(sum_im0_)
+    var0_ = var0.flatten()
+    var1_ = var1.flatten()
+    var_AB = np.zeros_like(var0_)
+
+    le = len(sum_im0_)
+    BLOCKSIZE = 16
+    n_blocks = le // BLOCKSIZE
+
+    def _merge(sum_im0, var0, sum_im1, var1):
+        # compute mean for each partitions
+        mean_A = sum_im0 / N0
+        mean_B = sum_im1 / N1
+
+        # compute mean for joint samples
+        delta = mean_B - mean_A
+        mean = mean_A + (N1 * delta) / N
+
+        # compute sum of images for joint samples
+        sum_im_AB = sum_im0 + sum_im1
+
+        # compute sum of variances for joint samples
+        delta_P = mean_B - mean
+        var_AB = var0 + var1 + (N1 * delta * delta_P)
+        return sum_im_AB, var_AB
+
+    for block in range(n_blocks):
+        for i in range(block*BLOCKSIZE, (block+1)*BLOCKSIZE):
+            sum_im_AB[i], var_AB[i] = _merge(
+                sum_im0_[i], var0_[i], sum_im1_[i], var1_[i]
+            )
+    for i in range(n_blocks*BLOCKSIZE, le):
+        sum_im_AB[i], var_AB[i] = _merge(
+            sum_im0_[i], var0_[i], sum_im1_[i], var1_[i]
+        )
+
+    return N, sum_im_AB.reshape(shape), var_AB.reshape(shape)
 
 
-    # compute mean for each partitions
-    mean_A = sum_im0 / N0
-    mean_B = sum_im1 / N1
-
-    # compute mean for joint samples
-    delta = mean_B - mean_A
-    mean = mean_A + (N1 * delta) / N
-
-    # compute sum of images for joint samples
-    sum_im_AB = sum_im0 + sum_im1
-
-    # compute sum of variances for joint samples
-    delta_P = mean_B - mean
-    var_AB = var0 + var1 + (N1 * delta * delta_P)
-
-    return N, sum_im_AB, var_AB
-
-
+@numba.njit
 def tile_sum_var(tile):
-    s = np.sum(tile, axis=0)
-    v = np.var(tile, axis=0, ddof=tile.shape[0]-1)
-    return s, v
+    BLOCKSIZE = 16
+    shape = tile.shape[1:]
+    N = tile.shape[0]
+    # Flatten signal dimension
+    tile_ = tile.reshape((N, -1))
+    le = tile_.shape[1]
+    s = np.zeros(le, dtype=tile_.dtype)
+    v = np.zeros(le, dtype=tile_.dtype)
+
+    n_blocks = le // BLOCKSIZE
+
+    means = np.zeros(BLOCKSIZE, dtype=tile_.dtype)
+
+    for block in range(n_blocks):
+        offset = block*BLOCKSIZE
+        for j in range(N):
+            for i in range(offset, offset + BLOCKSIZE):
+                s[i] += tile_[j, i]
+        for i in range(BLOCKSIZE):
+            means[i] = s[i + offset] / N
+        for j in range(N):
+            for i in range(offset, offset + BLOCKSIZE):
+                v[i] += (tile_[j, i] - means[i - offset])**2
+    for i in range(n_blocks*BLOCKSIZE, le):
+        for j in range(N):
+            s[i] += tile_[j, i]
+        mean = s[i] / N
+        for j in range(N):
+            v[i] += (tile_[j, i] - mean)**2
+
+    return s.reshape(shape), v.reshape(shape)
 
 
 # Helper function to make sure the frame count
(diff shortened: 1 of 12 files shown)


src/libertem/udf/stddev.py
```diff
@@ -1,56 +1,133 @@
...
 
     Parameters
     ----------
-    p0
-        Contains information about the first partition, including
-        sum of variances, sum of pixels, and number of frames used
+    dest_N : int
+        Number of frames aggregated in dest_sum and dest_varsum
+        The case :code:`dest_N == 0` is handled correctly.
+    dest_sum, dest_varsum
```
sk1p: The shape of `dest_sum` and `dest_varsum` is "a slice of a 'frame'", right? (flattened)

src/libertem/udf/stddev.py
```diff
@@ -1,56 +1,133 @@
...
+        Aggregation buffers, will be updated: sum of pixels, sum of variances
+    src_N, src_sum, src_varsum
+        Information about the second partition
```
sk1p: Hmm, so where is the first partition? Maybe one should describe this as merging a "new" or "additional" partition into an intermediate result in the aggregation buffers?

src/libertem/udf/stddev.py
```diff
@@ -161,26 +256,30 @@ def process_tile(self, tile):
...
+        tile.shape = (N1, -1)
 
-        self.results.sum_frame[:] = compute_merge.sum_im
-        self.results.num_frame[0][key] = compute_merge.N
+        self.results.sum_frame.shape = (-1, )
+        self.results.var.shape = (-1, )
+
+        if N0 == 0:
+            self.results.sum_frame[:] = tile.sum(axis=0)
+            self.results.var[:] = np.var(tile, axis=0, ddof=N1-1)
```
sk1p: Maybe the `ddof` trickery does merit a comment, like in the old version, but also in the buffer description, because here, `self.results.var` is actually `N * Var(X)`

src/libertem/udf/stddev.py
```diff
@@ -161,26 +256,30 @@ def process_tile(self, tile):
...
+            self.results.num_frame[0][key] = N1
```
sk1p: Indexing by key is :+1: until we have the proper `TilingScheme`.

src/libertem/udf/stddev.py
```diff
@@ -1,56 +1,133 @@
...
-    N = p0.N + p1.N
+    if dest_N == 0:
+        dest_sum[:] = src_sum
+        dest_varsum[:] = src_varsum
+        return src_N
+    else:
+        N = dest_N + src_N
+
+        le = len(dest_sum)
+        for i in range(le):
```
sk1p: Variable names: `i` and `le`? See discussion below.

src/libertem/udf/stddev.py
```diff
@@ -1,56 +1,133 @@
...
-
-    # compute sum of variances for joint samples
-    delta_P = mean_B - mean
-    var_AB = p0.var + p1.var + (p1.N * delta * delta_P)
+    Parameters
+    ----------
+    tile : numpy.ndarray
+        Tile with flattened signal dimension
+    N0 : int
+        Number of frames already aggegated in partition buffer
```
sk1p: "partition buffer" here means the `*_inout` parameters, right? Maybe should reference them directly.

tests/udf/test_stddev.py
```diff
@@ -50,3 +55,60 @@ def test_stddev(lt_ctx, use_roi):
...
+            src_varsum=partition_varsum
+        )
+    return N, s, varsum
+
+
+# This shouldn't be @pytest.mark.numba
+# since the calculation will take forever
+# with JIT disabled
+def test_stability(lt_ctx):
+    """
```
sk1p: docstring should be deleted or updated

src/libertem/udf/stddev.py
```diff
@@ -1,56 +1,133 @@
-import collections
-
 import numpy as np
+import numba
 
 from libertem.udf import UDF
 
 
-VariancePart = collections.namedtuple('VariancePart', ['var', 'sum_im', 'N'])
+@numba.njit(fastmath=True)
```
sk1p: Is the `fastmath` needed here? Is that to give the compiler additional freedom when inlining this function in the other `merge`?

src/libertem/udf/stddev.py
```diff
@@ -130,20 +214,31 @@ def merge(self, dest, src):
...
-        dest['sum_frame'][:] = compute_merge.sum_im
+        dest_N = _validate_n(dest['num_frame'][0])
+        src_N = _validate_n(src['num_frame'][0])
+
+        oldshape = dest['sum_frame'][:].shape
+
+        dest['sum_frame'][:].shape = (-1,)
+        dest['var'][:].shape = (-1, )
+        src['sum_frame'][:].shape = (-1,)
+        src['var'][:].shape = (-1, )
```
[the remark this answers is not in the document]
uellue: Today I learned that `flatten()` creates a copy while `reshape((-1, ))` creates a view if possible. It works with `reshape()` instead of assigning to `shape`.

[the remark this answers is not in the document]
sk1p: Yes :+1: :+1: I regularly reshape flat arrays before writing to them, works really well! AFAIK it will just assign different strides and dimensions on the ndarray object.

src/libertem/udf/stddev.py
uellue: Some tests revealed that `fastmath=True` only on merge_single brings the biggest speed-up. I have no idea why.

uellue: I optimized the function further with blocking. With that version, fastmath for both `merge_single` and `process_tile` works best.

src/libertem/udf/stddev.py
uellue: The function doesn't care about frames or slices at this stage, as long as the buffer shapes are 1D and match in length. :-)

src/libertem/udf/stddev.py
uellue: I've renamed all instances of `var` to `varsum` where it contains the sum of variances and not the variance.

Comment for ddof added!

uellue:
Thx @sk1p for the feedback! This should be ready for a second round of review.

uellue:
The Numba coverage for `process_tile` was not counted for some reason. It did work before, with the exception of not passing through the blocked part of the loop since the input data was too small.

sk1p:
> The Numba coverage for process_tile was not counted for some reason. It did work before, with the exception of not passing through the blocked part of the loop since the input data was too small.

coverage issues are best debugged locally (I don't trust codecov... the travis job says coverage was uploaded, but apparently it was not correctly merged into the full coverage report, sigh.) - did you try `tox -e numba_coverage` and checking the result in the `htmlcov` directory?

tests/udf/test_stddev.py
```diff
@@ -50,3 +55,58 @@ def test_stddev(lt_ctx, use_roi):
 
     std = np.std(data[roi], axis=0)
     assert np.allclose(std, res['std'])  # check standard deviation
+
+
+@numba.njit
```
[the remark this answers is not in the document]
uellue: Ok, bumping numba requirement to `>=0.47` for bounds checking as discussed. :-)

src/libertem/udf/stddev.py
```diff
@@ -1,56 +1,183 @@
...
+    Returns
+    -------
+    n : int
+        New number of frames in aggregation buffer
+    '''
+    n_frames = tile.shape[0]
+    n_pixels = tile.shape[1]
+    n = n_0 + n_frames
+
+    BLOCKSIZE = 1024
```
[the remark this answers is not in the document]
uellue: Re-testing for optimal block size with the latest iteration of the math kernel gave 256 as an universal optimum for float32, float64 and complex128 on my machine. :-) It seems that some other changes made it less sensitive to the size since 128, 512 gave good performance as well for all sizes.

uellue:
> did you try `tox -e numba_coverage` and checking the result in the `htmlcov` directory?

Good point! The function was actually never called since the partition did fit in a single tile. I've adapted the parameters accordingly. At least locally, the coverage is 100 % now.

src/libertem/udf/stddev.py
```diff
@@ -1,56 +1,183 @@
...
+            for i in range(BLOCKSIZE):
+                varsum[i] += (tile[frame, pixel_offset + i] - mean[i])**2
+        for i in range(BLOCKSIZE):
+            sum_inout[pixel_offset + i], varsum_inout[pixel_offset + i] = merge_single(
+                n,
+                n_0, sum_inout[pixel_offset + i], varsum_inout[pixel_offset + i],
+                n_frames, sumsum[i], varsum[i], mean[i]
+            )
+    for pixel in range(n_blocks*BLOCKSIZE, n_pixels):
+        sumsum2 = 0.
```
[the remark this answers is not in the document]
uellue: Good idea, much better!

sk1p:
Perfect, thanks, merged!

PR #641, opened 2020-02-22
Fix issue 600

### LINKED ISSUE #600: RAW: TypeError when omitting detector_size
To reproduce (file doesn't need to exist):

```python
ctx.load("raw", path="/tmp/meh/meh.raw", dtype="fliat32", scan_size=(256, 256))
```

result:

```
TypeError: 'NoneType' object is not iterable
```

Should instead throw an understandable error message.

twentyse7en:
Better error message when parameter detector_size is missing.
Fix #600 
## Contributor Checklist:

### DIFF
--- a/src/libertem/io/dataset/raw.py
+++ b/src/libertem/io/dataset/raw.py
@@ -191,8 +191,8 @@ class RawFileDataSet(DataSet):
             warnings.warn("tileshape argument is deprecated, ignored", DeprecationWarning)
 
         if crop_detector_to is not None:
-            warnings.warn("crop_detector_to and detector_size_raw are deprecated, ",
-                          "and will be removed after version 0.6.0. ",
+            warnings.warn("crop_detector_to and detector_size_raw are deprecated, "
+                          "and will be removed after version 0.6.0. "
                           "please specify detector_size instead or use EMPAD DataSet",
                           DeprecationWarning)
             if detector_size is not None:
--- a/src/libertem/io/dataset/raw.py
+++ b/src/libertem/io/dataset/raw.py
@@ -203,7 +203,7 @@ class RawFileDataSet(DataSet):
             detector_size = crop_detector_to
 
         if detector_size is None:
-            raise ValueError("missing 1 required argument: 'detector_size'")
+            raise TypeError("missing 1 required argument: 'detector_size'")
 
         self._path = path
         self._scan_size = tuple(scan_size)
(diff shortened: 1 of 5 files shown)


twentyse7en:
is bugfix folder the right place to add changelog entry for this issue ?

src/libertem/io/dataset/raw.py
```diff
@@ -201,6 +202,9 @@ def __init__(self, path, scan_size, dtype, detector_size=None, enable_direct=Fal
                                  "please use EMPAD DataSet")
             detector_size = crop_detector_to
 
+        if detector_size is None:
+            raise ValueError("missing 1 required argument: 'detector_size'")
```
sk1p: This should be a `TypeError`; see the descriptions on https://docs.python.org/3/library/exceptions.html, especially:

```
exception ValueError
Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as IndexError
```

Also, as you noted in #600, `TypeError` is also the exception type you get in all other cases if you omit a required parameter.

sk1p:
> is bugfix folder the right place to add changelog entry for this issue ?

Yes!

tests/test_context.py
```diff
@@ -6,3 +10,28 @@ def test_ctx_load(lt_ctx, default_raw):
...
+
+def test_missing_detector_size(lt_ctx, default_raw):
+    with pytest.raises(ValueError) as e:
+        lt_ctx.load(
+            "raw",
+            path=default_raw._path,
+            scan_size=(16, 16),
+            dtype="float32",
+            )
+    assert e.match("missing 1 required argument: 'detector_size'")
```
sk1p: Tests look good! They should be moved to `tests/io/test_raw.py` though. You may only be calling the `load` method of the `Context`, but the behavior you are testing comes from the raw dataset.

twentyse7en:
Added changes as requested.

tests/test_context.py
twentyse7en: I thought of adding the tests to `test_raw.py` at first. But when I search for the test which open raw file normal way, It was in `test_context.py`. So I got confused.

sk1p: Ah, I can see how that is confusing. All the tests in `test_raw` just use fixtures like `default_raw` and there was no test yet that exercised the different parameters. Looks good now!

twentyse7en:
Added changelog entry as you suggested.
 
> If you like, you can also update your description in `creators.json` - you have by now done a bit more than just fixing issue 81!

Thx :) I will update in some other pull request.
Sorry for late reply, I'm having internal exam this week. I couldn't invest enough time as I thought of.

sk1p:
> Added changelog entry as you suggested.

:+1: 

> Thx :) I will update in some other pull request.

Ok :+1:

> Sorry for late reply, I'm having internal exam this week. I couldn't invest enough time as I thought of.

All okay! Thanks, will merge now!

Issue #506, opened 2019-12-10
Implement and use type system


uellue:
Currently, we use type hinting only sparsely in the API.

In the future, we should pervasively specify types in such a way that they are both machine and human readable.

This helps to make our documentation clear and precise. Additionally, it may help to find issues by using automated checking routines like [mypy](http://mypy-lang.org/). The LiberTEM code base has grown to such a volume that such tests help to maintain coherence across the project.

Parts of this issue that can be solved individually:

- [x] Integrate a type checker like mypy into our CI pipeline
- [ ] Types for the `Context` class
  - [ ] Parameters
  - [X] Return types, done in #643 
- [ ] Types for all parts of the UDF interface

sk1p:
See also: https://glyph.twistedmatrix.com/2020/07/new-duck.html

sk1p:
As an update, much of `Context` now does have type hints available - while the UDF interface itself is still a bit patchy. There is now opt-in mypy checking in pre-commit, via the [.mypy-checked file](https://github.com/LiberTEM/LiberTEM/blob/master/.mypy-checked).

uellue:
We are using type hints in pretty much all new code. Closing here!

PR #643, opened 2020-02-23
Added types to libertem.api.Context

Iamshankhadeep:
Related to #506 

## Contributor Checklist:

### DIFF
--- a/packaging/creators.json
+++ b/packaging/creators.json
@@ -1,113 +1,120 @@
 [
-    {
-        "displayname": "Alexander Clausen",
-        "authorname": "Clausen, Alexander",
-        "affiliation": "Jülich Research Centre",
-        "github": "sk1p",
-        "contribution": "System design, engineering, implementation, documentation, management, majority of the code",
-        "orcid": "0000-0002-9555-7455"
-    },
-    {
-        "displayname": "Dieter Weber",
-        "authorname": "Weber, Dieter",
-        "affiliation": "Jülich Research Centre",
-        "github": "uellue",
-        "contribution": "Management, requirements analysis, system design, testing, documentation, communication",
-        "orcid": "0000-0001-6635-9567"
-    },
-    {
-        "displayname": "Karina Ruzaeva",
-        "authorname": "Ruzaeva, Karina",
-        "affiliation": "Jülich Research Centre" ,
-        "github": "kruzaeva",
-        "contribution": "Fast diffraction analysis",
-        "orcid": "0000-0003-3610-0989"
-    },
-    {
-        "displayname": "Vadim Migunov",
-        "authorname": "Migunov, Vadim",
-        "affiliation": "RWTH Aachen University, Jülich Research Centre, Ernst Ruska-Centre",
-        "github": "woozey",
-        "contribution": "Scientific advisory, discussion, sample data, binning prototype, holography features",
-        "orcid": "0000-0002-6296-4492"
-    },
-    {
-        "displayname": "Abijith Bahuleyan",
-        "authorname": "Bahuleyan, Abijith",
-        "affiliation": "APJ Abdul Kalam Technical University",
-        "github": "twentyse7en",
-        "contribution": "Fix issue 81"
-    },
-    {
-        "displayname": "Jan Caron",
-        "authorname": "Caron, Jan",
-        "affiliation": "Jülich Research Centre",
-        "github": "jan-car",
-        "contribution": "Discussions, advanced color wheel for vector field visualization",
-        "orcid": "0000-0002-0873-889X"
-    },
-    {
-        "displayname": "Rahul Chandra",
-        "authorname": "Chandra, Rahul",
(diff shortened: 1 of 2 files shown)


src/libertem/api.py
```diff
@@ -554,7 +557,7 @@ def create_pick_analysis(self, dataset, x: int, y: int = None, z: int = None):
         parameters = {name: loc[name] for name in ['x', 'y', 'z'] if loc[name] is not None}
         return PickFrameAnalysis(dataset=dataset, parameters=parameters)
 
-    def run(self, job: Union[Job, Analysis], roi=None):
+    def run(self, job: Union[Job, Analysis], roi=None) -> np.ndarray:
```
sk1p: The return type of this method is not always `ndarray`, as noted in the docstring. If you want to express something like "either X or Y", you can use the `Union` type, see also [the official documentation](https://docs.python.org/3/library/typing.html#typing.Union). The whole page on typing should be of interest for fixing this issue

src/libertem/api.py
```diff
@@ -602,7 +605,7 @@ def run(self, job: Union[Job, Analysis], roi=None):
             return analysis.get_results(out)
         return out
 
-    def run_udf(self, dataset: DataSet, udf: UDF, roi=None, progress=False):
+    def run_udf(self, dataset: DataSet, udf: UDF, roi=None, progress=False) -> BufferWrapper:
```
sk1p: Closely re-read the docstring again: `BufferWrapper` is not the correct type here.

src/libertem/api.py
```diff
@@ -635,7 +638,7 @@ def run_udf(self, dataset: DataSet, udf: UDF, roi=None, progress=False):
         """
         return UDFRunner(udf).run_for_dataset(dataset, self.executor, roi, progress=progress)
 
-    def map(self, dataset, f, roi=None):
+    def map(self, dataset, f, roi=None) -> BufferWrapper:
```
sk1p: Same comment as for `run_udf` applies here

src/libertem/api.py
```diff
@@ -95,7 +96,7 @@ def load(self, filetype: str, *args, **kwargs) -> DataSet:
     load.__doc__ = load.__doc__ % {"types": ", ".join(filetypes.keys())}
 
     def create_mask_job(self, factories, dataset, use_sparse=None,
-                        mask_count=None, mask_dtype=None, dtype=None):
+                        mask_count=None, mask_dtype=None, dtype=None) -> np.ndarray:
```
sk1p: `np.adarray` is not the correct type here, please compare with the docstring. You can also have a look at the code of the method itself: find the `return` statements and have a look at what kind of object is returned

src/libertem/api.py
Iamshankhadeep: After running Context.map() in jupyter I am getting return type as  libertem.common.buffers.BufferWrapper.

sk1p: Ah, I was wrong here, `BufferWrapper` is correct!

packaging/creators.json
```diff
@@ -1,113 +1,120 @@
...
+    "contribution": "Fixed issue 267"
+  },
+  {
+    "displayname": "Shankhadeep Dey",
+    "authorname": "Dey, Shankhadeep",
+    "affiliation": "Siliguri Institute of Technology",
+    "github": "iamshankhadeep",
+    "contribution": "Added types to libertem.api.Context #643"
+  }
+]
```
sk1p: Thanks for adding yourself to `creators.json`! It is a bit problematic though to re-indent the whole file. This makes it harder for us to review the changes, and may cause merge conflicts for other contributors, or maybe even yourself if you need to rebase this PR to the current master at a later point.

src/libertem/api.py
```diff
@@ -602,7 +607,7 @@ def run(self, job: Union[Job, Analysis], roi=None):
             return analysis.get_results(out)
         return out
 
-    def run_udf(self, dataset: DataSet, udf: UDF, roi=None, progress=False):
+    def run_udf(self, dataset: DataSet, udf: UDF, roi=None, progress=False) -> dict:
```
sk1p: I would prefer a more concrete type here - what are the key and value types of this `dict`?

src/libertem/api.py
```diff
@@ -554,7 +559,7 @@ def create_pick_analysis(self, dataset, x: int, y: int = None, z: int = None):
         parameters = {name: loc[name] for name in ['x', 'y', 'z'] if loc[name] is not None}
         return PickFrameAnalysis(dataset=dataset, parameters=parameters)
 
-    def run(self, job: Union[Job, Analysis], roi=None):
+    def run(self, job: Union[Job, Analysis], roi=None) -> Union[np.ndarray, AnalysisResultSet]:
```
sk1p: Looks good!

packaging/creators.json
Iamshankhadeep: Sorry, it was my mistake I did not notice it.

PR #647, opened 2020-02-24
Use type system parameter

Iamshankhadeep:
related to #506 
## Contributor Checklist:

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,2 @@
+[mypy]
+ignore_missing_imports = True
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,3 +1,0 @@
-[mypy]
-;to supress "No library stub file for module error"
-ignore_missing_imports = True
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,2 +1,3 @@
 [mypy]
+;to supress "No library stub file for module error"
 ignore_missing_imports = True
--- a/src/libertem/api.py
+++ b/src/libertem/api.py
@@ -7,6 +7,7 @@ from libertem.io.dataset import load, filetypes
 from libertem.io.dataset.base import DataSet
 from libertem.job.masks import ApplyMasksJob
 from libertem.job.raw import PickFrameJob
+from libertem.job.base import Job
 from libertem.common import Slice, Shape
 from libertem.common.buffers import BufferWrapper
 from libertem.executor.dask import DaskJobExecutor
@@ -563,8 +564,8 @@ class Context:
         parameters = {name: loc[name] for name in ['x', 'y', 'z'] if loc[name] is not None}
         return PickFrameAnalysis(dataset=dataset, parameters=parameters)
 
-    def run(self, job: BaseAnalysis,
-            roi: np.ndarray = None, progress: bool = False) -> Union[np.ndarray, AnalysisResultSet]:
+    def run(self, job: Union[Job, BaseAnalysis],
+            roi: np.ndarray = None) -> Union[np.ndarray, AnalysisResultSet]:
         """
         Run the given :class:`~libertem.job.base.Job` or :class:`~libertem.analysis.base.Analysis`
         and return the result data.
--- a/src/libertem/api.py
+++ b/src/libertem/api.py
@@ -21,7 +21,7 @@ from libertem.analysis.ring import RingMaskAnalysis
 from libertem.analysis.sum import SumAnalysis
 from libertem.analysis.point import PointMaskAnalysis
 from libertem.analysis.masks import MasksAnalysis
-from libertem.analysis.base import AnalysisResultSet, BaseAnalysis
+from libertem.analysis.base import AnalysisResultSet, Analysis
 from libertem.udf.base import UDFRunner, UDF
 from libertem.udf.auto import AutoUDF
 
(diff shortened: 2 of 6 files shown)


sk1p:
Thank you for your PR! I think the branch was created on top of the branch of your last PR - this now creates a conflict as it contains identical changes as the previous PR. Maybe you can try to rebase your changes on top of the current master branch. Thanks!

src/libertem/api.py
```diff
@@ -95,8 +96,12 @@ def load(self, filetype: str, *args, **kwargs) -> DataSet:
 
     load.__doc__ = load.__doc__ % {"types": ", ".join(filetypes.keys())}
 
-    def create_mask_job(self, factories, dataset, use_sparse=None,
-                        mask_count=None, mask_dtype=None, dtype=None) -> ApplyMasksJob:
+    MaskArrayType = Union[np.ndarray, sp.coo.coo_matrix, sp.dok.dok_matrix]
```
sk1p: This type alias should either be at the topmost level of the file (i.e. not as an attribute of the `Context` class, but as an attribute of the module itself), or even better, it could be moved to a more specific module, for example `libertem.masks` or `libertem.udf.masks`, from where it can be used in the future to type all the other functions that take this form of mask array.

(As a description why it should be moved: this exact type is used in quite a few places, like `libertem.udf.masks`. Now one could think: just let `MaskArrayType` live in `libertem.api` and import that from `libertem.udf.masks`, but this would create an import cycle, in this case via the analysis module. But if `MaskArrayType` lives in `libertem.masks`, you would only have imports going in one direction. No more circular imports!)

src/libertem/api.py
```diff
@@ -95,8 +96,12 @@ def load(self, filetype: str, *args, **kwargs) -> DataSet:
...
+    
+    def create_mask_job(self,
+            factories: Union[Callable[[], MaskArrayType], Iterable[Callable[[], MaskArrayType]]],
```
sk1p: The type for `factories` could also be extracted as an alias - as you are re-using the same construct in the `create_mask_analysis` function below. This new type could also be extracted to a specific module, like `libertem.masks`.

src/libertem/api.py
```diff
@@ -170,8 +175,10 @@ def create_mask_job(self, factories, dataset, use_sparse=None,
             dtype=dtype,
         )
 
-    def create_mask_analysis(self, factories, dataset, use_sparse=None,
-                             mask_count=None, mask_dtype=None, dtype=None) -> MasksAnalysis:
+    def create_mask_analysis(self,
+            factories: Union[Callable[[], MaskArrayType], Iterable[Callable[[], MaskArrayType]]],
```
sk1p: See comment above for `create_mask_job` - same applies here.

Iamshankhadeep:
Yes after re running mypy with this version it is showing less error messages, some of them I fixed in this version and I did not find how to fix other errors.

src/libertem/api.py
```diff
@@ -559,7 +567,8 @@ def create_pick_analysis(self, dataset, x: int, y: int = None,
         parameters = {name: loc[name] for name in ['x', 'y', 'z'] if loc[name] is not None}
         return PickFrameAnalysis(dataset=dataset, parameters=parameters)
 
-    def run(self, job: Union[Job, Analysis], roi=None) -> Union[np.ndarray, AnalysisResultSet]:
+    def run(self, job: BaseAnalysis,
```
Iamshankhadeep: Like you can see here, the type of job was `Union[Job, Analysis]` but `Job.TYPE` and `Analysis.TYPE` does not exists. After understanding the codebase I learned that `BaseAnalysis` does have `TYPE` in it but Analysis does not have TYPE. So I updated it with `BaseAnalysis`

sk1p: The problem with this approach is: the `run` function does need to accept objects of type `Job`, otherwise there would be no way to run them! The approach to fix a type error is not always so straight forward, in this case, the change needs to happen somewhere else.

Iamshankhadeep: Then I think it should be better to leave this as `Union[Job, BaseAnalysis]`

sk1p: Actually, instead of `BaseAnalysis` it should be `Analysis`. I guess this gives an error because `Analysis` doesn't have the `TYPE` attribute, among other things? If so, this is actually a problem with the `BaseAanlysis` and `Analysis` classes - the `Analysis` class should define the complete interface (including attributes), and `BaseAnalysis` should only contain some default implementations.

This actually shows very well the value of these type annotations - they catch mistakes that slipped through the code review.

Let me know if you have further questions!

Iamshankhadeep: I get it now, but I think if we implement `mypy` now in CI then this error will be there. Is there anyway that `mypy` passes?

sk1p: For the beginning, we can make a separate job in travis for mypy which is allowed to fail. That way, we get kind of a TODO list of errors to fix out of the CI job. If we type check the whole code base, we will have many error messages anyway.

Iamshankhadeep: That makes sense. Then I will try to implement `mypy` too.

src/libertem/udf/base.py
```diff
@@ -87,16 +88,16 @@ class UDFData:
     '''
     Container for result buffers, return value from running UDFs
     '''
-    def __init__(self, data):
+    def __init__(self, data: Dict):
```
sk1p: Would prefer a more concrete type here: what are the key and value types?

mypy.ini
```diff
@@ -0,0 +1,2 @@
+[mypy]
+ignore_missing_imports = True
```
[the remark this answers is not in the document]
Iamshankhadeep: to supress `No library stub file for module error` we need to use `ignore_missing_imports`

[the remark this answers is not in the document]
sk1p: These error messages are actually useful: we want to have type stubs or types for as many external libraries as possible. I would suggest to remove this setting. Then, anyone who wants to improve the typing can have a look at the CI output and use the errors as a kind of TODO list what still needs to be fixed.

Rest of the PR looks good :ok_hand:!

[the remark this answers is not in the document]
Iamshankhadeep: then I will remove the `mypy.ini` and update the pr again

sk1p:
Thanks, will merge after CI succeeds!

sk1p:
Merged, thank you for your patience!

PR #650, opened 2020-02-25
Added citation of dataset used

twentyse7en:
Citation for dataset used in screenshot
is added.

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -154,6 +154,13 @@ below the right window.
 
 ..  figure:: ./images/use/image.png
 
+Citation
+~~~~~~~~
+
+Ophus, Colin, & Savitzky, Benjamin. (2019). Simulated calibration dataset for
+4D scanning transmission electron microscopy [Data set].
+Zenodo. http://doi.org/10.5281/zenodo.3592520
+
 Application-specific documentation
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -82,7 +82,8 @@ below it lists all files and subdirectories in that directory. You select an
 entry by clicking once on it. You can move up one directory with the ".." entry
 on top of the list. The file browser is still very basic. Possible improvements
 are discussed in `Issue #83 <https://github.com/LiberTEM/LiberTEM/issues/83>`_.
-Contributions are highly appreciated!
+Contributions are highly appreciated! This example opens an HDF5 file.
+:cite:`ophus_colin_2019_3592520`
 
 ..  figure:: ./images/use/open.png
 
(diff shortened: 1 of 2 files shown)


docs/source/usage.rst
```diff
@@ -142,6 +142,13 @@ below the right window.
 
 ..  figure:: ./images/use/image.png
 
+Citation
+~~~~~~~~
```
sk1p: Thanks for working on this! I think there is no need for "Citation" headline. Think of the citation as a kind of thank-you note to the group of people publishing the data, and even more so a professional acknowledgement. It should be clear from this short note how we are using the data.

For actually referencing the data set, you can and should use the preconfigured citation system - see the [holography docs](https://libertem.github.io/LiberTEM/app/holography.html#hologram-simulation) for an example. For that to work, you need to add [the bibtex entry](https://zenodo.org/record/3592520/export/hx) of the data set to our bib file. This will also make the data set appear, properly formatted and liked, on [the list of citations](https://libertem.github.io/LiberTEM/index.html#citations) in our documentation, and later in the list of citations [on our zenodo entry](https://zenodo.org/record/3674003)

For another example, see also the beginning of https://nbviewer.jupyter.org/github/LiberTEM/LiberTEM-blobfinder/blob/master/examples/strainmap-poly.ipynb - although this is not using the bibtex citing system, as that is not so easily possible from a notebook.

/cc @uellue

twentyse7en: Thx! I have never used citation, Now it's very clear. Updated PR with requested changes.

twentyse7en:
@uellue Could you please check the PR, It would be helpful for those who are looking data used in GUI usage.

uellue:
Hm, strange that the citation triggered a CI error. I'll have to look at it later. Regarding datasets, perhaps it could make sense to collect a number of publicly available ones with a short description?

twentyse7en:
That would be great!

twentyse7en:
> Regarding datasets, perhaps it could make sense to collect a number of publicly available ones with a short description?

Where can I find the datasets , also where are we planning to add this section in  documentation ?

uellue:
For a good solution we should probably combine this with the overview table in #670.

- BibTeX entry for each dataset
- using `:cite:` instead of a download link in the table
- A very short description, which I can take care of since it is application-specific.

All in all, I'd propose to remove the `:cite:` that is causing issues for unknown reasons at the moment, add BibTeX entries for the other source(s) from #670 here to not create a merge conflict, and then merge both into master.

After that, I can debug the `:cite:` issue and add descriptions.

Sounds good? :-)

uellue:
Ah, for context: The goal is not only to give users access to datasets, but also to create visibility and impact for the people who uploaded the data. `:cite:` and BibTeX entries can create back links that are counted and show impact.

Furthermore, we should always make sure that the authors are visibly credited. 

CC #670 :-)

twentyse7en:
> 
> All in all, I'd propose to remove the `:cite:` that is causing issues for unknown reasons at the moment, add BibTeX entries for the other source(s) from #670 here to not create a merge conflict, and then merge both into master.
> 

I have removed `:cite:` from usage docs and added BibTeX for the other dataset .

>
>Sounds good? :-)
>

absolutely :smile:

Issue #629, opened 2020-02-20
Releasing docs: add item about handling deprecations


sk1p:
Before cutting a release, search the code base for `DeprecationWarning`s that are supposed to be removed in that release. Possibly also add a "things that will be removed in the next release" section in the changelog for proper messaging.

twentyse7en:
I would like to contribute.

sk1p:
> I would like to contribute.

Okay! Did you already find the release checklist? This issue is about adding new items in there to remind us to take care of deprecations.

sk1p:
Another point for the release checklist: Add the current LiberTEM version to CVL @ https://github.com/Characterisation-Virtual-Laboratory/CharacterisationVL-Software

twentyse7en:
Where can I find release checklist ?

sk1p:
It [on the releasing page](https://libertem.github.io/LiberTEM/releasing.html#release-checklist), which is linked from [Contributing -> Advanced](https://libertem.github.io/LiberTEM/contributing.html#advanced)

twentyse7en:
So I need to add items in checklist which will removed in release 0.5, right ?

sk1p:
The idea is to add general checklist items for handling deprecation and publishing to CVL, not specifically for one version

twentyse7en:
For finding `DeprecationWarning` I have gone through `LiberTEM/src`. Do I have to search in entire code(`LiberTEM/`) or just `Libertem/src`

sk1p:
Hmm, you are overthinking it. The idea is just to have a *very* simple item in the release checklist to remind us to take care of deprecations. There doesn't need to be a list of concrete warnings at all, and also no need for a detailed description: it is supposed to be terse. The target audience of the release checklist are experienced contributors of LiberTEM who want to either a) do a release themselves or b) help out with doing the quality assurance.

twentyse7en:
Thx, I was going in a different direction :sweat_smile: Hope the changes made in PR are something you expects.

PR #653, opened 2020-02-26
Add items to release checklist

twentyse7en:
Added items mentioned in issue #629

### DIFF
--- a/docs/source/releasing.rst
+++ b/docs/source/releasing.rst
@@ -8,8 +8,12 @@ contributions.
 Release checklist
 -----------------
 
-Not all aspects of LiberTEM are covered with automated unit tests. For that
-reason we should perform some manual tests before and after a release.
+* Not all aspects of LiberTEM are covered with automated unit tests. For that
+  reason we should perform some manual tests before and after a release.
+* Handling deprecation, search the code base for :code:`DeprecationWarning`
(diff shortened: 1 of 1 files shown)


docs/source/releasing.rst
```diff
@@ -8,8 +8,12 @@ contributions.
...
 -----------------
 
-Not all aspects of LiberTEM are covered with automated unit tests. For that
-reason we should perform some manual tests before and after a release.
+* Not all aspects of LiberTEM are covered with automated unit tests. For that
+  reason we should perform some manual tests before and after a release.
+* Handling deprecation, search the code base for :code:`DeprecationWarning`
+  that are supposed to be removed in that release.
+* Add the current LiberTEM version to
+  `CVL <https://github.com/Characterisation-Virtual-Laboratory/CharacterisationVL-Software>`_
```
sk1p: Replace Handling with Handle here. Other than that, the text itself is :ok_hand: 

The steps themselves should be put into the right order, under the "Before" headline below. Like: handling deprecations needs to happen before any QA, and publishing to CVL can only happen once the code is in its final form, that is, if the rest of the release process is done.

Otherwise it looks good! :+1:

sk1p:
Thanks, merged!

PR #655, opened 2020-02-29
Added progress bar support

### LINKED ISSUE #644: Progress bar support for `Context.map` and `Context.run`
Support in `Context.map` is straight-forward, same for `Context.run` with UDF-based analyses; I think adding progress bar support for the deprecated job-based analyses is not important, even if it is possible.

twentyse7en:
Added progress bar support for `Context.map` and `Context.run`
fix #644 
## Contributor Checklist:

### DIFF
--- a/docs/source/changelog/misc/progress_bar_support.rst
+++ b/docs/source/changelog/misc/progress_bar_support.rst
@@ -1,0 +1,5 @@
+[Misc] Added progress bar support
+=================================
+
+* Added progress bar support for :meth:`libertem.api.Context.run` and
+  :meth:`libertem.api.Context.map`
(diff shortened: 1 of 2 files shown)


uellue:
Hi @twentyse7en, thank you for the pull request! That turns the progress bar on all the time, if I interpret it correctly. The progress bar should only be shown if the user activates it since it can create issues in case the terminal doesn't support it or if LiberTEM is used as a back-end in another application. You can add a default parameter `progress=False` to `map()` and `run()` instead and pass it on.

twentyse7en:
Oh! I only thought of using in notebook :sweat_smile: . Added changes as requested.

uellue:
Looks good, merging. Many thanks, good work! :+1:

PR #659, opened 2020-03-01
Exception for Direct I/O for raw files on Windows

### LINKED ISSUE #658: RAW: Dataset with Direct I/O enabled fails to load on Windows
On Windows (with Python 3.7), trying to load a raw dataset with direct I/O enabled results in the following error: `module 'os' has no attribute 'O_DIRECT'`

I assume this only occurs on Windows as in #599 it is stated that "When running an analysis, note that direct I/O is enabled".

AnandBaburajan:
Fixes #658 

An exception is raised when Direct I/O is enabled for raw files on Windows and a user-friendly error message is shown on the GUI.

## Contributor Checklist:

### DIFF
--- a/docs/source/changelog/bugfix/raw_direct_io_windows.rst
+++ b/docs/source/changelog/bugfix/raw_direct_io_windows.rst
@@ -1,0 +1,4 @@
+[Feature] Exception for Direct I/O for raw files on Windows
+============================================================
+
+* Exception is raised when Direct I/O is enabled for raw files on Windows (:pr:`659`)
--- a/src/libertem/io/dataset/raw.py
+++ b/src/libertem/io/dataset/raw.py
@@ -101,11 +101,8 @@ class RawFile(File3D):
 
     def open(self):
         if self._enable_direct:
-            try:
-                fh = os.open(self._path, os.O_RDONLY | os.O_DIRECT)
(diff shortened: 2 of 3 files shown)


sk1p:
I think the AppVeyor test failure is not your fault, let me re-run the job

AnandBaburajan:
> If you want to decrease the "cognitive complexity", as detected by codeclimate, a possibility is to move the check for exact condition (`enable_direct` and no `O_DIRECT` support) to the `check_valid` function of the dataset, and raise the exception there.

How did you know that would solve the codeclimate issue? Does it come with more Python experience or would I get better at it once I'm familiar with the project's code? Thanks!

sk1p:
> How did you know that would solve the codeclimate issue?

Complexity is in most cases measured roughly as the number of paths the code can take in a given function/method. So each if-statement or try...except block ~doubles the number of paths. If you keep this in mind when writing code, you develop an intuition for how complex a given function is.

See also: https://en.wikipedia.org/wiki/Cyclomatic_complexity

> would I get better at it once I'm familiar with the project's code?

It is a combination of understanding the underlying metric, developing intuition for it, and experience of the code base: in this case, it makes sense to do this check as early as possible, so it is moved explicitly to the `check_valid` method.

PR #664, opened 2020-03-04
Improved installation docs

twentyse7en:
Generalised virtualenv creation for different
python version.
Added a optional instruction for installing pytorch.

## Contributor Checklist:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -47,13 +47,13 @@ The short version:
 
 .. code-block:: shell
 
-    $ virtualenv -p python3.6 ~/libertem-venv/
+    $ virtualenv -p python3 ~/libertem-venv/
     $ source ~/libertem-venv/bin/activate
-    (libertem) $ pip install libertem[torch]
+    (libertem) $ pip install "libertem[torch]"
 
 Please see `our documentation <https://libertem.github.io/LiberTEM/install.html>`_ for details!
 
-Deployment as a single-node system for a local user is thoroughly tested and can be considered stable. Deployment on a cluster is 
+Deployment as a single-node system for a local user is thoroughly tested and can be considered stable. Deployment on a cluster is
 experimental and still requires some additional work, see `Issue #105 <https://github.com/LiberTEM/LiberTEM/issues/105>`_.
 
 Applications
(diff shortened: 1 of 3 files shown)


docs/source/install.rst
```diff
@@ -58,7 +58,11 @@ To create a new virtualenv for LiberTEM, you can use the following command:
 
 .. code-block:: shell
 
-    $ virtualenv -p python3.7 ~/libertem-venv/
+    $ virtualenv -p python3.x ~/libertem-venv/
+
+Replace :code:`x` with current Python version installed. To find out which
+version of Python is installed on your system run the :code:`python --version`
+or :code:`python -V` command.
```
sk1p: Hmm, this is kind of redundant. If you do `python --version` and put that into the `virtualenv` command, you could as well do `virtualenv -p python ~/libertem-venv/` (or even leave off the `-p` switch entirely).

Also note that the `python` binary can point to either Python 2 or Python 3 - this is detailed in [PEP 394](https://www.python.org/dev/peps/pep-0394/#for-python-runtime-distributors). So if I followed your instructions on my system, I would get a Python 2 virtualenv, which is not compatible with LiberTEM. To make sure you get Python 3, you can use the `python3` binary:

>We expect Unix-like software distributions (including systems like macOS and Cygwin) to install the python2 command into the default path whenever a version of the Python 2 interpreter is installed, and the same for python3 and the Python 3 interpreter

The point of using `-p python3.7` was to get a compatible Python version, not just any installed python3. So maybe we should use `-p python3` as the easiest way that works on most systems, but point out that you need to substitute a compatible version if you have multiple python versions installed. For example, I have Python 3.6, 3.7 and 3.8 installed on my system, so they would be all compatible, but if someone had Python 3.5 and 3.6 installed, they possibly would have to explicitly use `python3.6` to get LiberTEM running.

Note that the README.rst also contains virtualenv creation, so it should be changed there, too.

docs/source/changelog/documentation/venv_doc.rst
```diff
@@ -0,0 +1,6 @@
+[Documentation] virtualenv instruction improved
+===============================================
+
+ * Instruction generalised for different python
+   version.
+ * optional pytorch installation
```
sk1p: pytorch installation was in docs before, describe a bit more what you changed

docs/source/install.rst
```diff
@@ -198,6 +202,13 @@ example from PyPi:
 
     (libertem) $ pip install libertem[torch]
 
+Some shell interprets above code differently and might cause an error, in this
+scenario quoting the package name argument might help.
+
+.. code-block:: shell
+
+    (libertem) $ pip install "libertem[torch]"
+
```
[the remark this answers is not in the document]
twentyse7en: Thats right, also too many codes way raise confusion too!

docs/source/install.rst
twentyse7en: Thanks! Such a wonderful explanation. Made some changes, hope that reflect what you mentioned.

README.rst
```diff
@@ -47,13 +47,13 @@ The short version:
 
 .. code-block:: shell
 
-    $ virtualenv -p python3.6 ~/libertem-venv/
+    $ virtualenv -p python3 ~/libertem-venv/
     $ source ~/libertem-venv/bin/activate
     (libertem) $ pip install libertem[torch]
```
sk1p: This line should also have quotes, like the corresponding line in the documentation.

twentyse7en: Oops! forgot that one, Fixed :smile:

sk1p:
Thanks, will merge once CI passes!

Issue #46, opened 2018-07-16
Export calculation result


uellue:
Currently, the GUI shows a false color map of the calculation result.

It should be possible to export the result as numerical values, for example as raw binary, HDF5 or CSV so that it can be used for further processing.

sk1p:
Digital micrograph supports a weird CSV format, which is just rows of intensity values separated by commas - we should also export to this dialect, in addition to the more straight forward "x, y, intensity1, ..., Intensity_n"

Maybe also float .tiff images?

bdalevin:
I would like to add a comment to this issue. 
Allowing users to save a float .tif image from the GUI would be extremely helpful for many of my users. 
Please implement this as soon as you can.

sk1p:
Thank you for the feedback! We will prioritize this issue and plan to implement this for the 0.5 release.

sk1p:
The basic functionality is now implemented, with support for HDF5, NPZ, TIFF, and RAW formats. It can be tested by [installing LiberTEM from git](https://libertem.github.io/LiberTEM/install.html).

If you have any feedback, let us know in this issue, the [followup issue](https://github.com/LiberTEM/LiberTEM/issues/687) or via E-Mail. Thanks!

PR #665, opened 2020-03-05
Downloadable results

### LINKED ISSUE #168: Keep more state on API server
* Keep results around, needed for interactive re-visualization (#166 , #128)
 * Keep analysis incl. parameters in API server, which then defines the lifetime of the result blobs (only if you remove an analysis / close the dataset should the result be cleaned up)

After this, closing and re-opening a browser tab should also keep the analysis around, not only the data sets. Opening a new analysis in the GUI should no longer re-calculate the average frame, as the result is already available. (Need to differentiate between the cases of actively pressing a button and "passively" requesting the average detector image. The first should always re-calculate, the second should access the existing result if possible.)

Per analysis, only one result should be kept, which is replaced when re-running the analysis, i.e. with different parameters. The partial result should be considered volatile and "thrown away" as it is now.

Action plan (not complete yet):

- [x] define the API for creating/reading/updating/deleting Analyses via the Web API
- [x] add analyses to the `SharedState`
- [x] change job part of the web api to reference analyses. running/canceling/... jobs should still work via the job api, just the parameters should be set on the analysis before. the job is then created from the `uuid` of the analysis.

### LINKED ISSUE #328: UI: saving of the analyses
Although it has been discussed several times, just want to see some feedback and that this topic stays somewhere in the TODOs:

As UI is very handy for interactive analyses, would be great to have several saving possibilities:

* Save raw results of the analyses ideally in several formats as an option: npy, hspy, raw binary, libertem object, ... - ?;
* Save jobs that one can run them using api;
* Save job parameters in some other way: text, python dictionary (npy-?).

sk1p:
This is more of a heads-up for anyone else working on GUI code right now: quite a few things are changing regarding running jobs and analyses. This is not fully finished but will hopefully be done for 0.5. 

This PR changes the state structure of the client and the Web API. What was previously called `Analysis` in the client is now called a `CompoundAnalysis` (it combines several `Analyses` instances), and `Analysis` matches more the meaning that is used in the Python API. `Analysis` objects are also created on the server, and after a successful run, the full results are kept around. A new endpoint for downloading results has been added.

As a MVP, a simple download link was added that downloads a HDF5 dataset of the results.

This also includes a bit of cleanup of the state handling in the Web API.

Still TODO:

- [x] Also keep `CompoundAnalysis` on the server, to allow cleanup after reconnecting to the server (not sure if I want to pursue full "sync" between multiple connections to the server yet... maybe this will come as a side-effect)
- [x] Generic save interface: input: `AnalysisResultSet` and analysis details, output: filename and content-type, and write bytes to a given output buffer (do we need to write in chunks? Does tornado do this for us?)
- [ ] Support for writing "metadata", in this case: LiberTEM version, analysis type, parameters, source dataset (might need to be censored? sensitive filenames/paths etc.)
- [ ] Generate filename from name entered in the GUI?
- [x] Support for float TIFF format (how to support multiple channels?)
- [x] Support for npz
- [x] Support for raw binary files
- [ ] Support for CSV
- [ ] Support for hspy
- [x] Documentation
- [x] Drop-down with the different formats in the GUI
- [ ] Try to fix as many codeclimate issues as possible
- [x] Re-think download UI: maybe better to have single DL button for CompoundAnalysis and select Analysis and Channels afterwards? Maybe even in a modal dialog? Or is a "dropdown-like" UI enough?

Fixes: #46, fixes #328, fixes #168

## Contributor Checklist:


sk1p:
This PR is already quite a monster, I think I'll merge it and continue work on a separate PR. That way we can also get it better tested before the next release, and potentially already can get some feedback from users.

Issue #498, opened 2019-12-04
File dialog: support for file stacks


sk1p:
This could be, in the easiest version, opening files via entering a glob (should present a file list as second step, when entering the other parameters). Could also consist of having a selection interface in the file dialog.

Also needs "natural" sorting, for example via the [natsort](https://natsort.readthedocs.io) package, as many data sets sadly don't use leading zeros in the filename.

Iamshankhadeep:
Can you tell me more about this issue? Like what should I need to do and what is glob?

sk1p:
> what is glob?

https://en.wikipedia.org/wiki/Glob_(programming)
https://docs.python.org/3.7/library/glob.html

> Can you tell me more about this issue?

Opening a "stack of files" refers to having a dataset that consists not of a single file, but of many smaller files. These are then "stacked" together to create a dataset of more dimensions than a single file. For example, you could have 256 files, each containing a single 2D image of size `(128, 128)`. You could then stack them to create a single 3D dataset of shape `(256, 128, 128)`, or a 4D dataset of `(16, 16, 128, 128)`. This is often used together with the DM file format, but could later also be useful for other formats, for example raw, tif or [edf](https://github.com/LiberTEM/LiberTEM/issues/220). 

Now, if you have a look at the current file browser in the GUI, only a single file can be selected, which works for most formats. The task is to improve the file browsing interface and add the possibility to select multiple files. This includes (not necessarily in this order):

1) Thinking about a user-friendly way to select multiple files (how should the UI look like and behave?)
2) Implementing the results of step 1 in the file browser
3) Adding a parameter form for the DM format
4) Changing the web API backend to support not just a single path, but multiple files and/or glob patterns (depending on what the result of step 1 was)

Now to get started, I would suggest to keep it very simple and think about the MVP, the minimum viable product. How can we support opening multiple files in the most simple way? Even writing only a prototype implementation would be a good preparation for your GSoC project.

For example data in DM format and some additional discussion, see also https://github.com/LiberTEM/LiberTEM/issues/291

Iamshankhadeep:
But what if we select multiple file of different types, then what happens?
>Adding a parameter form for the DM format
I think I did not get it. Can you provide me with more details?

sk1p:
> But what if we select multiple file of different types, then what happens?

Good question! In my opinion, the best thing to do would be to display an error, as it is the least surprising behavior.

>>Adding a parameter form for the DM format

[...]

PR #666, opened 2020-03-06
Error handling for unsupported datasets in GUI

### LINKED ISSUE #645: Proper error handling when opening certain files
If you try to open a dataset that is supported in LiberTEM itself, but not yet in the GUI (currently, that is  just DM), there is no proper error handling in place yet:

https://github.com/LiberTEM/LiberTEM/blob/b20b21dd455ae14d6f202259b995231e3f889e69/client/src/dataset/components/DatasetOpen.tsx#L159

THis is because for a long time, this truly could not happen. Instead, this should trigger a proper error message and dialog (like other errors).

AnandBaburajan:
Fixes #645 

## Contributor Checklist:


client/src/dataset/sagas.ts
```diff
@@ -63,12 +63,19 @@ export function* openDatasetSagaFullPath(action: ReturnType<typeof browserAction
 export function* openDatasetSaga(action: ReturnType<typeof browserActions.Actions.select>) {
     const config: ConfigState = yield select((state: RootReducer) => state.config);
     const fullPath = joinPaths(config, action.payload.path, action.payload.name);
-    yield call(doOpenDataset, fullPath);
+    if(fullPath.includes(".dm") || fullPath.includes(".DM")) {
+      const timestamp = Date.now();
+      const id = uuid();
+      yield put(datasetActions.Actions.detectFailed(fullPath));
+      yield put(datasetActions.Actions.error(`DM`, `DM dataset is currently not supported in the GUI`, timestamp, id));
+    } else {
```
sk1p: Nice, this should work well for .dm files! Though instead of `includes`, I would check the end of the string for dm3 or dm4, as a file in another format may very well be called `frame-54.dmsomething-123.bin`, for example, which would be caught by your above check.

As an improvement, it should be possible to do this in a more general way: all GUI-supported formats are listed in the `DatasetTypes` enum, so if the detect step returns a type that is not part of that enum, the above error could be triggered. This may need some "convincing" the type system, but should be possible. Let me know if you need help at any time!

If we have this more general way, if will work well even in the face of changes: if, some day soon (maybe as part of #498), we start supporting DM, the check doesn't need to be changed; also, if we later add a new format to the backend and don't support it from the beginning in the GUI, the same check will still work.

AnandBaburajan: > If we have this more general way, if will work well even in the face of changes

Thanks!

client/src/dataset/sagas.ts
```diff
@@ -44,8 +44,16 @@ export function* doOpenDataset(fullPath: string) {
...
-            yield put(datasetActions.Actions.detected(fullPath, detectResult.datasetParams));
+            if(Object.keys(DatasetTypes).some((v) => v === detectResult.datasetParams.type)) {
+              detectedParams = detectResult.datasetParams;
+              yield put(datasetActions.Actions.detected(fullPath, detectResult.datasetParams));
+            }
+            else {
+              const timestamp = Date.now();
+              const id = uuid();
+              yield put(datasetActions.Actions.detectFailed(fullPath));
+              yield put(datasetActions.Actions.error(detectResult.datasetParams.type, detectResult.datasetParams.type + ` dataset type is currently not supported in the GUI`, timestamp, id));
```
sk1p: I think there is a small mistake here - the first argument of `Actions.error` is supposed to be the dataset uuid. So while the types may match and there is no linter or compiler error, the meaning or semantics of the value is not right.

I quickly had a look, and it seems that the dataset doesn't even have an id yet at that point. The usual sequence is:

Open File Browser -> Select file in Browser -> Send `path` to server for detection -> Show "parameters" form -> Assign UUID and "create" the dataset in the redux store.

The dataset uuid in the `error` action is currently used [to immediately remove a dataset](https://github.com/LiberTEM/LiberTEM/blob/afe3c4292a2d55d31b0b2177aff0fa5e0cc9cb51/client/src/dataset/reducers.ts#L33) for which we get this action, if it exists. As this action is idempotent, it should not hurt to just generate a new uuid in `doOpenDataset` for this purpose, which then represents the "dataset-that-should-have-been". The alternative would be  to generate the "real" uuid earlier in the sequence mentioned above, but I think this is overkill here, as it would just be removed right away anyways...

sk1p:
Actually I just had an idea to fix the codeclimate issue: the whole "detection worked ok" branch could be extracted into its own function. What do you think?

AnandBaburajan:
> Actually I just had an idea to fix the codeclimate issue: the whole "detection worked ok" branch could be extracted into its own function. What do you think?

Sounds good! I'll try do that.

client/src/dataset/components/DatasetOpen.tsx
```diff
@@ -156,7 +155,7 @@ const DatasetOpen = () => {
             return renderForm(<EMPADParamsForm {...commonParams} initial={initial} />)
         }
     }
-    return assertNotReached("unknown dataset type");
```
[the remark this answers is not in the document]
AnandBaburajan: Right.

client/src/dataset/sagas.ts
```diff
@@ -44,8 +44,16 @@ export function* doOpenDataset(fullPath: string) {
         yield put(datasetActions.Actions.detect(fullPath));
         const detectResult: DetectDatasetResponse = yield call(detectDataset, fullPath);
         if (detectResult.status === "ok") {
-            detectedParams = detectResult.datasetParams;
-            yield put(datasetActions.Actions.detected(fullPath, detectResult.datasetParams));
+            if(Object.keys(DatasetTypes).some((v) => v === detectResult.datasetParams.type)) {
```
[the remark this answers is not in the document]
AnandBaburajan: I'll try to create a function for the "detection worked ok" part.

client/src/dataset/sagas.ts
AnandBaburajan: Got it.

client/src/dataset/sagas.ts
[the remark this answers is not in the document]
sk1p: :+1: - in addition, I would propose to still look at making the condition also a small function, as proposed above, and compare readability.

AnandBaburajan:
> the error is shown as currently, and the user is navigated back to the file browser

I went with this option by not dispatching `datasetActions.Actions.open`.

PR #667, opened 2020-03-06
Script to create formatted author list; reorder creators.json

uellue:
* Ordering following authorship policy

## Contributor Checklist:


sk1p:
From offline discussion, as a reminder: should be possible to generate and display HTML instead of single superscript ASCII chars, which would prevent inconsistent displaying / font rendering issues when copying to other programs.

Issue #679, opened 2020-03-14
Run CI with Mac OS X because of strange difference in NumPy behavior.


uellue:
When working on allowing to specify zero shifts with the blobfinder UDFs https://github.com/LiberTEM/LiberTEM-blobfinder/pull/23, I ran into a strange difference in NumPy behavior between Linux and OS X. Reporting here since it concerns our general build infrastructure and has potentially wider implications.

This function worked on Linux, but somehow modified the view or shape of the underlying array on OS X to a sub-array when views were set in `process_frame()`. It would then return only one number in the next round instead of a `(y, x)` pair, triggering failure. `zero_shift` is AUX data, `kind='nav', extra_shape=(2, )`.

```python
def get_zero_shift(self, index=None):
    if self.params.zero_shift is None:
        result = np.array((0, 0))
    elif index is None:
        # Called when masked with view
        ### This triggered the error, apparently
        result = self.params.zero_shift[:]
    else:
        # Called when not masked, in postprocess() etc.
        result = self.params.zero_shift[index]
    return result
```

This code worked:

```python
def get_zero_shift(self, index=None):
    if self.params.zero_shift is None:
        result = np.array((0, 0))
    elif index is None:
        # Called when masked with view
        ### This didn't trigger the error
        result = self.params.zero_shift
    else:
        # Called when not masked, in postprocess() etc.
        result = self.params.zero_shift[index]
    return result
```

I couldn't properly reproduce and debug it since it happened on someone else's laptop and there wasn't enough time. The unit tests passed, strangely. I'd like to work on reproducing this with a test case that runs in CI on OS X since this could potentially be a larger issue -- NumPy bug, compiler bug etc.

OS X version unknown (I'd have to ask), Python 3.7.7 from python.org in `python3 -m venv` virtual environment with fresh installation of all packages from pip.

sk1p:
Interesting. I guess the easiest way would be to add copies of the `TOXENV=py3{6,7}` jobs with `os: osx` set. Maybe even specifying a few different `osx_image` combinations (for testing different major OS X releases). Not sure how the Python is installed on travis though, if that is via the python.org package or via hombrew or similar.

uellue:
Closing here since we now test on OS X #681. I'll take care of the LiberTEM-blobfinder specific issue separately. Ref: https://github.com/LiberTEM/LiberTEM-blobfinder/pull/23

PR #681, opened 2020-03-17
Add OS X tests

sk1p:
See also: #679

TODO:
- [x] `large_raw` fixture: creating sparse files doesn't work like this on OS X, or the stat call works differently
- [x] asyncio failures: 
  - [x] `RuntimeError: There is no current event loop in thread 'ThreadPoolExecutor-34_0'`
  - [x] `OSError: [Errno 24] Too many open files`
- [x] numerics fixes:
  - [x] radial fourier
  - [x] feature vector: misses half the peaks?

Radial fourier, both for `JOB` and `UDF`:
```
>       assert np.allclose(np.angle(c_1_2[1, 0]), np.pi)
E       assert False
E        +  where False = <function allclose at 0x10e45d6a8>(-3.141592653589793, 3.141592653589793)
E        +    where <function allclose at 0x10e45d6a8> = np.allclose
E        +    and   -3.141592653589793 = <function angle at 0x10e58dd90>((-0.981522951893353-9.77719708014513e-19j))
E        +      where <function angle at 0x10e58dd90> = np.angle
E        +    and   3.141592653589793 = np.pi
```

Feature vector: 

```
>       assert np.all(result['feature_vec'].data[0:3, 0:2])
E       assert False
E        +  where False = <function all at 0x10e445510>(array([[ True, False],\n       [ True, False],\n       [ True, False]]))
E        +    where <function all at 0x10e445510> = np.all
```

## Contributor Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ cache:
   - pip
   - directories:
     - client/node_modules/
-    - $HOME/.cache/pyenv
 
 install:
   - pip install tox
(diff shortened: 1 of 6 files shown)


sk1p:
Increased limit of open files, let's see if that is fixed. I'm now completely skipping the `large_raw` test, as support on OS X is spotty: only very old or quite new file systems support sparse files, and apparently not the same way as other posix systems...

sk1p:
All asyncio related errors are fixed by the `ulimit -n` change, only numerics related stuff is missing. @uellue could you take care of those? Feel free to push into my branch!

sk1p:
Btw. I just ran the tests locally (on the `web-api-more-state` branch) - I also got the feature vector failure. This is on Linux, using Python 3.7.7

PR #683, opened 2020-03-17
Add pytest-xdist

### LINKED ISSUE #682: Speed up tests by using pytest-xdist
Installing `pytest-xdist` and using `pytest -n auto tests/` speeds up test runs almost by a factor of 2, this can possibly also improve CI turnaround. Need to check if coverage reporting still works with `xdist` enabled.

sk1p:
Closes #682

### DIFF
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -1,5 +1,6 @@
 pytest>=5,<6
 pytest-cov
 pytest-asyncio
+pytest-xdist
 aiohttp
 websockets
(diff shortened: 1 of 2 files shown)


sk1p:
Has little effect on CI runtime, as we only get two worker processes (I guess only 1 core w/ two hyper-threads?)

sk1p:
Merging this, as it brings considerable improvements when running the whole suite locally.

Issue #602, opened 2020-02-14
Performance regression for disk analysis with remote Dask cluster


uellue:
Using a remote Dask cluster running a disk analysis leads to a massive performance regression in comparison to a plain mask analysis.

Setup: Dedicated dask cluster running on moellenstedt. connection to the cluster through VPN and SSH tunneling from a Windows machine over the internet, and locally from moellenstedt itself over TCP to localhost for comparison purposes.

Versions: LiberTEM (nearly) master, dask and distributed 2.10.1 /2.10.0 (latest)

![image](https://user-images.githubusercontent.com/25689052/74548347-f170d080-4f4d-11ea-96ce-21a9ce1c9248.png)

The same effect is observable in the GUI, both with `libertem-server` running on moellenstedt and on a remote Windows machine. 

Connecting to the same cluster from a Jupyter notebook on moellenstedt in the same Python environment triggers the error as well, meaning this is not a cross-platform or version mismatch issue.

An excessive size of the pickled tasks can be excluded judging from the output of debugging code. The regression occurs on repeated execution. The CPU is busy during that time, judging from the output of `top`.

It does not occur when running the same test on the same machine with a local cluster directly with JupyterHub. The performance with a local cluster is 4x (mask analysis) to 50x (disk analysis) better compared to the remote cluster over TCP to localhost.

![image](https://user-images.githubusercontent.com/25689052/74548547-562c2b00-4f4e-11ea-9cd7-c3d1fb1b1c2d.png)

It also doesn't occur when starting libertem-server with a local cluster on moellenstedt.

uellue:
- [ ] Test with stand-alone Dask, minimal reproducer
- [ ] Bisect

uellue:
The regression was caused by not setting the thread count for the stand-alone workers correctly. Refs #546

sk1p:
Ugh, okay. Well done tracking this down, thanks!

uellue:
@sk1p Well, I dug out a new problem in the process: #684 🙄

Issue #546, opened 2020-01-10
Improve control over number of threads for numerics libraries


uellue:
Currently, a script or process that uses LiberTEM should set the thread count to one for numerics libraries by setting appropriate environment variables before loading these libraries, as demonstrated in https://libertem.github.io/LiberTEM/api.html#basic-example

The current approach has a number of disadvantages:
* If LiberTEM is embedded in a larger application, this may have to happen very, very early in the start-up process of this application.
* A Jupyter notebook or embedding application may want to use threads for these numerics libraries when it is using them directly and not through LiberTEM
* The LiberTEM control node may benefit from threading in a cluster set-up.
* Depending on the processing, it may be efficient to mix threading and multiprocessing on the workers. In particular, this can use the L3 cache more efficiently if the workload is dominated by processing tasks that are using the libraries that are affected by the threading settings. That includes a wide range of linear algebra tasks, such as the solvers for grid matching.

In the future, the number of threads for relevant numerics libraries on the workers should be controllable independent from the number of threads on the control process or embedding application.

Ideally, a UDF should be able to signal how many threads it would like to use, and the back-end should adapt the setting based on that.

sk1p:
This can be quite challenging, as it depends on the numerics libraries used as a backend (see also [this numpy discussion](https://github.com/numpy/numpy/issues/11826)). Besides the environment variable, most of these support some kind of API call to control the threading. Let's try to compile a list of these here:

 * OpenBLAS: you can call a C library function called [openblas_set_num_threads](https://github.com/xianyi/OpenBLAS/blob/develop/USAGE.md)
 * Intel MKL: uses both OMP and mkl-specific mechanisms to control number of threads (both using function calls and env vars), [with mkl specifics taking precedence over OpenMP](https://software.intel.com/en-us/mkl-macos-developer-guide-techniques-to-set-the-number-of-threads)
 * PyFFTW: number of threads can be passed [as an argument to fft functions](https://hgomersall.github.io/pyFFTW/pyfftw/interfaces/interfaces.html#additional-arguments), or via the `pyfftw.config.NUM_THREADS` configuration.
 * numpy fft? I think it is single threaded only
 * numba? depends on threading backend?

Anything I'm forgetting?

I found a partial solution in https://github.com/joblib/threadpoolctl - it can control BLAS backends OpenBLAS/BLIS/MKL - I don't think it has support for `numba` or `pyfftw`.

[...]

PR #685, opened 2020-03-17
Set thread count dynamically for each UDF run

uellue:
Having to set the thread count through environment variables can be problematic in some situations #546 and is an error source with severe performance penalties from over-subscription on machines with many cores #602.

In this PR I am planning to include support for setting the thread count to 1 on the workers, i.e. not having to set environment variables anymore. Allowing multi-threaded UDFs should likely be implemented in a follow-up PR later.

With this feature in mind, I am already setting the thread count in `run_for_partition()`.

- [x] Basic proof of concept for OpenBLAS
- [x] Try `threadpoolctl`
- [ ] Intel MKL -- Should be covered by `threadpoolctl`, but couldn't get LiberTEM installed in their Python distribution.
- [x] PyFFTW -- not tested if effective yet, but default thread count is 1 anyway
- [ ] Numba will have a convenient interface in the 0.49 release: https://numba.pydata.org/numba-doc/dev/user/threading-layer.html#setting-the-number-of-threads

Getting the previous number of threads is not always reliable, as it turns out. For that reason I am just setting the number and not trying to restore it after.

## Contributor Checklist:

### DIFF
--- a/docs/source/changelog/misc/threading.rst
+++ b/docs/source/changelog/misc/threading.rst
@@ -1,0 +1,6 @@
+[Misc] Set thread count at run time
+===================================
+
+* No need to set thread count environment variables anymore since the thread count
+  for OpenBLAS, OpenMP, Intel MKL and pyFFTW is now set on the workers at run-time.
+  Numba support will be added as soon as Numba 0.49 is released. (:pr:`685`).
--- a/docs/source/gsoc.rst
+++ b/docs/source/gsoc.rst
@@ -217,20 +217,7 @@ LiberTEM can work as well.
     *Primary contact:* Alex (@sk1p)
 
 
-5. **Beginner/Intermediate**: Set number of threads and workers dynamically for UDFs :issue:`546`.
-    Your task is to implement, document and validate methods to set the thread
-    count for all relevant numerics back-ends in LiberTEM dynamically on a
-    case-by-case basis on the worker processes and main node. Furthermore, you
-    can implement an interface for UDFs to specify their preferred number of
-    threads. This can improve L3 cache efficiency for UDFs that can use a
-    multi-threaded back-end and at the same time avoids oversubscription of the
-    CPU.
-
-    *Skills:* Python programming, using Goole.
-
-    *Domain knowledge:* None
-
-    *Primary contact:* Alex (@sk1p), Dieter (@uellue)
+5. (*Removed since being implemented*: Set number of threads and workers dynamically for UDFs :issue:`546`.)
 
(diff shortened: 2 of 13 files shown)


sk1p:
> Allowing multi-threaded UDFs should likely be implemented in a follow-up PR later.
> With this feature in mind, I am already setting the thread count in run_for_partition().

Sounds sensible! :+1:

> Getting the previous number of threads is not always reliable, as it turns out. For that reason I am just setting the number and not trying to restore it after.

This should be sensible, as long as we never set the number of threads in the main process. Especially important when embedding into another application.

Considering you already put in quite a bit of work, maybe we should remove the idea from our GSoC page? I think only implementing the interface for setting the number of threads in the UDF is a bit small in scope for GSoC, right? Even if it involves some fiddling with the executor, too...

uellue:
Good point, I am removing it from the GSoC page.

uellue:
`threadpoolctl`is confirmed to be effective with OpenBLAS NumPy. The performance of the included notebook on a machine with many cores is a good indicator if it actually works. It will tank by a factor of about 100 when running with 18x36 multithreading.

On my Windows machine I can't test it properly for lack of cores. Furthermore, I am not sure if OpenBLAS actually uses multithreading on my machine since the CPU load doesn't go up to 100 %. At least `threadpoolctl` does show that it changes the setting for number of threads from 4 to 1 if I instruct it to.

I ran into difficulties when trying to get the Intel Python distribution running. I couldn't  update the pre-installe packages to the required versions for LiberTEM. Unless we have someone using Intel MKL + Python, I'd tent to trust `threadpoolctl` for now and fix issues as they appear.

PyFFTW should be tested properly as soon as we support UDFs that use more than one thread since the default is one thread anyway.

Numba requires the upcoming 0.49 for a clean interface.

Since `threadpoolctl` already offers a context manager and the other two cases have a clean, reliable Python interface to get and set the number of threads, we can use a context managers for all.

sk1p:
> Unless we have someone using Intel MKL + Python, I'd tent to trust threadpoolctl for now and fix issues as they appear

:+1:

> Numba requires the upcoming 0.49 for a clean interface.

Good that it's in the works!

sk1p:
Looks good, merging!

sk1p:
Do we have an issue for the numba part of the TODO? 0.49.1 was just released which means there should be a stable and installable version of numba available which has the thread control interface. Not urgent, just making sure we don't lose track of it.

uellue:
Why open an issue if you can open a pull request? 😁

PR #686, opened 2020-03-18
Not necessary to set environment variables anymore

uellue:
Thx @sk1p , good catch!

Refs #546 #685 

## Contributor Checklist:


sk1p:
Thx!

PR #690, opened 2020-03-19
Don't set as default Dask cluster

uellue:
Changing the default Dask cluster can have unwanted and unexpected side
effects for users.

## Contributor Checklist:


sk1p:
Thanks, merged!

PR #694, opened 2020-03-26
Buffer

saisunku:
## Contributor Checklist:

* [X] I have added [a changelog entry](https://github.com/LiberTEM/LiberTEM/tree/master/docs/source/changelog) for my contribution
* [X] I have added/updated documentation for all user-facing changes
* [X] I have added/updated test cases

### DIFF
--- a/docs/source/changelog/bugfix/buffer_extra_shape.rst
+++ b/docs/source/changelog/bugfix/buffer_extra_shape.rst
@@ -1,0 +1,5 @@
+[Bugfix] Allow BufferWrapper to accept a Shape object
+===========================================================
+
+* Both tuples and Shape objects can be used as extra_shape now
+* Added a test for each case to test_bufferwrapper.py
--- a/docs/source/changelog/bugfix/buffer_extra_shape.rst
+++ b/docs/source/changelog/bugfix/buffer_extra_shape.rst
@@ -1,5 +1,5 @@
 [Bugfix] Allow BufferWrapper to accept a Shape object
 ===========================================================
 
-* Both tuples and Shape objects can be used as extra_shape now (:pr:`640`)
-* Added a test for each case to test_bufferwrapper.py (:pr:`640`)
+* Both tuples and Shape objects can be used as extra_shape now (:pr:`694`)
+* Added a test for each case to test_bufferwrapper.py (:pr:`694`)
--- a/docs/source/changelog/bugfix/buffer_extra_shape.rst
+++ b/docs/source/changelog/bugfix/buffer_extra_shape.rst
@@ -1,5 +1,5 @@
 [Bugfix] Allow BufferWrapper to accept a Shape object
 ===========================================================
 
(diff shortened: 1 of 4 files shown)


src/libertem/common/buffers.py
```diff
@@ -62,17 +63,24 @@ def __init__(self, kind, extra_shape=(), dtype="float32"):
...
+            For a Shape object, sig_dims is discarded and the entire shape is used.
 
         dtype : string or numpy dtype
             The dtype of this buffer
         """
+
+        if isinstance(extra_shape, Shape):
+            self._extra_shape = extra_shape._nav_shape + extra_shape._sig_shape
+        else:
+            self._extra_shape = extra_shape
```
sk1p: Even though it works, accessing internal attributes of other classes (starting with an underscore) should not be done - they are used to separate the interface from implementation details (in other languages, there are stricter mechanisms for ensuring this, like visibility specifiers `private` etc.). Imagine the `Shape` implementation is changed to store the whole shape as a single tuple (or even another type: list, ndarray?) - you would have to change `_nav_shape` and `_sig_shape` usages all over the code base.

There is an even better way to do the same: `Shape` implements `__iter__`, and  any iterable can be converted to a `tuple` using the `tuple()` built-in function. So there is also no need for the `isinstance` check anymore - just unconditionally call `tuple` on `extra_shape`.

That means we accept any iterable, not just tuple and `Shape`, [which is even better IMHO.](https://en.wikipedia.org/wiki/Robustness_principle)

docs/source/changelog/bugfix/buffer_extra_shape.rst
```diff
@@ -0,0 +1,5 @@
+[Bugfix] Allow BufferWrapper to accept a Shape object
+===========================================================
+
+* Both tuples and Shape objects can be used as extra_shape now
```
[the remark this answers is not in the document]
saisunku: Done

src/libertem/common/buffers.py
saisunku: Thanks for the detailed explanation, I've updated accordingly

sk1p:
Hm, for some reason the travis build doesn't show up in the PR, sorry for that. This is the associated travis build:

https://travis-ci.org/github/LiberTEM/LiberTEM/builds/668279968

Note that [the flake8 job](https://travis-ci.org/github/LiberTEM/LiberTEM/jobs/668279974) failed - there are a few style fixes to be done. Please see our [contributing guideline](https://libertem.github.io/LiberTEM/contributing.html#code-style) on how to run them locally.

The mypy job that also failed is allowed to fail - don't worry about that one. Other than that, it looks ready to merge!

saisunku:
Thanks! I had some issues with `tox`, so I ran `pytest` manually which didn't include `flake8`. Fixed now.

sk1p:
Thanks, merged!

Issue #601, opened 2020-02-13
Allow several attempts to specify a remote cluster


uellue:
How to reproduce: Connect to a remote cluster in the GUI with the wrong URL (cluster not running in this case).

What should happen:
* Error message
* Try again

What happens:
* Error message on the console, not in the GUI
* Input box still visible, but "connect" button faded out
* Not possible to try again

Edit: Reloading the browser window works as a work-around.

twentyse7en:
What are the prerequisite to solve this issue ?

sk1p:
> What are the prerequisite to solve this issue ?

In this case, a prerequisite to fixing the bug is building the client from source, as there most likely need to be changes in the TypeScript client code. You may need to know the basics of some libraries used in the client, for example react and possibly redux. You may need to understand the very basics of backend and front end web dev, for debugging and understanding what happens when the user clicks the connect button etc.

Were you able to reproduce the bug?

twentyse7en:
> 
> Were you able to reproduce the bug?

Yeah, I did reproduce.

I'm not familiar with web dev. But I would love to solve this bug, I think knowing web dev would be helpful for GSOC project.

sk1p:
We have a small section on [debugging the web GUI](https://libertem.github.io/LiberTEM/debugging.html#debugging-the-web-gui) in our documentation, which has some pointers to indispensable debugging tools (browser's devtools, react and redux plugins for these, ...), which should help you trace what is going on.

As a hint on how to start with this, I would suggest to try to understand what exactly, on client and server, step by step, is happening in the "happy case", that is, in case the user presses the "Connect" button and either connects to an existing dask cluster or successfully creates a local cluster (the two cases are in general very similar). Let me know if you get stuck or need help to understand any specific part!

twentyse7en:
Thanks :+1:

twentyse7en:
I tried to figure out how "happy case" works. The client makes a connection request and server returns status `ok`. Then it makes action `connected`. I didn't understand how "browser" button showed after that.

Also in the "bad case" server returns `500: Internal Server Error`.

sk1p:
> I tried to figure out how "happy case" works. The client makes a connection request and server returns status `ok`.

[...]

PR #699, opened 2020-04-02
To handle cluster connection fail in GUI

twentyse7en:
fix issue #601 
## TODO
* [x] Handle connection failure in server
* [x] Display error message
* [x] <strike> Reset submit button</strike> Show spinner while connecting to server
 
## Contributor Checklist:


sk1p:
Closing and re-opening PR to try and fix travis status

src/libertem/web/connect.py
```diff
@@ -39,9 +39,13 @@ def initialize(self, state: SharedState, event_registry):
...
-            sync_executor = await sync_to_async(partial(DaskJobExecutor.connect,
-                scheduler_uri=connection['address'],
-            ))
+            try:
+                sync_executor = await sync_to_async(partial(DaskJobExecutor.connect,
+                    scheduler_uri=connection['address'],
+                ))
+                connected = True
+            except OSError:
+                connected = False
```
sk1p, line 48: The `connected` approach here works, but makes it a bit hard to follow the execution of the code. What do you think about just writing the failure response here, and returning? It also means you keep the context of the exception when returning the error response (see below)

src/libertem/web/connect.py
```diff
@@ -57,23 +61,31 @@ def initialize(self, state: SharedState, event_registry):
...
+            await self.send_existing_job_results()
+            self.write({
+                "status": "ok",
+                "connection": connection,
+            })
+        else:
+            self.write({
+                "status": "disconnected",
+                "connection": {},
+            })
```
sk1p, line 90: This could (should?) also return the concrete error message from the exception - possibly also using `"status": "error"` instead? See also `messages.py`, where several `_error` messages are defined. If you like, you can also extract the response into a method of the `Messages` class, so it won't be forgotten once the schema project is implemented.

src/libertem/web/connect.py
twentyse7en: Yep! that would be very clean and clear.

twentyse7en:
>
>The GUI properly goes back to the "not connected" status after an error, right? 

Yes
>
>Do you want to add error displaying functionality to the GUI, or should that be done as a follow-up PR?

Will do it here, If there are no other problems : )

sk1p:
> Will do it here, If there are no other problems : )

:+1:

tests/server/test_startup.py
```diff
@@ -93,6 +93,21 @@ def _get_raw_params(path):
...
 
+@pytest.mark.asyncio
+async def test_cluster_connect_error(base_url, http_client):
+    url = "{}/api/config/connection/".format(base_url)
+    conn_details = {
+        'connection': {
+            'type': 'TCP',
+            'address': 'tcp://unknown:8786',
+        }
+    }
```
sk1p, line 104: :+1: - instead of "unknown", you could also [use the "invalid" name that is guaranteed to not resolve to any hostname](https://tools.ietf.org/html/rfc6761#section-6.4) (RFC 6761). It should also guarantee that the name resolution fails early, so the error should happen earlier ("Name resolution APIs and libraries SHOULD recognize "invalid" names as special and SHOULD always return immediate negative responses.  Name resolution APIs SHOULD NOT send queries for "invalid" names to their configured caching DNS server(s).")

Maybe this will also fix [the timeout issue here](https://travis-ci.org/github/LiberTEM/LiberTEM/jobs/670861793?utm_medium=notification&utm_source=github_status)?

twentyse7en: "invalid" address is raising a `ValueError: missing port number in address 'invalid'`. It only takes around 8 ms to respond. For "invalid:8080" it takes 10s to respond.

sk1p: Very nice - yeah I think we can't get around the 10s timeout without doing something funky (mocking the `Executor` comes to mind...) - I guess that's a retry-loop somewhere in Dask that retries even the DNS lookup :grinning: 

I like the new solution, let's keep it :+1:

twentyse7en:
How can I display the error message, there is no action for error in the cluster.

sk1p:
> How can I display the error message, there is no action for error in the cluster.

Yeah, error displaying implementation in the GUI is a bit rough, it needs a refactor I think. For now, you can go ahead and add a "cluster error" action, add that to the `errorReducer` and be done with it.

twentyse7en:
I'm stuck :crying_cat_face: , what to do next!

sk1p:
> I'm stuck, what to do next!

One thing to always do is to keep the CI pipeline happy and green. In this case, the [typescript build and lint job](https://travis-ci.org/github/LiberTEM/LiberTEM/jobs/672114639) is easily fixed I think - not sure what the problem with [the other job is](https://travis-ci.org/github/LiberTEM/LiberTEM/jobs/672114631) - possibly an intermittent problem? (For TypeScript: see the "Organize Imports" command of vs code, if you are using that!)

Then, if you are stuck, sometimes it is good to take a step back and have a look at the original problem again. What were you trying to accomplish? What were the parts of the whole change, and which parts are already done? What is missing? If you trigger the new behavior, what is happening, what should happen? Maybe also re-read some of the previous discussion and issues. This is most of the time enough for me to get unstuck. If you still need guidance then, please let me know!

I also like to make a checklist in the process, which can then remind me later what is still to be done.

twentyse7en:
Resolved the lint error. In the previous commit forget to build the client :sweat_smile: how can I resolve conflict?

I have added `clusterError` action, also added to `errorReducer` and dispatch action when the connection fails in `sagas.ts`. But the error message isn't popping up : /

sk1p:
> Resolved the lint error

:+1:

> I have added `clusterError` action, also added to `errorReducer` and dispatch action when the connection fails in `sagas.ts`. But the error message isn't popping up : /

Hmm. Have you verified that the action is dispatched? This can be done with the redux dev tools. There you can also have a look at the action and see if it contains the data you expect it to contain. You can also inspect the complete redux state before/after the action is dispatched. Let me know if this helps!

> how can I resolve conflict?

Do you know how this is usually done, when you get a conflict in a rebase? The issue here is that the conflicts are in generated code - so it would be hard to manually resolve this. What I usually do is I start to rebase to master as usually, but then when there are conflicts in those `web/client/**` files, I just rebuild the client again, which should make sure that your changes take effect. Then I do `git rebase --continue` as usual...

I'm sure there is a better way, but that's how I do it :grinning: 

If you want to practice this, I would highly suggest to just create a new branch from the current one as backup, so if you mess up, you can just go back to the old branch.

twentyse7en:
error_handle

Action is dispatched, where it should receive this information and display the message? I'm a beginner in react, can you go through the code and give the necessary suggestions. I did try different ways but nothing works.

I have never resolved conflicts. Thanks for the suggestion : )

sk1p:
> where it should receive this information and display the message?

This, or similar, is how an error message is supposed to look like:

image

(You can easily trigger this one yourself by entering a non-existing path into the file browser, to compare with what happens with your own error.)

Then, you can select your `CLUSTER_ERROR` action and check the "State" tab. If you expand the `errors` key there, you should see your error message:

image

The "Diff" tab should also show that the `errors` key changed. If this is not the case, maybe your change in the reducer doesn't work for some reason?

Let me know if this helps, if you are still stuck I can check out your branch and try to help debug this.

twentyse7en:
Screenshot from 2020-04-09 13-03-05

Here I tried to connect to the cluster first, so it dispatched a `cluster error`. Sadly no error message popped up. Then connected to local cluster. It was successful, now the previous error message is showing.I'm still stuck can you check my branch?

sk1p:
Hm, I'll have a detailed look later, but: is it possible that the error messages are not even shown before the cluster connection is there? Hint: have a look at `ErrorList.tsx`! That may need changes...

twentyse7en:
In `ErrorList.tsx`

`
const mapStateToProps = (state: RootReducer) => {
    return {
        errors: state.errors,
        clusterConnected: (state.clusterConnection.status === "connected" ||
            state.clusterConnection.status === "disconnected"),
        channelConnected: (state.channelStatus.status === "connected" ||
            state.channelStatus.status === "ready"),
    }
}
`
Now it's working!! Is that ok or do we need any other changes?

sk1p:
> `clusterConnected: (state.clusterConnection.status === "connected" || state.clusterConnection.status === "disconnected")`

Hm, that's a bit misleading, as `clusterConnected` will be `true`, even though the cluster connection status is `"disconnected"`. Maybe you can have a look where the variable is used, and change that instead?

twentyse7en:
Yeah! why was there a constrain to show error only after the cluster connection is successful ?

sk1p:
> Yeah! why was there a constrain to show error only after the cluster connection is successful ?

I'm not sure - possibly in an older version there were spurious errors in that state. Did you get any other errors when testing? You can try to remove that constraint - if any spurious errors show up, we can fix those in a different way.

twentyse7en:
I removed the constrain and its working fine : )

sk1p:
For the last point on your list, a small hint: https://jaredpalmer.com/formik/docs/1.5.8/guides/form-submission

(note that we are still using an earlier version of formik, as we have not migrated to 2.x yet)

twentyse7en:
Thx :+1: So when the server responds with an error, by calling `setSubmitting(false)` the form can be reset. We are making the request in `api.ts`, is this the place to handle this?

sk1p:
> Thx  So when the server responds with an error, by calling `setSubmitting(false)` the form can be reset. We are making the request in `api.ts`, is this the place to handle this?

Hmm, looking at it again: `setSubmitting(false)` is only available in the form itself and `handleSubmit` function etc. -  it can't easily be called from a saga or from inside the API helper function. (Aside: API helper functions like `connectToCluster` are supposed to do only one thing, and that is to speak with the HTTP API. Ideally other things, like interacting with forms, would be done by the callers of those functions.)

Maybe it is easier to introduce a new cluster status `"connecting"`, replace the whole form with a spinner (like the `channel` status is handled - you can test this by stopping the libertem server but keeping the browser tab open - it should show the spinner and try to reconect.)

An alternative would be to completely take the cluster state out of redux/redux-saga and into a react component. I think this may be harder for you to do, because this would change the way the state is handled completely.

twentyse7en:
I have no idea why tests are failing! @sk1p can you have a look

sk1p:
> I have no idea why tests are failing! @sk1p can you have a look

Sorry, yeah, these appear to be unrelated to your pull request. I've opened a separate issue to track this: https://github.com/LiberTEM/LiberTEM/issues/713

twentyse7en:
>
>If a user runs into this bug, they probably want to re-try with similar parameters, for example if they had a typo in the URL, or the remote cluster was not running yet. So it would be really nice if the form re-opened with the original parameters.

Sounds good! I'll try to implement that :+1:

PR #700, opened 2020-04-02
Update detect function and add dropdown for HDF5 ds_paths

### LINKED ISSUE #692: Allow ds_path to be chosen from a dropdown in the HDF5 form in GUI
Currently when loading a HDF5 file, the 'HDF5 Dataset Path' is initially set to the largest dataset in the file. It'd be helpful if the textbox were replaced with a dropdown with multiple dataset's paths.

AnandBaburajan:
Fixes #692 and updates datasets' `detect` function and `libertem.io.dataset.detect`. 

## Contributor Checklist:

changes were made to the GUI)


client/src/dataset/components/DatasetParams.tsx
```diff
@@ -16,12 +16,15 @@ const renderParamValue = (value: any) => {
 
 const renderParams = (params: any) => {
     return Object.keys(params).map((key: string, idx: number) => {
-        return (
-            <Table.Row key={idx}>
-                <Table.Cell>{key}</Table.Cell>
-                <Table.Cell>{renderParamValue(params[key])}</Table.Cell>
-            </Table.Row>
-        );
+        if(key !== "dataset_paths") {
```
AnandBaburajan, line 19: I've removed `dataset_paths` from the info as:
1. The `ds_path`s are shown as [[object Object], [object Object], ...]
2. All the `ds_path`s are already specified below (in the info box) individually

AnandBaburajan:
@sk1p Should I fix the test failures now or should I wait for your suggestions on the code?

sk1p:
> @sk1p Should I fix the test failures now or should I wait for your suggestions on the code?

I'll give you some short feedback, sorry don't have much time today.

What do you think about, instead of putting `dataset_paths` into the same "namespace" as the parameters, to create an explicit namespace for this? So instead of returning
```python
{
"ds_path": "something",
"tileshape": [...],
"dataset_paths": {....},
}
```

The detect function could be changed to return something like:

```python
{
"parameters": { ... see above ...},
"info": { "dataset_paths": { ... } }
}
```

Then, other users (like the `"auto"` load functionality, or `libertem.io.dataset.detect`) can specifically select only the parameters which can directly be passed on to the dataset, and LiberTEM knows what to display as parameters and what is "additional info". What do you think?

AnandBaburajan:
> Then, other users (like the `"auto"` load functionality, or `libertem.io.dataset.detect`) can specifically select only the parameters which can directly be passed on to the dataset, and LiberTEM knows what to display as parameters and what is "additional info". What do you think?

Sounds exactly what is needed to help with the issues!

> instead of putting dataset_paths into the same "namespace" as the parameters, to create an explicit namespace for this?

I'll try that out, thanks!

AnandBaburajan:
@sk1p Here's what I found out:

> Then, other users (like the "auto" load functionality, or libertem.io.dataset.detect) can specifically select only the parameters which can directly be passed on to the dataset

For the GUI to receive the dropdown options, `libertem.io.dataset.detect` needs to return `dataset_paths`. So the 'selection' can't be done there.

> Then, other users (like the "auto" load functionality, or libertem.io.dataset.detect) can specifically select only the parameters which can directly be passed on to the dataset

That would require changes in:
1. the `detect_params` function of all the datasets
2. all the functions which use `libertem.io.dataset.detect`
3. the way the GUI accesses the parameters (Example: [HDF5ParamsForm.tsx](https://github.com/LiberTEM/LiberTEM/blob/master/client/src/dataset/components/HDF5ParamsForm.tsx#L49))

I think that would require a lot of changes to the current structure.

So instead of selecting the parameters to be passed to the dataset, can we simply remove `dataset_paths` from the detected parameters in `auto_load`? So that `dataset_paths` isn't passed down to the dataset as a parameter and the GUI still gets the dropdown options.

Please let me know your feedback, thanks.

sk1p:
> For the GUI to receive the dropdown options, `libertem.io.dataset.detect` needs to return `dataset_paths`. So the 'selection' can't be done there.
> 
> That would require changes in:
> 
> 1. the `detect_params` function of all the datasets
> 2. all the functions which use `libertem.io.dataset.detect`
> 3. the way the GUI accesses the parameters (Example: [HDF5ParamsForm.tsx](https://github.com/LiberTEM/LiberTEM/blob/master/client/src/dataset/components/HDF5ParamsForm.tsx#L49))
> 
> I think that would require a lot of changes to the current structure.

Yes, that's true. I think mostly 2) is a problem - that's a function that should ideally be kept backwards compatible.

I think 1) and 3) are inevitable. This actually relates to some of the use cases we talked about with regard to your GSoC project: to properly select the shape and possibly number of frames to ignore at the beginning/end of the dataset, the user may need a bit more information about the file they are currently opening - for example "number of frames" would be good to know. That is something the detect function could return, if possible.

Instead of 1), *maybe* it would be possible to add another method to the `DataSet`, but I think it fits quite well into the detect method.
For 2), a possibility is to add a new parameter, something like `only_parameters`, which defaults to `True`, so existing users can get the current behavior. If `only_parameters` is set to `False`, then the return type could be changed, to:

```python
{
"type": "hdf5", # for example
"parameters": { ... see above ...},
"info": { "dataset_paths": { ... } }
}
```

(Note that I also moved the `type` key from `parameters` to the top level, as most callers of detect just `pop` it out of the dict anyways ([example 1](https://github.com/Brow71189/libertem-hackaton/blob/master/nionswift_plugin/LiberTEM_IO/__init__.py#L78), [example 2](https://github.com/LiberTEM/LiberTEM/blob/70198d238aaa114e3f86a49e061f3caa181aac0b/src/libertem/io/dataset/__init__.py#L29)) - if this doesn't work well for the typescript side of things, I'm open to add a "type tag" in parameters, too, but I'd do it [only in the web API](https://github.com/LiberTEM/LiberTEM/blob/master/src/libertem/web/dataset.py))
 
> So instead of selecting the parameters to be passed to the dataset, can we simply remove `dataset_paths` from the detected parameters in `auto_load`? So that `dataset_paths` isn't passed down to the dataset as a parameter and the GUI still gets the dropdown options.

I dislike this, because `dataset_paths` is an implementation detail of the HDF5 dataset, and the `auto_load` function should not even know about it (["separation of concerns"](https://en.wikipedia.org/wiki/Separation_of_concerns)). If we implemented it in this fashion, after a while, there would be a list of parameters to ignore, from various data sets, and any time any data set decides to change parameters, `auto_load` also has to change. This is a violation of the ["Single-responsibility principle"](https://en.wikipedia.org/wiki/Single-responsibility_principle) - `auto_load` should, if possible, only change if the auto detection feature itself changes in some way.

Also, other users of `detect` would also have to maintain this list (see [example 1 above again](https://github.com/Brow71189/libertem-hackaton/blob/master/nionswift_plugin/LiberTEM_IO/__init__.py#L94) - the parameters from `detect` are passed on to `load`, exactly like in `auto_load`)

I hope this helps, and justifies investing a bit more work to do things properly.

AnandBaburajan:
> Note that I also moved the `type` key from `parameters` to the top level

I didn't find the `type` key in the parameters, only this [line](https://github.com/LiberTEM/LiberTEM/blob/master/src/libertem/io/dataset/__init__.py#L113) which adds `type` to the parameters.

@sk1p Your feedback please, thanks!

sk1p:
> I didn't find the `type` key in the parameters, only this [line](https://github.com/LiberTEM/LiberTEM/blob/master/src/libertem/io/dataset/__init__.py#L113) which adds `type` to the parameters.

Yeah, that's the one. The idea is: currently, to open a data set from detected parameters, because the type is just put into the returned params dict, the pattern is:

```python
params = detect(...)
filetype = params.pop("type")
ds = load(filetype, **params)
```

Now, if we just add more keys with additional info to the returned params dict, this becomes:

```python
params = detect(...)
filetype = params.pop("type")
del params['info_key_1']
del params['info_key_2']  # etc - different keys here depending on the concrete dataset!
ds = load(filetype, **params)
```

That's why I'm proposing to separate the parameters from the other fields, to create separate name spaces.

src/libertem/io/dataset/blo.py
```diff
@@ -164,17 +164,22 @@ def initialize(self, executor):
...
+            parameters = {
                 "path": path,
                 "tileshape": (1, 8) + tuple(ds.shape.sig),
                 "endianess": "<",
             }
+            additional_info = {}
+            # If additional info is required
+            if info:
+                parameters.update(additional_info)
+            return parameters
```
sk1p, line 182: Hmm, to make the datasets easier, what do you think about always returning the info? And then decide in `detect` if we want to return it or not. Then we have a single place where the decision is made, and not one in each dataset.

AnandBaburajan:
> Now, if we just add more keys with additional info to the returned params dict, this becomes:
> 
> ```python
> params = detect(...)
> filetype = params.pop("type")
> del params['info_key_1']
> del params['info_key_2']  # etc - different keys here depending on the concrete dataset!
> ds = load(filetype, **params)
> ```

@sk1p But the updated code separates the params from the additional info and there won't be any `info_key`s to delete in the params, unless `info`=true.

I think I understood what we're trying to achieve, but sorry if I still haven't got the right idea. Can you briefly tell me how your idea is different from the code? Thanks for your guidance again!

src/libertem/io/dataset/blo.py
AnandBaburajan: Then what do you suggest I do when the dataset's `detect_params` is called directly? [Example](https://github.com/LiberTEM/LiberTEM/blob/master/tests/io/test_hdf5.py#L247)

sk1p: Those places need to be updated - other than tests, `detect_params` should only be called in the `detect` function AFAIK

sk1p:
> But the updated code separates the params from the additional info and there won't be any info_keys to delete in the params, unless info=true

But with `info=True` I now have to delete all other keys from the result if I just want to have the info

> Can you briefly tell me how your idea is different from the code?

Your implementation doesn't introduce separate namespaces for info and parameters - the info is just merged into the parameters in the `detect_params` methods. Which means I can't separate info from parameters without knowing which keys belong to which part. Which then means anything that tries to handle `info` in any way needs to be dataset specific - but it could as well be generic and used for all data sets in many cases.

src/libertem/io/dataset/hdf5.py
```diff
@@ -160,17 +160,26 @@ def _do_detect():
         # try to guess the hdf5 dataset path:
         try:
             datasets = executor.run_function(_get_datasets, path)
-            largest_ds = sorted(datasets, key=lambda i: i[1], reverse=True)[0]
-            name, size, shape, dtype = largest_ds
+            datasets_list = sorted(datasets, key=lambda i: i[1], reverse=True)
+            ds_paths = [ds_path[0] for ds_path in datasets_list]
+            # options for the semantic-ui dropdown
+            dataset_paths = [{"key": dpath, "text": dpath, "value": dpath} for dpath in ds_paths]
```
sk1p, line 166: I would prefer if the "formatting" into key/text/value would happen in the TypeScript code - for usage from Python this is redundant information, it's just there because of how the dropdown works

AnandBaburajan:
> But with `info=True` I now have to delete all other keys from the result if I just want to have the info

I see, there are use cases where only info is required. I've got it now, thanks!

src/libertem/io/dataset/hdf5.py
AnandBaburajan: Fixed!

client/src/dataset/components/DatasetParams.tsx
AnandBaburajan: Fixed!

src/libertem/io/dataset/blo.py
AnandBaburajan: Done!

client/src/dataset/components/HDF5ParamsForm.tsx
```diff
@@ -21,7 +21,30 @@ const HDF5ParamsForm: React.SFC<MergedProps> = ({
     handleSubmit,
     handleReset,
     onCancel,
+    setFieldValue,
 }) => {
+
+    const isTimeOut = (values.dataset_paths === "" ) ? true : false;
+    const dsPathOptions = values.dataset_paths.split(", ").map(dsPath => ({ key: dsPath, text: dsPath, value: dsPath }));
```
sk1p, line 28: Adding the redundant keys needed by the `Dropdown` here is the right approach, compared to the previous one where this was done in the Python code :+1:

`dataset_paths` should be a list, though, see the corresponding comment on the Python part.

src/libertem/io/dataset/__init__.py
```diff
@@ -99,18 +100,20 @@ def get_dataset_cls(filetype):
     return cls
 
 
-def detect(path, executor):
+def detect(path, executor, getParams=True, getInfo=False):
```
sk1p, line 103: Hmm. Seeing this now makes me a bit unhappy - having these boolean parameters is a code smell that this function should really be two functions. I originally suggested the `info` parameter because we could keep backwards-compatibility, but I noticed we don't even document the `detect` method yet. Sorry for flip-flopping on this, but I think it is important to get this function right.

@uellue what do you think about breaking backwards-compat. for `detect`, and documenting the new behavior? I think the only outside user is really the nion swift prototype, and that we can easily PR against to keep it up to date.

So the new version would just behave as if both the `get*` parameters were `True`, and not have those parameters.

(Btw. small style comment: parameters in **Python** usually are `snake_case`, not `camelCase`. I know this can be a bit confusing, as TypeScript is the other way around...)

src/libertem/io/dataset/__init__.py
```diff
@@ -25,12 +25,13 @@ def _auto_load(path, executor):
             "please specify the `path` kwarg to allow auto detection"
         )
 
-    params = detect(path, executor=executor)
-    filetype_detected = params.pop('type', None)
+    detected_params = detect(path, executor=executor)
+    filetype_detected = detected_params.pop('type', None)
```
sk1p, line 29: `type` no longer needs to be `pop`ed off - just use `get` instead.

src/libertem/io/dataset/hdf5.py
```diff
@@ -160,16 +160,26 @@ def _do_detect():
         # try to guess the hdf5 dataset path:
         try:
             datasets = executor.run_function(_get_datasets, path)
-            largest_ds = sorted(datasets, key=lambda i: i[1], reverse=True)[0]
-            name, size, shape, dtype = largest_ds
+            datasets_list = sorted(datasets, key=lambda i: i[1], reverse=True)
+            dataset_paths = ", ".join([ds_path[0] for ds_path in datasets_list])
```
sk1p, line 164: Why is `dataset_paths` not a list? Using a single string is problematic, as a "dataset name" in HDF5 can contain special characters:

```python
In [2]: import h5py                                                                                                                                             
In [3]: f = h5py.File("test.h5", "w")                                                                                                                           
In [4]: f.create_dataset("a,b,c", shape=(16, 16), dtype="float32")                                                                                              
Out[4]: <HDF5 dataset "a,b,c": shape (16, 16), type "<f4">
```

Now your dropdown contains three entries...

All other `detect_params` changes look good :+1:

src/libertem/io/dataset/__init__.py
sk1p: Discussion with @uellue: no need for backward compat. - not documented, only used in prototype code

client/src/dataset/components/HDF5ParamsForm.tsx
```diff
@@ -21,7 +21,30 @@ const HDF5ParamsForm: React.SFC<MergedProps> = ({
...
+
+    let dsPathInput;
+    let dsPathDropdown;
+
+    if (isTimeOut) {
+      dsPathInput = <Field name="ds_path" id="id_ds_path" />;
+      dsPathDropdown = null;
+    } else {
+      dsPathInput = <Field name="ds_path" id="id_ds_path" hidden={true} />;
+      dsPathDropdown = <Dropdown name="options" id="id_options" placeholder="Select dataset" fluid={true} search={true} selection={true} defaultValue={values.ds_path} onChange={onDSPathChange} options={dsPathOptions} />;
```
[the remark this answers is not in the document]
AnandBaburajan: > What is the reason for not having the `Dropdown` with `name="ds_path"`

Didn't cross my mind :) Fixed it, and added a comment about the value being manually set.

src/libertem/web/dataset.py
```diff
@@ -93,14 +93,19 @@ async def get(self):
...
+        if not detected_params:
             msg = Message(self.state).dataset_detect_failed(path=path)
             log_message(msg)
             self.write(msg)
             return
-        params['type'] = params['type'].upper()
+        params = detected_params["parameters"]
+        params.update({"type": detected_params["type"].upper()})
+        if "info" in detected_params:
+            params.update(detected_params["info"])
```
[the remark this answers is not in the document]
AnandBaburajan: Sorry, it does seem like a task requiring thorough understanding of the code. I'll surely work on this once I'm a bit more familiar with the codebase!

client/src/dataset/components/HDF5ParamsForm.tsx
AnandBaburajan: Thanks, fixed!

src/libertem/io/dataset/hdf5.py
AnandBaburajan: Fixed!

src/libertem/io/dataset/__init__.py
AnandBaburajan: Done! Please do let me know if there are any occurrences of `detect` to be updated and how I should properly document the function. I'm not aware of the 'nion swift' prototype though.

src/libertem/io/dataset/__init__.py
AnandBaburajan: Fixed!

src/libertem/io/dataset/__init__.py
sk1p: For the nion swift integration prototype, see this repository: https://github.com/Brow71189/libertem-hackaton

sk1p:
Thank you for the PR, merged!

src/libertem/io/dataset/__init__.py
AnandBaburajan: I opened a PR at https://github.com/Brow71189/libertem-hackaton/pull/9.
Are the changes okay?

sk1p: Looks good! Andreas just merged them :tada:

PR #398, opened 2019-07-17
Starting to include correction facilities in LiberTEM

uellue:
See https://github.com/LiberTEM/nexus-4dstem

Optimized version for the perhaps most important correction: Dark frame, gain and faulty pixels. Thx @fwin22 for the feedback to give priority to this!

It achieves between 1 GB/s and 3 GB/s per core on my development system, depending on the number of faulty pixels. It handles edge cases such as adjacent faulty pixels. 

A naive numpy version achieved at most 500 MB/s.

PR #706, opened 2020-04-08
WIP: Tiled reading in more file formats

### LINKED ISSUE #27: Properly handle large frames: sub-frame tiling
Right now, performance is suboptimal once a stack of frames is larger than (rough guess, depends on number of masks) half of the L3 cache. To properly handle large frames, we need to create ``DataTile``s that a) subdivide the frames, and b) still stack a few of these sub-frames on top of each other. Example: let's say we have a dataset with shape ``(128, 128, 4096, 4096)``, dtype=float. Each frame then takes up 64MB. On a system with 8MB L3 cache, only a fraction of a frame can fit into memory. Something like ``(1, 8, 256, 256)`` could still fit the L3 cache though.

In #11 we already had a look at the optimal execution order. We need to implement one of the efficient tilings presented there, throughout the system. 

- [ ] sub-frame tiling for our file format(s)
  - [x] HDF5
  - [ ] HDFS - basically #20 or a custom implementation for our binary file format
- [x] properly handle sub-frame tiling in the mask application task
- [x] proper result handling for sub-frame tiling

On disk the data needs to be in an efficient tiling, otherwise we need to do multiple RAM<>L3 roundtrips for each frame. This part may need to wait until we are using TileDB as our backend, so we are mostly waiting on https://github.com/TileDB-Inc/TileDB-Py/issues/38 there. A remaining question would be: do we want to support this on non-cluster systems? That would mean we need to pre-process even on single node systems (for example have a HDF5 -> HDF5 conversion, or HDF5 -> TileDB, that creates 2MB chunks of shape ``(1, 8, 256, 256)``). Then we would have non-straight-forward chunking for those files...

Result handling should be easy, we basically need to add to the (maybe) existing partial result, instead of only copying into the result array. The ``ResultTile`` needs to be clearly marked as being a partial result.

Comfortably handling sub-frame tiles at the lower level is not so easy, once we go beyond applications that use mask application as their basis. Then we may need to think about boundary handling etc. For now, we need to slice the masks array such that it matches the sub-frame tiling.

### LINKED ISSUE #247: Automatically determine tileshape
The tileshape is one of the more confusing parameters currently when opening a dataset. We should verify that our current assumption (1MB tiles work well universally) is true, and when implementing #198, we should include variations over tileshape into the benchmark suite. It may be a good idea to still offer it as a tuning parameter, as different analyses may have different optimum tile sizes.

Many formats will get this for free from #322, with a stackheight set to fit 1MB. This of course only works for smaller frames, until #331 is fixed.

Format checklist:
- [ ] BLO
- [x] EMPAD (tileshape can't be set at all)
- [ ] HDF5
- [ ] K2IS (fixed to `(stackheight, 930, 16)`, need to automatically set stackheight)
- [x] MIB (tileshape optional, sets stackheight from tileshape)
- [x] RAW
- [x] SER

### LINKED ISSUE #331: Implement subframe tiling for readers that use Partition3D
- [ ] implement efficient cropped reading (at most read one contiguous block from each frame, then crop the requested area)
  - [ ] use efficient cropped reading for both `crop_to` and "subframe tiling"
- [ ] maybe share more of the low-level code (some readers could use a striding definition and could share most of the low-level reading code, I think?)

### LINKED ISSUE #373: Tiling in readers with potentially large frames
Right now, only K2IS actually returns tiles, while other readers return frame stacks. The "magic 1 MB" is reached with only 4 frames for 256x256 float32 and a single 512x512 float32 frames.

What happens:
Many algorithms, namely dot product, benefit from depths in the region of 32 frames. We lose performance with very shallow frame stacks, in particular when combines with deep mask stacks.

What should happen:
We should support a tiling scheme with 1 MB tiles that contain data from about 32 frames for all formats that can have frame sizes of 256x256 or larger. That should make many operations significantly faster.

See also #27, #331, #247

### LINKED ISSUE #435: Check and optionally update default tileshape behavior for BLO files
How to reproduce:

Open a BLO file that is not in the parameter cache, for example the Glasgow file.

Observed behavior: The default tile shape is 1,8,128,128. However, the signal dimension of this dataset is (144,144), which makes (128,128) seem like an odd choice.

Question: Is this sensible, is the tileshape parameter still necessary at all for BLO, and/or can we guess better defaults? Perhaps something like (1, 6, 144, 144)?

Since BLO files tend to be small, this is not a big issue in practice at the moment.

### LINKED ISSUE #554: Tile size and read mode negotiation for UDFs
In relation to #549 and #551: The optimal tile size for `process_tile()` depends on the circumstances:

* The UDF creates intermediate data similar in size to the tile (log scaling etc), the dataset doesn't use mmap or the data is converted before passing to the UDF: Tiles of approx. 1 MB are optimal.
* The dataset can return mmapped data directly without any conversion and the UDF works efficiently on large tiles, for example with a pure dot product with a shallow mask stack: Very large tiles up to partition size can be optimal.

For that reason, the UDF interface should allow for negotiation of the tile size.

Idea by @sk1p: Instead of directly using `get_tiles()` with loads of parameters, the partition could allow to create a ProtoTileStream object which handles the decision logic on Dataset side (full frame? mmap? dtype?) and can give information on the actual read strategy with the given parameters before tiles are loaded.

The UDF can indicate a preferred upper limit for tiles, which could be 1 MB by default. This is in no way guaranteed since reading full K2 frames will be way above this limit.

The UDFRunner can then decide on a preferred tile size based on the ProtoTileStream, UDF method and UDF preferred size information.

### LINKED ISSUE #634: Allow arbitrary tiling in both signal and nav dimension for MemoryDataSet
Currently, `MemoryDataSet` only allows tiling of the navigation dimension.

In the future, it should allow to specify all kinds of tiling schemes that LiberTEM supports in order to be able to test them with UDFs. Particular example where this is relevant: `StdDevUDF` #625 . 

Furthermore, #554 would require being emulate a DataSet that has limitations in tiling.

The updated `MemoryDataSet` should also be able to test #382, including testing eventual limitations of partition size that a `DataSet` back-end might have vs. the size of items that are passed on for merging on the central node.

### LINKED ISSUE #635: Information on tiling scheme in UDF
Discussion with @sk1p :

The tiled `StdDevUDF` #625 requires a per-tile frame counter. In the current version, this is solved with a somewhat awkward `dtype='object'` buffer and custom dynamic data structure. In the future, information on the tiling scheme should be available as meta information. That allows to allocate the buffers in `get_result_buffers()` or `get_task_data()`.

Ideas by @sk1p :

image

image

sk1p:
Fixes #554, fixes #634, fixes  #27, fixes #331, fixes #373, fixes #635, fixes #435, fixes #247 

## Functionality implementation

- [x] Update `MemoryDataSet` #634 
- [x] Create API for tileshape negotiation #554 
- [x] Add tiling scheme and current index to `UDFMeta` #635 
- [x] Update all `DataSet` impls to work with the new infrastructure
  - [x] hdf5 - can't use the I/O backend as I/O is done by HDF5 itself
  - [x] ser - I/O is done by ncempy, so we can't use the I/O backend
  - [x] raw
  - [x] mib
  - [x] blo
  - [x] k2is
  - [x] frms6
  - [x] empad
  - [x] memory
  - [x] dm
  - [x] cluster ds
- [x] Completely remove `crop_to`
- [x]  update `PickFrameJob` to use ROI
- [x] perf optimization: try not to recompile decoders etc. too much (test case: repeatedly call `get_tiles` with a non-native dtype, it will recompile something every time)
- [x] rudimentary debugging facilities for tiling

## Quality improvements
- [x] Copy test cases etc. from #398 
- [ ] Add test cases for flattening the signal dimension of tiles using `libertem.common.buffers.reshaped_view()` #739
- [x] Possibly rename `Partition3D` and related classes - maybe merge back to `Partition`? Need to look at `DataSet`s that don't use `Partition3D` currently (k2is, hdf5, cached)
- [x] Clean up WIP commits (squash)
- [ ] Look at and/or improve test coverage (really needs #86 to have impact!)
- [ ] Remove `tileshape` and `enable_direct` parameters where they still exist (throw some kind of warning?) → moved to #754 
- [x] Write a quick developer documentation for future-me, that details the interaction between `get_read_ranges`, `decode`, `Decoder`, `IOBackend`, etc. - dev docs for implementing new data set support
- [ ] `File`s should be instantiated via the I/O backend, should really use URLs (`file:///` for local files, ...) → related to #753 
  - [ ] Related: do we really need the `LocalFile3D._mmap_to_array` method? Look at impls and try to move into base class 

## Contributor Checklist:


uellue:
> @uellue the `test_mask_correction` still fails sometimes:

Yes, that's because it uses randomized data and masks and kicks out pixels. It then tries to reconstruct the pixels from their neighbors. If you are unlucky, this changes the result significantly. I'll look into it. Perhaps I can "massage" the data in such a way that the reconstruction is always correct.

sk1p:
Ahh, hm. 

> Perhaps I can "massage" the data in such a way that the reconstruction is always correct

If that is possible, that would be great. Maybe in addition compare the correcting mask with a "first correct, then dot" approach?

uellue:
Just for information, it was a combination of amplified numerical errors and a bug in the mask correction function.

sk1p:
`test_mask_patch` failed again, no need to fix before the release, just want to keep the output somewhere:

<details>
```
___________________________________________________________________________________ test_mask_patch ___________________________________________________________________________________

    @pytest.mark.with_numba
    def test_mask_patch():
        for i in range(10):
            print(f"Loop number {i}")
            num_nav_dims = np.random.choice([1, 2, 3])
            num_sig_dims = np.random.choice([2, 3])
    
            nav_dims = tuple(np.random.randint(low=8, high=16, size=num_nav_dims))
            sig_dims = tuple(np.random.randint(low=8, high=16, size=num_sig_dims))
    
            # The mask-basedcorrection is performed as float64 since it creates
            # numerical instabilities otherwise
            data = _make_data(nav_dims, sig_dims).astype(np.float64)
    
            gain_map = (np.random.random(sig_dims) + 1).astype(np.float64)
            dark_image = np.random.random(sig_dims).astype(np.float64)
    
            exclude = _generate_exclude_pixels(sig_dims=sig_dims, num_excluded=3)
    
            damaged_data = data.copy()
            damaged_data /= gain_map
            damaged_data += dark_image
            damaged_data[(Ellipsis, *exclude)] = 1e24
    
            print("Nav dims: ", nav_dims)
            print("Sig dims:", sig_dims)
            print("Exclude: ", exclude)
    
            masks = (np.random.random((2, ) + sig_dims) - 0.5).astype(np.float64)
            data_flat = data.reshape((np.prod(nav_dims), np.prod(sig_dims)))
            damaged_flat = damaged_data.reshape((np.prod(nav_dims), np.prod(sig_dims)))
    
            correct_dot = np.dot(data_flat, masks.reshape((-1, np.prod(sig_dims))).T)
            corrected_masks = detector.correct_dot_masks(masks, gain_map, exclude)
    
            assert not is_sparse(corrected_masks)
    
            reconstructed_dot =\
                np.dot(damaged_flat, corrected_masks.reshape((-1, np.prod(sig_dims))).T)\
                - np.dot(dark_image.flatten(), corrected_masks.reshape((-1, np.prod(sig_dims))).T)
    
>           _check_result(
                data=correct_dot, corrected=reconstructed_dot,
                atol=1e-8, rtol=1e-5
            )

tests/corrections/test_detector.py:370: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

data = array([[ 6.35857393e-01, -1.50785961e+01],
       [ 6.23128996e-01, -1.62525856e+01],
       [ 6.10398171e-01, -1.7426....40667313e-01, -8.66919544e+01],
       [-1.53397012e-01, -8.78659390e+01],
       [-1.66130820e-01, -8.90399287e+01]])
corrected = array([[ 6.35857294e-01, -1.50785962e+01],
       [ 6.23128986e-01, -1.62525855e+01],
       [ 6.10398216e-01, -1.7426....40667168e-01, -8.66919548e+01],
       [-1.53396663e-01, -8.78659388e+01],
       [-1.66131353e-01, -8.90399286e+01]])
atol = 1e-08, rtol = 1e-05

    def _check_result(data, corrected, atol=1e-8, rtol=1e-5):
        m1 = np.unravel_index(np.argmax(np.abs(data - corrected)), data.shape)
        relative_error = np.zeros_like(data, dtype=np.result_type(data, np.float32))
        np.divide(
            np.abs(data - corrected),
            np.abs(data),
            out=relative_error,
            where=(data != 0)
        )
        m2 = np.unravel_index(
            np.argmax(relative_error), data.shape
        )
        # From np.allclose documentation
        select = np.abs(data - corrected) > (atol + rtol * np.abs(data))
        print("Maximum absolute error: data %s, corrected %s" % (data[m1], corrected[m1]))
        print("Maximum relative error: data %s, corrected %s, relative %s" %
            (data[m2], corrected[m2], relative_error[m2])
        )
        print("Triggering tolerance limits: data %s, corrected %s, relative %s" %
            (data[select],
            corrected[select],
            relative_error[select])
        )
    
>       assert np.allclose(data, corrected, atol=atol, rtol=rtol)
E       assert False
E        +  where False = <function allclose at 0x7f198155f820>(array([[ 6.35857393e-01, -1.50785961e+01],\n       [ 6.23128996e-01, -1.62525856e+01],\n       [ 6.10398171e-01, -1.7426....40667313e-01, -8.66919544e+01],\n       [-1.53397012e-01, -8.78659390e+01],\n       [-1.66130820e-01, -8.90399287e+01]]), array([[ 6.35857294e-01, -1.50785962e+01],\n       [ 6.23128986e-01, -1.62525855e+01],\n       [ 6.10398216e-01, -1.7426....40667168e-01, -8.66919548e+01],\n       [-1.53396663e-01, -8.78659388e+01],\n       [-1.66131353e-01, -8.90399286e+01]]), atol=1e-08, rtol=1e-05)
E        +    where <function allclose at 0x7f198155f820> = np.allclose

tests/corrections/test_detector.py:68: AssertionError
-------------------------------------------------------------------------------- Captured stdout call ---------------------------------------------------------------------------------
Loop number 0
Exclude item:  (6, 2, 9)
Exclude item:  (6, 6, 1)
Exclude item:  (5, 2, 2)
Remaining free pixel map:  [[[False False False ... False False False]
  [False False False ... False False False]
  [False False False ... False False False]
  ...
  [False False False ... False False False]
  [False False False ... False False False]
  [False False False ... False False False]]

 [[False False False ... False False False]
  [False  True  True ...  True  True False]
  [False  True  True ...  True  True False]
  ...
  [False  True  True ...  True  True False]
  [False  True  True ...  True  True False]
  [False False False ... False False False]]

 [[False False False ... False False False]
  [False  True  True ...  True  True False]
  [False  True  True ...  True  True False]
  ...
  [False  True  True ...  True  True False]
  [False  True  True ...  True  True False]
  [False False False ... False False False]]

 ...

 [[False False False ... False False False]
  [False  True  True ... False False False]
  [False  True  True ... False False False]
  ...
  [False  True  True ...  True  True False]
  [False  True  True ...  True  True False]
  [False False False ... False False False]]

 [[False False False ... False False False]
  [False  True  True ...  True  True False]
  [False  True  True ...  True  True False]
  ...
  [False  True  True ...  True  True False]
  [False  True  True ...  True  True False]
  [False False False ... False False False]]

 [[False False False ... False False False]
  [False False False ... False False False]
  [False False False ... False False False]
  ...
  [False False False ... False False False]
  [False False False ... False False False]
  [False False False ... False False False]]]
Nav dims:  (10, 9, 9)
Sig dims: (10, 11, 11)
Exclude:  [[6 6 5]
 [2 6 2]
 [9 1 2]]
Maximum absolute error: data 5.047596902207772, corrected 5.047597431370191
Maximum relative error: data 1.4925991518605457, corrected 1.4925987773808824, relative 2.5089097954528127e-07
Triggering tolerance limits: data [], corrected [], relative []
Loop number 1
Exclude item:  (6, 3, 4)
Exclude item:  (7, 4, 4)
Exclude item:  (5, 11, 4)
Exclude item:  (7, 9, 7)
Remaining free pixel map:  [[[False False False ... False False False]
  [False False False ... False False False]
  [False False False ... False False False]
  ...
  [False False False ... False False False]
  [False False False ... False False False]
  [False False False ... False False False]]

 [[False False False ... False False False]
  [False  True  True ...  True  True False]
  [False  True  True ...  True  True False]
  ...
  [False  True  True ...  True  True False]
  [False  True  True ...  True  True False]
  [False False False ... False False False]]

 [[False False False ... False False False]
  [False  True  True ...  True  True False]
  [False  True  True ...  True  True False]
  ...
  [False  True  True ...  True  True False]
  [False  True  True ...  True  True False]
  [False False False ... False False False]]

 ...

 [[False False False ... False False False]
  [False  True  True ...  True  True False]
  [False  True  True ...  True  True False]
  ...
  [False  True  True ...  True  True False]
  [False  True  True ...  True  True False]
  [False False False ... False False False]]

 [[False False False ... False False False]
  [False  True  True ...  True  True False]
  [False  True  True ...  True  True False]
  ...
  [False  True  True ...  True  True False]
  [False  True  True ...  True  True False]
  [False False False ... False False False]]

 [[False False False ... False False False]
  [False False False ... False False False]
  [False False False ... False False False]
  ...
  [False False False ... False False False]
  [False False False ... False False False]
  [False False False ... False False False]]]
Nav dims:  (15,)
Sig dims: (9, 15, 11)
Exclude:  [[ 6  5  7]
 [ 3 11  9]
 [ 4  4  7]]
Maximum absolute error: data -273.7909675366364, corrected -273.7909669451431
Maximum relative error: data -273.7909675366364, corrected -273.7909669451431, relative 2.160382773854681e-09
Triggering tolerance limits: data [], corrected [], relative []
Loop number 2
Exclude item:  (2, 5)
Exclude item:  (3, 13)
Exclude item:  (4, 11)
Remaining free pixel map:  [[False False False False False False False False False False False False
  False False False]
 [False  True  True  True False False False  True  True  True  True  True
   True  True False]
 [False  True  True  True False False False  True  True  True  True  True
  False False False]
 [False  True  True  True False False False  True  True  True False False
  False False False]
 [False  True  True  True  True  True  True  True  True  True False False
  False False False]
 [False  True  True  True  True  True  True  True  True  True False False
  False  True False]
 [False  True  True  True  True  True  True  True  True  True  True  True
   True  True False]
 [False  True  True  True  True  True  True  True  True  True  True  True
   True  True False]
 [False  True  True  True  True  True  True  True  True  True  True  True
   True  True False]
 [False  True  True  True  True  True  True  True  True  True  True  True
   True  True False]
 [False False False False False False False False False False False False
  False False False]]
Nav dims:  (8, 8)
Sig dims: (11, 15)
Exclude:  [[ 2  3  4]
 [ 5 13 11]]
Maximum absolute error: data 0.07574389320898511, corrected 0.07574446700331317
Maximum relative error: data -0.0006397557975983403, corrected -0.0006401413949226331, relative 0.0006027257990319382
Triggering tolerance limits: data [ 0.0120933  -0.00063976], corrected [ 0.01209388 -0.00064014], relative [4.74472723e-05 6.02725799e-04]
```
</details>

tests/udf/test_stddev.py
```diff
@@ -23,14 +23,14 @@ def test_stddev(lt_ctx, use_roi):
...
-    dataset = MemoryDataSet(data=data, tileshape=(3, 2, 257),
-                            num_partitions=2, sig_dims=2)
     if use_roi:
+        dataset = MemoryDataSet(data=data, tileshape=(3, 2, 516),
+                                num_partitions=2, sig_dims=2)
         roi = np.random.choice([True, False], size=dataset.shape.nav)
         res = run_stddev(lt_ctx, dataset, roi=roi)
     else:
+        dataset = MemoryDataSet(data=data, tileshape=(3, 2, 257),
+                                num_partitions=2, sig_dims=2)
```
sk1p, line 33: @uellue Is this OK? Maybe additionally test w/ contiguous tileshape and no `roi`?

uellue: Looks perfect, if it passes. Maybe with a FIXME to change it back as soon as ROIs are supported? Do we have an issue or item for that already?

docs/source/dev/how-io-works.rst
```diff
@@ -0,0 +1,134 @@
+How does I/O work in LiberTEM?
+==============================
+
```
uellue, line 3: Many algorithms benefit from :ref:`tiled` where the same subset in the signal dimension is processed for several frames in a row. In many cases, algorithms have specific minimum and maximum sizes in signal dimension, navigation dimension or total size where they operate efficiently. Smaller sizes might increase overheads, while larger sizes might reduce cache efficiency.

On the other side, file formats might operate well within specific size and shape limits. The :ref:`k2is` raw format is a prime example where data is saved in tiled form and can be processed efficiently in specific tile sizes and shapes that follow the native layout. Furthermore, some formats require decoding or corrections by the CPU, such as :ref:`frms6`, where tiles that fit the L3 cache can speed up subsequent processing steps. Requirements from the I/O method such as alignment and efficient block sizes are taken into account as well.

The LiberTEM I/O back-end negotiates a tiling scheme between UDF and dataset that fulfills requirements from both UDF and dataset side as far as possible. However, it is not always guaranteed that the supplied data will fall within the requested limits.

docs/source/dev/how-io-works.rst
```diff
@@ -0,0 +1,134 @@
...
+bytes, beginning at offset :code:`start`, from the file :code:`file_idx`
+in the corresponding :code:`FileSet`.
+
+Overriding :code:`DataSet`\ s are free to add additional fields to the end, for
+example if the decoding functions need additional information.
+
+As an example when you would generate custom read ranges, have a look at the
+implementations for MIB, K2IS, and FRMS6 - they may not have a direct 1:1 mapping
+to a numpy :code:`dtype`, or the pixels may need to be re-ordered after decoding.
+In these cases,
```
uellue, line 73: Incomplete sentence: "In these cases, ..."

docs/source/dev/how-io-works.rst
```diff
@@ -0,0 +1,134 @@
...
+implementations for MIB, K2IS, and FRMS6 - they may not have a direct 1:1 mapping
+to a numpy :code:`dtype`, or the pixels may need to be re-ordered after decoding.
+In these cases, 
+
+Notes for implementing a :class:`~libertem.io.dataset.base.DataSet`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- Read file header(s) in :meth:`~libertem.io.dataset.base.DataSet.initialize` -
+  make sure to do the actual I/O in a function dispatched via the
+  :code:`JobExecutor` that is passed to :code:`initialize`.
```
uellue, line 80: See also :ref:`TODO add ref to tips.rst` regarding platform-dependent code.

tests/udf/test_stddev.py
sk1p: Updated `MemoryDataSet` with very simple support for roi w/ arbitrary tileshapes, so we won't need the FIXME at all.

docs/source/dev/how-io-works.rst
sk1p: Good catch! Removing for now, can add more detail later.

uellue:
Congratulations and thanks for this major improvement! 🍾 🏆

Issue #687, opened 2020-03-18
Improved Export/Download support


sk1p:
This is a continuation of #665, related to discussion in tickets #46, #168, #328

Now that the basic functionality is there, we can get first feedback from users; then it can be further improved:

- [ ] Support for CSV/Text
- [ ] Support for hspy
- [ ] Support for more formats?
- [ ] Better filenames, maybe including name from the GUI?
- [ ] Support for writing analysis parameters into the exported file, to support reproducing the results
- [ ] Find out: can other programs properly open multi-page TIFF documents? DM, HyperSpy, anything else?

Known issues:

- [ ] Image formats (currently TIFF) don't support complex data
- [ ] Code quality in the GUI according to code climate can be improved, although most of the issues are not new...

sk1p:
HDF5 export of CoM is broken:

```
HTTPServerRequest(protocol='http', host='localhost:3000', method='GET', uri='/api/compoundAnalyses/88354ca5-bba4-43ca-b026-a90f5aa91136/analyses/ebac42fd-180a-41c1-bdb9-e641215fdf1d/download/HDF5/', version='HTTP/1.1', remote_ip='127.0.0.1')
Traceback (most recent call last):
  File "/home/alex/.virtualenvs/libertem/lib/python3.7/site-packages/tornado/web.py", line 1699, in _execute
    result = await result
  File "/home/alex/source/LiberTEM/src/libertem/web/analysis.py", line 94, in get
    result_formatter.serialize_to_buffer(buf)
  File "/home/alex/source/LiberTEM/src/libertem/io/writers/results/formats.py", line 19, in serialize_to_buffer
    f[k] = self._result_set[k]
  File "/home/alex/.virtualenvs/libertem/lib/python3.7/site-packages/h5py/_hl/group.py", line 387, in __setitem__
    ds = self.create_dataset(None, data=obj, dtype=base.guess_dtype(obj))
  File "/home/alex/.virtualenvs/libertem/lib/python3.7/site-packages/h5py/_hl/group.py", line 136, in create_dataset
    dsid = dataset.make_new_dset(self, shape, dtype, data, **kwds)
  File "/home/alex/.virtualenvs/libertem/lib/python3.7/site-packages/h5py/_hl/dataset.py", line 83, in make_new_dset
    else base.guess_dtype(data)))
  File "/home/alex/.virtualenvs/libertem/lib/python3.7/site-packages/numpy/core/_asarray.py", line 85, in asarray
[...]

PR #723, opened 2020-04-14
Fix CoM result download

sk1p:
- Allow `AnalysisResult` to be tagged with `include_in_download=False`
 - When downloading, filter `AnalysisResultSet` to only include results that should be
   downloaded
 - Fix `AnalysisResult.__array__` to always coerce the raw data to np.ndarray

## Contributor Checklist:


uellue:
Looking good, confirmed to fix the bug on my system.

When looking at the data, I noticed that we need some sort of indication which layers or components are omitted from the result set. Before, there was an 1:1 matching between items in the drop-down and layers in the result. At least in Fiji I can't see any indication which layer is which.

Perhaps a zip file with individual files for each layer as discussed in #665 and a JSON file with metadata would be better? Maybe being able to encode all information to completely restore an `AnalysisResultSet` could be a good guiding principle for designing the download format?

sk1p:
> When looking at the data, I noticed that we need some sort of indication which layers or components are omitted from the result set. Before, there was an 1:1 matching between items in the drop-down and layers in the result. At least in Fiji I can't see any indication which layer is which.

True.

> Perhaps a zip file with individual files for each layer as discussed in #665 and a JSON file with metadata would be better? 

I don't think it would be strictly better - a TIFF file you can directly open in DM/Fiji/GIMP/... - zipped, you first need to fiddle around with file management.  I would agree that it would be nice to have a zip w/ layers + parameters.json download format in addition.

Sadly, TIFF doesn't support layer names, or this would be a no-brainer... HDF5 is much better suited for this kind of application, IMHO:

```
In [2]: f = h5py.File("/home/alex/Downloads/results.h5", "r")                   

In [3]: f.keys()                                                                
Out[3]: <KeysViewHDF5 ['divergence', 'magnitude', 'x', 'y']>
```

> Maybe being able to encode all information to completely restore an `AnalysisResultSet` could be a good guiding principle for designing the download format?

Yeah, see #687 - we should write analysis and dataset parameters into formats that support those.

sk1p:
> Perhaps indicating in the GUI layer drop-down or in the result download dialogue which layers are included in download?

Sounds good. Could be done by serializing the "included with download" flag together with the result set, and querying that from the download modal.

sk1p:
image

sk1p:
This is now implemented, let me know if this works for you. Should be merge-able after rebuilding the client, feel free to take over the PR tomorrow.

uellue:
Fix confirmed, many thanks! :-)

PR #729, opened 2020-04-16
Reduce UDF result overhead

### LINKED ISSUE #725: Size of task results exchanged with Dask cluster suspected to be 16x inflated
LiberTEM is unusably slow when using a remote Dask cluster over a typical broadband internet connection of about 1 MB/s, with task results coming in only every few seconds. The system monitor shows maxed-out network traffic while an analysis is running and drops to almost zero background load when it finishes.

This is in contradiction with the expected size of task results: On 256x256 frames, a task result from `SumUDF` should have a size of  256 kB. That means it should manage about four updates per second. Instead, it manages only one update every four seconds.

Setup: LiberTEM master, Python 3.6, LiberTEM server on Windows, Dask cluster on Linux (moellenstedt).

Unfortunately, I didn't manage to capture the network traffic because sniffing localhost on Windows  is difficult.

@sk1p can you reproduce the issue with LiberTEM server on Linux?

sk1p:
Fixes #725 

## Contributor Checklist:

### DIFF
--- a/docs/source/changelog/bugfix/reduce_udf_overhead.rst
+++ b/docs/source/changelog/bugfix/reduce_udf_overhead.rst
@@ -1,0 +1,4 @@
+[Bugfix] reduce overhead of UDF results
+=======================================
+
+* No longer send the Partition instances together with the result (:pr:`729`, :issue:`725`)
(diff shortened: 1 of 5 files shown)


uellue:
Fix confirmed, awesome!

PR #731, opened 2020-04-16
Handling disable_quickedit() failure in Bash

sayandip18:
## Contributor Checklist:

changes were made to the GUI)


docs/source/usage.rst
```diff
@@ -172,6 +172,54 @@ download the available results.
...
 ..  figure:: ./images/use/download-modal.png
 
+Possible error
+~~~~~~~~~~~~~~~~
+
+You may get the following error if you try to run libertem-server for a second time
+
+.. code-block:: shell
+
+    OSError: [WinError 10048] Only one usage of each socket address (protocol/network address/port) is normally permitted
```
sk1p, line 182: What is the root cause of this error? Is it reproducible? If so, how? Instead of documenting it here, we should strive to fix the cause instead.

src/libertem/win_tweaks.py
```diff
@@ -2,10 +2,21 @@
...
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.ERROR)
+
+formatter = logging.Formatter('%(levelname)s : %(message)s : %(asctime)s')
+
+file_handler = logging.FileHandler('win_tweaks.log')
+file_handler.setFormatter(formatter)
+
+logger.addHandler(file_handler)
+
```
sk1p, line 19: Is there a reason you are adding a file logger here? IMHO logging configuration should be done in one central place.

src/libertem/win_tweaks.py
```diff
@@ -47,10 +58,13 @@ def set_console_mode(mode, stream=sys.stdin):
...
-    mode &= ~ENABLE_QUICK_EDIT
-    mode |= ENABLE_EXTENDED_FLAGS
-    set_console_mode(mode)
+    try:
+        mode = get_console_mode()
+        mode &= ~ENABLE_QUICK_EDIT
+        mode |= ENABLE_EXTENDED_FLAGS
+        set_console_mode(mode)
+    except TypeError:
+        logger.error("Incompatible types!")
```
sk1p, line 68: Instead of reacting to `TypeError`, what about catching the original `Exception` here, that is, the `ctypes.WinError`? In the `get_console_mode` function, if you are catching the exception, it is not clear what to do in case of an error, because people could be calling the function for whatever reason. That's why it is throwing an exception in the first place! In `disable_quickedit`, you know what to do: ignore the error, because the console likely doesn't have quickedit functionality, and log a message accordingly.

A note on the log messages: "Incompatible types" and "Failed!!!" are not very helpful: imagine a user reading these, not knowing where they come from. They would be alarmed! But in reality, most likely, it is just a side-effect of them using git bash. The message should be clear and describe what is happening.

The log severity should reflect this, too: not disabling quick edit on git bash should not be a cause for concern, it should be logged as `DEBUG` (or `INFO` at most) message, in case this ever fails outside of git bash for some reason.

docs/source/usage.rst
sayandip18: I think it is because of some problem in starting the server. I came across [this](https://stackoverflow.com/questions/12362542/python-server-only-one-usage-of-each-socket-address-is-normally-permitted) and it appears that we need to shut down the current process in order to start the next process. You can reproduce this by making the same changes in `win_tweaks.py` file as I have made.

src/libertem/win_tweaks.py
sayandip18: I thought that if someone faces some problem in the future it would make their job easier to pinpoint the issue.

src/libertem/win_tweaks.py
sayandip18: Okay, I will try to work on the `ctypes.WinError` and make the other changes as well.

sayandip18:
How do I take care of codecov/patch?

sk1p:
> How do I take care of codecov/patch?

Usually, this is fixed by adding test cases for your changes. In this case, as we can't test this automatically in a realistic scenario (the CI pipeline doesn't run in a git bash window!), I would disregard the coverage check. It is more important the rest of the PR is in good shape.

docs/source/usage.rst
sk1p: > down the current process in order to start the next process

Yes, but what was the reason the current process was not shut down? This should not be caused by the changes you did in this PR.

src/libertem/win_tweaks.py
```diff
@@ -27,11 +38,11 @@ def get_console_mode(stream=sys.stdin):
...
     mode = wintypes.DWORD(0)
-    if getConsoleMode(file_handle, byref(mode)):
-        return mode.value
-    else:
-        err = ctypes.get_last_error()
-        raise ctypes.WinError(err)
+    try:
+        if getConsoleMode(file_handle, byref(mode)):
+            return mode.value
+    except ctypes.WinError(ctypes.get_last_error()):
```
sk1p, line 44: I don't think this is right - why are you constructing an instance of the `WinError` class here in the `except` clause?

src/libertem/win_tweaks.py
```diff
@@ -2,10 +2,21 @@
...
 
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.ERROR)
+
+formatter = logging.Formatter('%(levelname)s : %(message)s : %(asctime)s')
+
+file_handler = logging.FileHandler('log_file.log')
+file_handler.setFormatter(formatter)
+
+logger.addHandler(file_handler)
```
sk1p, line 18: Still not happy about having logging setup code in here. As it is, this will create a `log_file.log` in the directory where LiberTEM is started, but only on Windows systems. Currently, we don't do logging to files anywhere, we only log to the console where the server is started, so that is where I (and other users) would expect such messages to end up.

If you want to add file logging to LiberTEM, maybe a good idea would be to do this in a separate PR, maybe together with #743?

src/libertem/win_tweaks.py
```diff
@@ -47,10 +58,13 @@ def set_console_mode(mode, stream=sys.stdin):
...
-    mode &= ~ENABLE_QUICK_EDIT
-    mode |= ENABLE_EXTENDED_FLAGS
-    set_console_mode(mode)
+    try:
+        mode = get_console_mode()
+        mode &= ~ENABLE_QUICK_EDIT
+        mode |= ENABLE_EXTENDED_FLAGS
+        set_console_mode(mode)
+    except TypeError:
+        logger.info("In win_tweaks.py, get_console_mode() returns None.")
```
sk1p, line 67: I'm still not 100% happy with this. Catching the `WinError` in `get_console_mode` causes a `TypeError` in `disable_quickedit`, which then again has to be caught... why catch the `WinError` at all in `get_console_mode`, if you effectively convert it into a `TypeError` down here, which you then again have to catch?

About the logged message: this is currently a "statement": it just states something about code internals. Would a user know what to do with that? How about this instead:

"Quick Edit mode could not be disabled, you are probably using a console which doesn't support Quick Edit (like git bash)"

That way, if the user **knows** they are using a console which supports Quick Edit (like CMD), and get this message, they **know** something is wrong. If they are using git bash, they are not alarmed because the message tells them that this is to be expected.

docs/source/usage.rst
sayandip18: This happens only when libertem-server is run from Bash. Everything's alright with Anaconda Prompt.

src/libertem/win_tweaks.py
sayandip18: Are these things (like quick edit) a part of /related to win32 API? If not, where can I learn more about them? The Win 32 API documentation isn't much helpful.

src/libertem/win_tweaks.py
sayandip18: Don't we have to supply `get_last_error()`?

docs/source/usage.rst
sk1p: When specifically does this happen? Does this also happen without your changes? If so, please log a separate issue for this, with all details on how to reproduce this, so we can fix this. The fix should not be part of this PR: each PR should only be a single logical change, if possible at all.

src/libertem/win_tweaks.py
sk1p: When **raising** a `WinError`, you have to provide the error code, yes. But if you are **handling** exceptions, this is different: only the **types** are matched in the `except` clause. The code that wants to **handle** the error can't know the details, like error codes or messages - they are only known *where the error occurs*!

Please, do read [the "Errors" chapter of the Python tutorial](https://docs.python.org/3/tutorial/errors.html), which explains the basics of exception handling, and let me know if this is now clear.

docs/source/usage.rst
sayandip18: > 
> 
> When specifically does this happen? 

At first the libertem-server has to be run from Bash. Then, both the Bash console and webpage have to be closed. Now, if you try to run the server again (either from Bash or Anaconda prompt), you will get the error. If we start the server from Anaconda at first, there will not be any error.

> 
> 
> Does this also happen without your changes? 

 No, because without my changes the server won't start at all from Bash.

>
> If so, please log a separate issue for this, with all details on how to reproduce this, so we can fix this. The fix should not be part of this PR: each PR should only be a single logical change, if possible at all.

sk1p: > Then, both the Bash console and webpage have to be closed

So you directly close the bash console using the "X" GUI element, without stopping the libertem-server before? If so, does this also happen if you close libertem-server using Ctrl+C before closing the bash console?

> No, because without my changes the server won't start at all from Bash.

￼Right, thanks. Please do log this *as a separate issue*, which can then be fixed after this PR is merged.

sayandip18: > 
> So you directly close the bash console using the "X" GUI element, without stopping the libertem-server before? 

Yes, happens only when the 'X' GUI element is used. 


> 
> If so, does this also happen if you close libertem-server using Ctrl+C before closing the bash console?
 
No, doesn't happen with Ctrl+C

sk1p: :+1: that sounds like something that should be fixed.

src/libertem/win_tweaks.py
sk1p: `GetConsoleMode` and `SetConsoleMode` are part of the Win32 API, yes. [This](https://docs.microsoft.com/en-us/windows/console/getconsolemode) is the API documentation for `GetConsoleMode`, which contains a short description of quick edit. From a user's perspective, [this is how you enable/disable it](https://tweaks.com/windows/36732/command-prompt-quick-edit-mode/) "manually".

src/libertem/win_tweaks.py
```diff
@@ -47,10 +50,13 @@ def set_console_mode(mode, stream=sys.stdin):
...
-    mode = get_console_mode()
-    mode &= ~ENABLE_QUICK_EDIT
-    mode |= ENABLE_EXTENDED_FLAGS
-    set_console_mode(mode)
+    try:
+        mode = get_console_mode()
+        mode &= ~ENABLE_QUICK_EDIT
+        mode |= ENABLE_EXTENDED_FLAGS
+        set_console_mode(mode)
+    except TypeError:
```
sk1p, line 58: Is `TypeError` here the right exception type to catch? Not `ctypes.WinError`? Where does the `TypeError` come from?

sayandip18: TypeError was there in the original file. Also, the issue here was incompatibility of types.

sk1p: Sorry, I don't understand...

sayandip18: `mode` is supposed to be `int`. However, if we get an error, `get_console_mode()` returns `NoneType` and we get a `TypeError` like this:

`line 51, in disable_quickedit mode &= ~ENABLE_QUICK_EDIT `               
`TypeError: unsupported operand type(s) for &=: 'NoneType' and 'int'`

sk1p: > However, if we get an error, `get_console_mode()` returns `NoneType`

Hmm, but if there is an error in `get_console_mode`, it throws an exception, which means it doesn't return anything...

sayandip18: If we use `except` keyword, we cannot put `ctypes.WinError` as it doesn't inherit from `BaseException` class. However, if we use `raise ctypes.WinError()` we get 
` File "c:\users\saswati\venv\libertem\src\libertem\win_tweaks.py", line 45, in get_console_mode`
    `raise ctypes.WinError(err)`
`OSError: [WinError 0] The handle is invalid.`

Thus, back to square one.

sk1p: > back to square one

Have a close look at the traceback again: apparently, the exception type is `OSError`, not `WinError`. Try to catch `OSError` and see if that fixes things.

sayandip18: Right, `OSError` works just fine.

sk1p:
This now looks good, will merge shortly!

PR #732, opened 2020-04-17
Fix additional info for cached file params

### LINKED ISSUE #730: HDF5: drop down not shown when opening from history
When opening a HDF5 file that is in the file history (from before the drop down was implemented), instead of the drop down, only a simple text input is shown.

Instead, even for these files, the drop down should work as expected.

I think the root cause is actually the mixing of info and params, as the params are taken from the history (from Local Storage), and the HDF5 dataset expects the params to also contain the info, which is not the case. 

Maybe a short-term fix would be to change the merging of Local Storage and detection result to always take "info" from the detection. Long-term this should be fixed by properly separating info and params in the GUI

/cc @AnandBaburajan

AnandBaburajan:
Fixes #730 

## Contributor Checklist:

changes were made to the GUI)


AnandBaburajan:
@sk1p Is the solution correct? I'm trying to work on the type safety of the function.

sk1p:
> Is the solution correct? 

Overall it looks correct on a first glance - can't do a detailed test right now sadly. 

> I'm trying to work on the type safety of the function.

:+1: - I imagine they are not so trivial - try to use the `LocalConfig` type as a base. I guess it will have to be ORd together with a type for the info parameters?

client/src/dataset/components/DatasetOpen.tsx
```diff
@@ -50,6 +51,14 @@ const getDefaultDSType = (didReset: boolean, openState: OpenDatasetState) => {
  * @param openState complete OpenDatasetState instance
  */
 
+ // Fix this after separating info from params
+ const mergeCachedParamsInfo = (formCachedParams: any, formDetectedParams: any) => {
+     Object.keys(formDetectedParams).filter((key: string) => isAdditionalInfo(key)).forEach((key: string) => {
+       formCachedParams[key] = formDetectedParams[key];
+     });
```
sk1p, line 58: `.filter((key: string) => isAdditionalInfo(key))` can be replaced with just `.filter(isAdditionalInfo)`, as `isAdditionalInfo` already matches the signature of functions that are expected by `Array.filter`.

Instead of modifying `formCachedParams`, which can be confusing to the calling function, it would be preferable to return a new `Array`. Maybe use `Array.map` here instead?

sk1p:
Apparently there were changes to the client after you created the PR - this means you need to rebase to the current master, too. This is a bit annoying, sorry about that.

AnandBaburajan:
Is the Travis CI build okay?

sk1p:
> Is the Travis CI build okay?

Travis is currently [down for maintenance](https://www.traviscistatus.com/), we'll have to wait for them to come back online

sk1p:
Travis is back, I'm closing and re-opening the PR to trigger a rebuild

sk1p:
For reference, travis build is at https://travis-ci.org/github/LiberTEM/LiberTEM/builds/677137736

sk1p:
Merged, thank you!

PR #738, opened 2020-04-21
Fix StdDevUDF issue with full frames vs tiled

uellue:
The results should be equal within numerical stability limits.

They are not, unfortunately.

Edit: Fixed, refs #739 

## Contributor Checklist:


sk1p:
Test fail because of #737, I'll merge that now, you can then rebase to fix CI for this PR

src/libertem/common/buffers.py
```diff
@@ -267,6 +284,50 @@ def get_view_for_tile(self, partition, tile):
...
+            key = tuple(map(slice_to_tuple, sl))
+            if key in self._contiguous_cache:
+                return self._contiguous_cache[key]
+            view = self._data[sl]
+            oldshape = view.shape
+            try:
+                # See if the signal dimension can be flattened
+                # https://docs.scipy.org/doc/numpy/reference/generated/numpy.reshape.html
+                view.shape = (-1, )
+                view.shape = oldshape
```
sk1p, line 316: Could also check `view.flags.C_CONTIGUOUS` instead - that way we don't have to pay the exception handling cost for every tile

uellue: Ok, thx! :-)

PR #749, opened 2020-04-23
Add tuples to Shape objects

saisunku:
## Contributor Checklist:

* [X] I have added or updated my entry in [the creators.json file](https://github.com/LiberTEM/LiberTEM/blob/master/packaging/creators.json)
* [X] I have added [a changelog entry](https://github.com/LiberTEM/LiberTEM/tree/master/docs/source/changelog) for my contribution
* [X] I have added/updated documentation for all user-facing changes
* [X] I have added/updated test cases
changes were made to the GUI)


docs/source/changelog/bugfix/add_shape_to_tuple.rst
```diff
@@ -0,0 +1,7 @@
+[Bugfix] Addition between Shape and tuple
+===========================================================
+
+* Tuples can be added directly to Shape objects when either
+ the signal or navigation dimensions of the Shape object are zero
```
sk1p, line 5: FYI I think the CI is complaining about this line: the second line of the first bullet point needs to be indentation-aligned to the beginning of the first line ("the" directly below "Tuples")

src/libertem/common/shape.py
```diff
@@ -151,6 +151,42 @@ def __eq__(self, other):
...
+                raise ValueError("To add a Shape object to a tuple, either the navigation "
+                                 "or the signal dimensions of the Shape object must be zero")
+            elif self._sig_dims == 0 and len(self._nav_shape) == 0:
+                raise ValueError("Shape object is empty")
+            elif self._sig_dims == 0:
+                return Shape(self._nav_shape + other, sig_dims=0)
+            else:
+                return Shape(self._sig_shape + other, sig_dims=len(self._sig_shape + other))
+        else:
+            return NotImplemented
```
sk1p, line 171: Is it possible to somehow not have duplication in `__add__` and `__radd__`? Maybe by calling a helper method to do most of the actual work and validation, and just call it with different parameters in add and radd?

tests/common/test_shape.py
```diff
@@ -69,3 +69,27 @@ def test_shape_eq_4():
     s1 = Shape((17, 16, 128, 128), sig_dims=2)
     s2 = Shape((16, 16, 128, 128), sig_dims=2)
     assert s1 != s2
+
```
sk1p, line 72: These tests should probably also check the `sig_dims` of the result

src/libertem/common/shape.py
```diff
@@ -151,6 +151,42 @@ def __eq__(self, other):
...
+        if isinstance(other, tuple):
+            if self._sig_dims != 0 and len(self._nav_shape) != 0:
+                raise ValueError("To add a Shape object to a tuple, either the navigation "
+                                 "or the signal dimensions of the Shape object must be zero")
+            elif self._sig_dims == 0 and len(self._nav_shape) == 0:
+                raise ValueError("Shape object is empty")
+            elif self._sig_dims == 0:
+                return Shape(other + self._nav_shape, sig_dims=0)
+            else:
+                return Shape(other + self._sig_shape, sig_dims=len(self._sig_shape + other))
```
sk1p, line 186: Hmm. I'm not sure the `sig_dims` behavior should be like this - but let me think a bit more about it. Does this match the usage from your grep survey?

docs/source/changelog/bugfix/add_shape_to_tuple.rst
saisunku: Oh I ignored Travis because I thought it was still broken. I fixed it.

src/libertem/common/shape.py
saisunku: Thanks, I noticed the repetition but wasn't sure how to fix it. I made a helper function called `_add_tuple_helper` - is that a good name?

src/libertem/common/shape.py
saisunku: Actually, it seems that this is a bit arbitrary. In some cases, `sig_dims` of the final `Shape` is set to the `sig_dims` of the starting `Shape` object but there is at least one instance (src\libertem\io\dataset\base\partition.py line 84) where the number 2 was used. Since `__add__` cannot accept additional parameters, I'm not sure what what's a good solution.

src/libertem/common/shape.py
sk1p: Sure! Naming things is most of the times subject to [the bikeshedding problem](https://en.wiktionary.org/wiki/bikeshedding), so for internal names, as long as a name is not misleading, I tend not to care as much, as it can also be changed later. For public APIs, where the name is fixed for a long time, it is more important to get it right, of course :grin:

src/libertem/common/shape.py
sk1p: > In some cases, sig_dims of the final Shape is set to the sig_dims of the starting Shape object

This may actually be a good default behavior, at least for concatenating something to the left side of a `Shape` (which could be though of as: concatenating to the navigation dimensions). When concatenating a tuple to the right side (which could be thought of as: concatenating to the signal dimensions), changing the `sig_dims` to include the "other" length may be a good idea.

> where the number 2 was used

I think that is actually a bug in `get_macrotile`! The `sig_dims` of the original `Shape` should be used there. /cc @uellue 

> Since __add__ cannot accept additional parameters, I'm not sure what what's a good solution.

Hmm. The solution could also be to not use `+` at all, but instead provide specialized methods for this. If `sig_dims` is completely replaced with a new value, it is okay to keep things as-is, that is, to construct a new `Shape` from scratch:

```python
t_sig = (16, 16)
Shape(tuple(some_shape) + t_sig, sig_dims=2)
```

Or, one could think about an API like this (assuming `shape.sig_dims == 0`, or that we don't care about the previous `sig` part):

```python
t_sig = (16, 16)
s2 = shape.replace_sig(t_sig)  # assuming `shape.sig_dims == 0`
```

But, it is still possible to use `+` here, by throwing away the old `sig` part:

```python
t_sig = (16, 16)
s2 = shape.nav + t_sig  # assuming we set sig_dims to len(t_sig) here
assert s2.sig.dims == 2
```

This pattern tends to come up sometimes:

```python
t_nav = (16, 16)
s2 = Shape(t_nav + tuple(shape.sig), sig_dims=shape.sig.dims)
```

With addition, this would be:

```python
t_nav = (16, 16)
s2 = t_nav + shape.sig  # assuming we keep the original sig_dims of `shape` here
assert s2.sig.dims == shape.sig.dims
```

Or, with a specialized API:

```python
t_nav = (16, 16)
s2 = shape.replace_nav(t_nav)
```

Now, looking at the alternatives, I like the addition variant more than the `replace_*` method. What do you think?

saisunku: @sk1p Addition while keeping the `sig_dims` of the starting `Shape` object also looks best to me. I feel that creating a new `Shape` object is not so hard and is much more flexible than specialized methods like `replace_*`. So I don't think `replace_*` will be used much.

> This may actually be a good default behavior, at least for concatenating something to the left side of a Shape (which could be though of as: concatenating to the navigation dimensions). When concatenating a tuple to the right side (which could be thought of as: concatenating to the signal dimensions), changing the sig_dims to include the "other" length may be a good idea.

This sounds good but it's different from what we started with. In this case, it actually doesn't matter what `sig_dims` and `nav_dims` of the original `Shape` object are. One of them doesn't have to be zero. Whether we add to the signal or navigation dimensions depends only on whether the addition is from the left or from the right. Is that OK?

src/libertem/common/shape.py
saisunku: This is good to know, thanks!

src/libertem/common/shape.py
sk1p: > I feel that creating a new Shape object is not so hard and is much more flexible than specialized methods like replace_*. So I don't think replace_* will be used much.

:+1:

> Whether we add to the signal or navigation dimensions depends only on whether the addition is from the left or from the right. Is that OK?

Yeah, let's try that!

saisunku: @sk1p please take a look now

PR #750, opened 2020-04-24
HDF5 sample dataset

AnandBaburajan:
## Contributor Checklist:

### DIFF
--- a/docs/source/sample_datasets.rst
+++ b/docs/source/sample_datasets.rst
@@ -30,50 +30,48 @@ Creating random data
 
 Random data can be generated in the following way. It should be kept in mind
 that the data generated in this way can only be used for simple testing as it
-has no physical significance. You can then load the data through the
-:ref:`api documentation`.
+has no physical significance.
 
 **Raw file:**
 
-.. code-block:: python
+.. testcode:: createraw
 
     # Create sample raw file
     import numpy as np
-    real_data = np.random.randn(16, 16, 16, 16).astype("float32")
-    real_data.tofile("/tmp/real_raw_file.raw")
+    sample_data = np.random.randn(16, 16, 16, 16).astype("float32")
+    sample_data.tofile("raw_sample.raw")
 
-.. code-block:: python
+.. testcode:: loadraw
 
-    # Load the file
+    # Load through Python API
     from libertem.api import Context
     if __name__ == '__main__':
       ctx = Context()
-      ds = ctx.load("raw", path="/tmp/real_raw_file.raw", scan_size=(16, 16), dtype="float32", detector_size=(16, 16))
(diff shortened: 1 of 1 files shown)


docs/source/sample_datasets.rst
```diff
@@ -30,22 +30,48 @@ Creating random data
 
 Random data can be generated in the following way. It should be kept in mind
 that the data generated in this way can only be used for simple testing as it
-has no physical significance.
+has no physical significance. You can then load the data through the
+:ref:`api documentation`.
```
sk1p, line 34: This should link directly to the ["Loading data"](https://libertem.github.io/LiberTEM/formats.html) section, I think.

docs/source/sample_datasets.rst
```diff
@@ -30,22 +30,48 @@ Creating random data
...
 .. code-block:: python
-    
+
+    # Create sample HDF5 file
+    import h5py
+    import numpy as np
+    file = h5py.File('hdf5_sample.h5','w')
+    dataset = file.create_dataset("dataset",(16,16,16,16), h5py.h5t.STD_I32BE)
+    dataset = file['/dataset']
+    data = np.random.randn(16,16,16,16).astype("float32")
```
sk1p, line 63: Hmm, you are creating float data, but setting the `dtype` of the HDF5 dataset to `h5py.h5t.STD_I32BE` - that is, 32bit big-endian integers. I don't think this will work as intended. Instead of specifying the `dtype` explicitly, you can also provide the `data` argument instead, [see the reference docs](http://docs.h5py.org/en/stable/high/group.html#Group.create_dataset) and [the creating datasets](http://docs.h5py.org/en/stable/high/dataset.html#dataset-create) section for details.

docs/source/sample_datasets.rst
```diff
@@ -30,22 +30,48 @@ Creating random data
...
+.. code-block:: python
+
+    # Load the file
     from libertem.api import Context
-    ctx = Context()
-    ds = ctx.load("raw", path="/tmp/real_raw_file.raw", scan_size=(16, 16), dtype="float32", detector_size=(16, 16))
-    
+    if __name__ == '__main__':
+      ctx = Context()
+      ds = ctx.load("hdf5", path="./hdf5_sample.h5", ds_path="/dataset", tileshape=(1, 4, 16, 16))
```
sk1p, line 73: The `tileshape` parameter can be removed, as it is ignored in the current master branch.

docs/source/sample_datasets.rst
AnandBaburajan: Sorry, thanks!

sk1p:
Thanks, merged!

PR #751, opened 2020-04-24
Detect discrepancies in `Buffer` cache use

uellue:
* Use `Slice` objects as keys, which are hashable
* Assert that they don't overlap
* Make sure we only flush correctly and trigger an error otherwise

We only have to check once when a new key is created for the cache,
which means it shouldn't noticeably impact performance. Plus, we can
deactivate asserts for production.

## Contributor Checklist:


sk1p:
Thanks, merged!

Issue #198, opened 2018-12-11
Well-organized benchmarking suite


uellue:
Right now, we do have automated tests, but we don't have automated benchmarks. 

We should have a benchmarking suite that can be executed to gauge the impact of a change. The unit tests are not really suitable because they often use LiberTEM in a non-standard way: inline executor instead of dask, very small in-memory data sets, test of "weird" partitioning schemes, using sparse matrices where one really shouldn't #182 .

Instead, we should have a reliable test program with the following properties:

* Test on large real-world data sets with typical analyses
* Executed under reproducible conditions: Dedicated machine without background tasks
* Executed on a small cluster as well as a typical single-node workstation.
* Collect profiling information
* Regular automated collection and reporting of benchmark results to catch changes with a negative impact on performance.

sk1p:
See also: https://github.com/airspeed-velocity/asv

uellue:
The benchmarking suite should separate processing and overheads. Overheads can be a major factor #232

uellue:
We should routinely compare math libraries:

* Torch vs openblas (vs mkl?)
* pyfftw vs numpy.fft vs scipy.fft (vs mkl?)

As a side note, it seems that openblas might have caught up to torch recently -- further tests desirable.

sk1p:
Btw. for FFT comparison, see also https://www.kcod.de/gearshifft/

sk1p:
Idea: use an Azure on-premise Agent to run the benchmarks as part of the CI pipeline, on a dedicated server. Start with tests that can run in a few seconds, maybe later introduce daily benchmark run for longer running tests (for example ones that create some TB of data on demand, also to test our writing facilities - things you don't want to run on every commit)

uellue:
Asvlooks like the best bet so far and should cover all points when used appropriately.

A few more detailed thoughts on use cases and intermediate results after playing with asv:

* Routinely check during development if a change introduced regressions to catch things like #814 before merging
  - That means asv has to run code from current fork, current branch.
  - Using the "same repo" layout collides with the existing structures of LiberTEM.
  - Using the "separate repo" layout in a separate repo LiberTEM/benchmarks required setting a Git URL or path to the actual repo to benchmark, meaning one would have to merge into main repo master first or adapt the path to the way how the repos are checked out, which is fragile.
  - Using the "separate repo" layout in a subfolder of the main repo, and setting the repo URL/path to a relative path pointing to the repository root ('..') worked.
[...]

PR #760, opened 2020-04-27
GPU support through CuPy

### LINKED ISSUE #364: Support GPU processing
Right now, LiberTEM is limited to using the CPU.

For some operations, namely large-scale matrix products with sparse matrices, Fourier transforms and correlations, may benefit significantly from using GPUs if they are available.

We should look into ways to enable that in LiberTEM. Options include:

* cupy https://docs-cupy.chainer.org/en/stable/index.html
* pyopencl https://documen.tician.de/pyopencl/
* ArrayFire http://arrayfire.org/arrayfire-python/
* Numba https://numba.pydata.org/numba-doc/dev/cuda/index.html
* https://numba.pydata.org/numba-doc/dev/roc/index.html

Do you have experience with any of these and are there other options as well?

uellue:
Closes #364 

## General approach

* Use dask.distributed worker and task resources to make sure we don't oversubscribe and assign workers to CPU or GPU duty
* Use `SpecCluster` to start a heterogeneous cluster
* UDF signals which back-ends it can use, forwarded in UDFTask
* Executor translates this to resource requests as part of Dask futures. CPU-only, CUDA-only, and universal compute for universal UDFs that can use both
* Side effect: We can start extra workers for small tasks, addresses https://github.com/LiberTEM/LiberTEM/issues/186#issuecomment-534518019
* Dask.distributed scheduler does its thing and assigns tasks to free workers that have required resources
* UDFRunner decides based on environment variables on the workers which back-end to actually use

## Done

- [x] Dynamically decide on workers which back-end to use
- [x] Create local heterogeneous cluster with GPU devices
- [x] Assign tasks to the correct worker pools
- [x] Option to allocate result buffers on device
- [x] Test cases using Inline executor pass
- [x] Test on real CUDA host with actual cupy, not only mock

## TODO

- [x] Review and refine architecture (@sk1p can you have a look?)
- [ ] Device array support in datasets, can be naive for the beginning
- [x] Test cases for starting a hybrid cluster
- [x] Test case for GPU-only UDF
- [ ] Run CI on host with CUDA
- [x] Throw exception rather than stalling if no worker has the requested resources
- [x] Make sure all methods to start a cluster assign resources to workers
   - [x] Unit tests
   - [x] Web interface
   - [x] Instructions for manual cluster
- [x] Roll out in UDFs
   - [x] Holography
   - [x] ApplyMasksUDF
   - [x] Single Side Band (ptychography)
   - [ ] Blobfinder?
   - [ ] ...?
- [ ] Combine with tile shape negotiation
- [x] Concept to connect plain C++ CUDA code without CuPy
   * https://docs-cupy.chainer.org/en/stable/reference/interoperability.html
   * https://numba.pydata.org/numba-doc/latest/cuda/cuda_array_interface.html
   * CC @SimeonEhrig for comment :-)
- [x] Documentation and proper change log entry

## Contributor Checklist:

* [x] TODO I have added/updated documentation for all user-facing changes
* [x] TODO add more I have added/updated test cases


src/libertem/udf/base.py
```diff
@@ -684,6 +757,10 @@ def __init__(self, partition, idx):
     def get_locations(self):
         return self.partition.get_locations()
 
+    def get_resources(self):
```
sk1p, line 760: Short docstring would be nice here. The 1 means we need one CPU? What happens if I request 2 CPUs here? Will it raise an exception because it can't service that request?

src/libertem/executor/dask.py
```diff
@@ -151,7 +160,8 @@ def _make_items_all():
         if all_nodes:
             items = _make_items_all()
         else:
-            items = ((lambda: fn(p), p.get_locations())
+            # TODO check if we should request a compute resource
```
sk1p, line 163: ... or possibly a "service" resource? Depends on the `fn`, I guess - current usage is only "service".

src/libertem/api.py
```diff
@@ -691,9 +705,51 @@ def _create_local_executor(self):
...
-            cluster_kwargs={"threads_per_worker": 1, "n_workers": cores},
-            client_kwargs={'set_as_default': False}
+        workers_spec = {}
+
+        cpu_base_spec = {
+            "cls": dd.Nanny,
+            "options": {"nthreads": 1, "resources": {"CPU": 1, "compute": 1}}
+        }
+
+        # Service workers not for computation
```
sk1p, line 715: Hmm, maybe we can make the service workers multi-threaded? They will mostly block on I/O anyways, so we probably won't have GIL problems. Just need to take care w/ thread safety.

Should the service worker define a `{"service": 1}` resource, which then can be acquired by `executor.run_function(service=True)` and similar?

uellue:
@sk1p Thx for the review, helpful comments! I'll let my head process it in the background a bit and then work on improving.

sk1p:
Btw. for testing I added a `backends` parameter to `run_udf` - this was helpful for comparing CPU+GPU, CPU only, GPU only variants. Maybe something like this is generally useful: https://github.com/uellue/LiberTEM/compare/workworkwork...sk1p:run-udf-backends?expand=1

uellue:
> Btw. for testing I added a `backends` parameter to `run_udf` - this was helpful for comparing CPU+GPU, CPU only, GPU only variants. Maybe something like this is generally useful: https://github.com/uellue/LiberTEM/compare/workworkwork...sk1p:run-udf-backends?expand=1

That looks practical, thx! :-)

SimeonEhrig:
@uellue Looks good. I would like to test it in the next days with the binding prototype (https://github.com/Ptychography-4-0/ptychography/pull/23). Unfortunately I am very busy at the moment, just like you. So I can't say when I have time for it, but I'm trying to use every gap I find in my schedule ;-) 

For GPU-CI: I do not know any CI service which provides GPUs. In our case, we use self-hosted GitLab runners. For GitHub projects is this a problem. We solved it with an automated mirror on gitlab.com, where we can use our self-hosted GitLab runner. Since we set up this solution, GitHub has developed its own CI tool called GitHub Actions. GitHub Actions also allows us to use our self-hosted runner: https://github.blog/2019-11-05-self-hosted-runners-for-github-actions-is-now-in-beta/
Maybe this is a solution for you.

sk1p:
For CI, at least for the main LiberTEM repository we are planning to move to Azure DevOps, which provides more parallel builds than travis, and hopefully a more consistent experience than our current mix of travis and appveyor. They sadly also don't provide GPU access out of the box, but they also allow to run an on-premise "agent". Then, a challenge remains to properly isolate that agent, while still allowing GPU access.

Are you running the gitlab-runner on a dedicated CI system? With a dedicated system, one could run the agent/runner/... in a properly isolated VM and map the GPU into the VM via PCI passthrough.

SimeonEhrig:
Our "public" runner are dedicated systems in a isolated area of the HZDR network (4 Nodes, with different CPU and GPU configurations). We use docker for isolation and setting up the environment.
We also develop a special HPC runner, which can execute slurm jobs from the CI on the cluster. Unfortunately, the HPC runner has no isolation. So we can use it just for trusted private repositories.

sk1p:
> Our "public" runner are dedicated systems in a isolated area of the HZDR network (4 Nodes, with different CPU and GPU configurations). We use docker for isolation and setting up the environment.

Ah, nice! I don't think we have such an isolated network here, at least not as a centralized service. For ptychography 4.0, we should use those HZDR public runners, but I don't think it is appropriate to use them for LiberTEM, sadly. An advantage of our move to Azure DevOps will be that we will then have only a single CI pipeline, as opposed to Travis+AppVeyor; mirroring to gitlab would mean we will lose this again...

SimeonEhrig:
> > Our "public" runner are dedicated systems in a isolated area of the HZDR network (4 Nodes, with different CPU and GPU configurations). We use docker for isolation and setting up the environment.
> 
> Ah, nice! I don't think we have such an isolated network here, at least not as a centralized service. For ptychography 4.0, we should use those HZDR public runners, but I don't think it is appropriate to use them for LiberTEM, sadly. An advantage of our move to Azure DevOps will be that we will then have only a single CI pipeline, as opposed to Travis+AppVeyor; mirroring to gitlab would mean we will lose this again...

The unification of the CI pipeline is a good idea. It's annoying to learn the CI script description language for each CI :-( But in our case, we also use the GitLab runner in our HZDR GitLab instance. So it is not possible to switch to a first party CI for GitHub. We plan to reduce the number of CI system to two. GitHub Actions for compiling tests on Windows, Linux and MacOS and to avoid some mirroring problems and GitLab CI for runtime tests on Linux.

SimeonEhrig:
@uellue 
Do I need to set something special for the CUDA environment? I ran the `cupy-testbed.ipynb` notebook, but it looks like a fallback executed the UDF on the CPU.

The output of Cell 6 is:
``` python
# ctx.executor.client.run(info)
{'tcp://149.220.76.79:33973': ('10', None),
 'tcp://149.220.76.79:34135': ('12', None),
 'tcp://149.220.76.79:34751': ('9', None),
 'tcp://149.220.76.79:35285': ('0', None),
 'tcp://149.220.76.79:36451': ('1', None),
 'tcp://149.220.76.79:36675': ('4', None),
 'tcp://149.220.76.79:36977': ('11', None),
 'tcp://149.220.76.79:40873': ('13', None),
 'tcp://149.220.76.79:41065': (None, None),
 'tcp://149.220.76.79:41747': ('7', None),
 'tcp://149.220.76.79:42321': ('3', None),
 'tcp://149.220.76.79:43127': ('8', None),
 'tcp://149.220.76.79:43443': ('6', None),
 'tcp://149.220.76.79:44147': ('5', None),
 'tcp://149.220.76.79:44783': ('2', None)}
```
on a 14-Core processor and `np.any(res['backend'].data == 'cupy')` return `False`. The system has two different GPUs (Nvidia K20m and Quadro V100).

sk1p:
@SimeonEhrig is there a warning about `CudaSupportError`? How does `numba.cuda.gpus` look like on that machine?

SimeonEhrig:
I found no `CudaSupportError` in the Notebook output or on the terminal.

I executed the following cell in the same notebook:
```python
from numba import cuda

for g in cuda.gpus:
    print(g)

# output:
<Managed Device 0>
<Managed Device 1>

```

SimeonEhrig:
Just as side note. On my system `CUDA_VISIBLE_DEVICES=1` is set by default. But it also doesn't work when I `unset` the variable.

sk1p:
Hmm, the output is missing a line like `'tcp://134.94.162.235:36236': (None, '0'),` - using no CPU and GPU 0.  Maybe you can try to re-raise this exception, to make sure it is 'heard': https://github.com/LiberTEM/LiberTEM/pull/760/files#diff-e7157d6875c615ed1c9ecdc316d0c16cR743

SimeonEhrig:
`cupy` was not installed:

``` python
import numba.cuda
import cupy

if cupy:
    try:
        for device in numba.cuda.gpus:
            cuda_spec = deepcopy(cuda_base_spec)
            cuda_spec['options']['preload'] = \
                f'import os; os.environ["CUDA_VISIBLE_DEVICES"] = "{device.id}"; \
                    os.environ["LIBERTEM_USE_CUDA"] = "{device.id}"'
            workers_spec[f'cuda-{device.id}'] = cuda_spec

    except numba.cuda.CudaSupportError as e:
        # Continue running without GPU in case of errors
        # Keep LiberTEM usable with misconfigured CUDA, CuPy or numba.cuda
        # This DOES happen, ask @uellue!
        warnings.warn(repr(e), RuntimeWarning)
```
output:
```
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-16-1947ac746ad1> in <module>
      1 import numba.cuda
----> 2 import cupy
      3 
      4 if cupy:
      5     try:

ModuleNotFoundError: No module named 'cupy'
```

But I did not find the error message either in the notebook or in the terminal output. Just in this special cell. Now I have CUDA errors. Maybe caused by the two different GPUs. I have to check it.

sk1p:
>But I did not find the error message either in the notebook or in the terminal output

Ah, probably warnings stuff needs special setup to see in notebooks: https://libertem.github.io/LiberTEM/tips.html#show-deprecation-warnings

Maybe the same is needed for runtime warnings? Not sure. Anyway, we should have some kind of functionality that allows validating the cuda/cupy/numba/... setup, @uellue: a function `get_backend_diagnostics` or similar?

SimeonEhrig:
It is a problem of the multi GPU setup. If I restrict it to a single GPU with the environment variable `CUDA_VISIBLE_DEVICES`, everything is fine. But our system is not real world system, because we combine the latest HPC GPU (SM 7.0) with a really old GPU (SM 3.5). So, it is possible that is a problem of `cupy`.

Do you have a homogeneous Multi-GPU system?

sk1p:
> Do you have a homogeneous Multi-GPU system?

Actually we do, but we haven't tested there yet - some nodes of JUWELS do have multiple GPUs: https://www.fz-juelich.de/ias/jsc/EN/Expertise/Supercomputers/JUWELS/Configuration/Configuration_node.html

Looking here https://apps.fz-juelich.de/jsc/hps/juwels/quickintro.html#quick-avail-partitions the nodes in the `develgpus` partition actually have up to 4x V100 and with `--gres=gpu:2` or `--gres=gpu:4` we can request two/four GPUs.

uellue:
@sk1p @SimeonEhrig Thx for testing and feedback! Yes, the whole thing is at quite an early development stage. I still have to test it on a multi-GPU system.

It seems that CuPy can be a bit brittle. On my laptop without a CUDA GPU it is currently broken, not sure if it is because it doesn't find a device or for other reasons. Some ideas on how to handle this, what do you think?

* Possibility to disable CuPy support so that it doesn't even try to load the module, in order to be able to run LiberTEM in an environment with broken CuPy.
* Specify to DO use CuPy to get an error if CuPy is not installed
* Diagnostics, as @sk1p wrote
* Specify which GPUs to use, independent of environment variables, to exclude problematic ones
* Fail hard instead of just warning if CuPy support is enabled and any problems are encountered.

Right now I have to take care of other tasks, in particular finishing this paper: https://github.com/openjournals/joss-reviews/issues/2006

After that I'll look into the feedback and improve the general architecture, make it work on JEWELS with four GPUs, and then we can test again. :-)

SimeonEhrig:
I simply asked because you may have an easier way to test. It's the same for me. I can also use our HPC system. I just need to set it up.

In general, it is really important for a productive application to verify that the CUDA environment is working. There are a lot of real reasons why CUDA does not work. For example, you are using a module environment system and an environment variable is not set correctly.

Before I can give some useful hints on how LiberTEM should behave when CUDA is not working properly, I have a question. Should GPU acceleration be implicit (if there is a GPU worker and the problem size fits, then offload) or explicit (this UDF should run on a GPU)?

For a first draft it is no problem if multi GPU does not work. I just tested mulit GPU because I was not sure if the environment variable `CUDA_VISIBLE_DEVICES` causes problems. So the test was only a side result ;-) In general multi GPU is not easy because it is not implicit. Fortunately, the worker concept of LiberTEM already implements the difficult part of multi GPU support. It should allow homogeneous and heterogeneous GPU setups. In my opinion, the support is limited by the cupy library. We should check how cupy handles multi GPU. In the end a homogeneous GPU setup should work. Heterogeneous setups are not important because I don't know of any production system that has such a configuration.

SimeonEhrig:
> Right now I have to take care of other tasks, in particular finishing this paper: [openjournals/joss-reviews#2006](https://github.com/openjournals/joss-reviews/issues/2006)

It is not a problem if you do not answer immediately. I am constantly switching between different projects, depending on the progress and pressure of each project. I will not run out of work if you don't answer immediately ;-)

SimeonEhrig:
I have successfully tested the Alpaka binding. But there is one problem. I don't know how to limit the worker to GPU workers. A simple CPU worker ran the CUDA algorithm, which is possible on my system.

[TestAlpakaBinding.ipynb.zip](https://github.com/LiberTEM/LiberTEM/files/4580832/TestAlpakaBinding.ipynb.zip)

sk1p:
>  I don't know how to limit the worker to GPU workers. 

This should be possible by implementing the `get_backends` method in your UDF and returning just `('cupy',)`. When using other UDFs, forcing them to run only on GPU is possible using my branch and specifying the backend in `run_udf`. Hope this helps!

SimeonEhrig:
I used this commit: https://github.com/sk1p/LiberTEM/commit/52fb146f7dc1e162c3d2937a774c4c0e31bd705f and modified my Notebook:

```python
class SumOverSig(UDF):
    def get_result_buffers(self):
        return {
            'gpures' : self.buffer(
                kind="nav", 
                extra_shape=(10,), 
                dtype=int),
            'workkind' : self.buffer(
                kind="nav",
                dtype=int)
        }
    
    def process_frame(self, frame):
        array_size = 10
        a = binding.AlgoFI(array_size)
        
        input = a.get_input_view()
        
        for i in range(array_size):
            input[i] = frame[0][i]
        
        a.compute()
        
        self.results.workkind[:] = 2 if os.environ.get("LIBERTEM_USE_CUDA", None) else 1
        self.results.gpures[:] = a.get_output_view()
        
    def get_backends(self):
        return ('cupy', )
```

```python
res = ctx.run_udf(
    udf=SumOverSig(),
    backends=('cupy',),
    dataset=ds)
``` 
Now it looks like the UDF only runs on the CUDA worker.

src/libertem/udf/base.py
```diff
@@ -606,20 +656,43 @@ def get_preferred_input_dtype(self):
...
         }
 
+    def get_backends(self):
+        # TODO see interaction with C++ CUDA modules requires a different type than CuPy
+        '''
+        Signal which computation back-ends the UDF can use.
+
+        :code:`numpy` is the default CPU-based computation.
+
+        :code:`cupy` is CUDA-based computation through CuPy.
```
SimeonEhrig, line 674: I think, the CUDA worker should named `cuda`, because you can use every CUDA-based library.

uellue: In principle, yes. That's why the more low-level interfaces call it "cuda" and not "cupy", for example the resource tags in the cluster. At the level of UDFs, some things are cupy-specific:

* self.xp is set to the cupy module
* tiles/frames/partitions are passed as cupy arrays
* Result buffers are allocated as cupy arrays if `where="device"`
* In the future, datasets will likely read cupy arrays directly

In the future, CuPy may get support for other devices, such as AMD, or we might want to allow using other interfaces than cupy, for example a different device array library. For that reason the UDF specifies the interface it uses, and not the back-end on which the library runs.

src/libertem/udf/base.py
```diff
@@ -400,12 +425,37 @@ def init_task_data(self):
...
+
     def set_meta(self, meta):
         self.meta = meta
 
     def set_slice(self, slice_):
         self.meta.slice = slice_
 
+    def set_backend(self, backend):
+        assert backend in self.get_backends()
+        self._backend = backend
```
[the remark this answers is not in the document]
uellue: Hm I'm trying to wrap my head around that one since these methods will also be called on the leader node, right? That means one shouldn't rely on `self.xp` being the same outside of partition scope.

In that case, one could let `self.xp` return np until instructed otherwise. Is that what you had in mind?

src/libertem/executor/dask.py
uellue: Currently, service doesn't request any resource, that means it can run on any free worker. Would it make sense to restrict service to only specific workers?

src/libertem/api.py
uellue: I could make the service worker multithreaded, just a question of setting options in the `cluster_spec()` function. Probably we should leave that until we have this covered in benchmarks. That way we can see if it makes any difference.

Are there use cases where restricting service tasks to specific workers would be better than just running on any worker?

src/libertem/executor/dask.py
sk1p: Right. I don't know if we should restrict - I think we can see if there are any problems with the unrestricted case. I think currently worst case a GPU worker will be doing non-GPU work?

src/libertem/api.py
sk1p: > Are there use cases where restricting service tasks to specific workers would be better than just running on any worker?

I don't think there are any use cases yet, let's keep it simple for now.

src/libertem/common/buffers.py
```diff
@@ -49,6 +50,21 @@ def zeros_aligned(size, dtype):
...
 
+def to_numpy(a):
+    # .. versionadded:: 0.6.0.dev0
+    cuda_device = os.environ.get("LIBERTEM_USE_CUDA")
+    if isinstance(a, np.ndarray):
+        return a
+    elif cuda_device:
+        # Try to avoid importing unless necessary
+        import cupy
+        cupy.cuda.Device(cuda_device).use()
```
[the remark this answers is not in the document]
uellue: I've moved it to run_for_partition() and use a `try ... finally` block to set it back to the previous one.

src/libertem/udf/base.py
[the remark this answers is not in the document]
sk1p: > will also be called on the leader node, right?

Yes, for getting the dataset-shaped result buffer(s)

>That means one shouldn't rely on self.xp being the same outside of partition scope.

Correct. I think this is mostly important for consistency - if `self.xp` isn't available in `get_result_buffers`, one has to explicitly use numpy, which means when porting a UDF to support GPUs, you have to remember which methods have `xp` available, and which don't. 

> In that case, one could let self.xp return np until instructed otherwise. Is that what you had in mind?

That could work, yes. Don't remember what I had in mind, as it's been a while :grin:

src/libertem/common/buffers.py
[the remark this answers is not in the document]
uellue: ...ah, that was another instance of use(). I've removed this one since it doesn't seem to be necessary -- pending tests on a multi GPU machine.

src/libertem/common/buffers.py
```diff
@@ -49,6 +50,21 @@ def zeros_aligned(size, dtype):
     return res
 
 
+def to_numpy(a):
+    # .. versionadded:: 0.6.0.dev0
+    cuda_device = os.environ.get("LIBERTEM_USE_CUDA")
```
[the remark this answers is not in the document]
uellue: OK, I had to move backend.py to common to avoid a circular import.

src/libertem/executor/dask.py
uellue: Yes, that is the worst that can happen. Let's see if this becomes an issue in the future.

uellue:
@sk1p I remember having a conversation on making it easier to start a local dask cluster. Right now it is quite difficult to get the resources and initialization right. Do you have any idea what could be the best way to handle this?

src/libertem/udf/base.py
SimeonEhrig: Okay, but in the case of using the alpaka-pybind11-bindings, just the CUDA environment is necessary and not the `cupy` library. So there is already a use-case, where `cupy` is not required.

uellue: Yes, good point. Is using CuPy arrays for handling data exchange with C++ CUDA code suboptimal or is it a significant burden to install CuPy? In that case adding support for a third back-end would make sense.

src/libertem/udf/base.py
```diff
@@ -829,106 +881,116 @@ def _get_dtype(self, dtype):
...
-                tiling_scheme=tiling_scheme,
-                roi=roi, dest_dtype=dtype,
-            )
-            # Guaranteed to be the same by UDFTask.get_resources()
-            xp = udfs[0].xp
-            for tile in tiles:
-                # Work-around, should come from dataset later
-                device_tile = xp.asanyarray(tile)
+            try:
+                previous_id = False
```
sk1p: Hmm, this is starting to reach indentation levels that are a bit too much - maybe the `with` and `try...finally` blocks can be done in this function, and the content can be moved into its own method? As this method is central to UDF processing, we need to take care to keep it readable, and over time try to increase readability, not the other way around :grin:

src/libertem/executor/inline.py
```diff
@@ -20,19 +17,13 @@ def run_job(self, job, cancel_id=None):
             yield result
 
     def run_tasks(self, tasks, cancel_id):
-        # The UDFRunner expects one of these environment variable to be set
```
sk1p: So the default backend is now handled in the `UDFRunner`, right?

sk1p:
About the `dist` failures:

- the `docker-compose` file has `--n_cpus` vs. `--n-cpus` which is used in `libertem-worker` (I think this is different from the dask worker, but that's ok)
- once that is corrected, dask complains about the following: `worker-2_1   | distributed.nanny - ERROR - Failed while trying to start worker process: Unexpected response from register: {'status': 'error', 'message': 'name taken, cpu-0', 'time': 1591643237.8349252}` - one of the names used appears to be globally unique, where we assume it is only per-node unique. That's my guess at least...

The other failures look like the test code needs some updates.

src/libertem/udf/base.py
SimeonEhrig: CuPy should not affect the data exchange with CUDA C++. The exchange is done by numpy arrays, which are also the base for CuPy. 

But, I see another problem. CuPy needs to install a package depending on the version of the CUDA toolkit (in my case 10.2). I think, this behaviour make it hard to automate the installation and the package is really huge. In my case about 340 MB. So, I'm not sure, if CuPy should required for CUDA. But in general, it does not cause any problem with the bindings.

sk1p:
The travis integration is for some reason disconnected again, see https://travis-ci.org/github/LiberTEM/LiberTEM/pull_requests for the current results

src/libertem/executor/inline.py
uellue: Actually, in `common.backend.get_device_class()` which returns 'cpu' if no environment variable is set. This is then used in UDFRunner.

src/libertem/udf/base.py
uellue: Ok, I've split it up, it was about time...

src/libertem/udf/base.py
uellue: Ok, I've rewritten the CUDA support in such a way that it should work completely without CuPy. By requesting backend "cuda", a UDF is scheduled on CUDA workers, but otherwise behaves as a regular NumPy UDF, i.e. no CuPy result buffers, input data etc. That is then left to the user. :-)

SimeonEhrig: Sounds good. I will test it with the alpaka-bindings in the next days.

uellue:
About selecting devices: Setting the environment variable `CUDA_VISIBLE_DEVICES` apparently masks out devices so that the numbering starts at 0. This has to happen before any CUDA library is loaded, otherwise it is ineffective. For that reason I'll not set it in `set_use_cuda()`. Users can still use it to make a device invisible before workers start.

uellue:
@SimeonEhrig @sk1p Thx for your help and feedback! I've (hopefully) included all your feedback and fixed all the errors and glitches. Just now I tested it successfully on a quad-GPU system with and without installed CuPy. It should now be ready for review as a merge candidate.

Performance is another topic, see #198 for that. As soon as we have such a benchmarking suite, we can start tuning parameters.

tests/udf/test_device.py
```diff
@@ -0,0 +1,133 @@
...
+
+    use_cpu = bae.get_use_cpu()
+    use_cuda = bae.get_use_cuda()
+    backend = bae.get_device_class()
+
+    with mock.patch.dict(os.environ, {'LIBERTEM_USE_CUDA': "23"}):
+        # This should set the same environment variable as the mock above
+        # so that it will be unset after the "with"
+        bae.set_use_cuda(23)
+        with mock.patch('numpy.cuda', return_value=MockCuda, create=True):
```
sk1p, line 91: The mock may need a quick comment, something like:
```suggestion
        # add `numpy.cuda` so we can make `numpy` work as a mock replacement for `cupy` 
        with mock.patch('numpy.cuda', return_value=MockCuda, create=True):
```

src/libertem/udf/holography.py
```diff
@@ -139,7 +138,7 @@ def get_result_buffers(self):
         else:
             dtype = np.complex128
         return {
-            "wave": self.buffer(kind="nav", dtype=dtype, extra_shape=extra_shape)
+            "wave": self.buffer(kind="nav", dtype=dtype, extra_shape=extra_shape, where='device')
```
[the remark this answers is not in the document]
uellue: The final result on the leader node will be allocated in RAM even if where='device' since the leader node may not even have a GPU. Nevertheless, I've switched it to a version that exports right away, with a FIXME to check what is better.

tests/udf/test_device.py
uellue: Thx! :-)

SimeonEhrig:
> > > @SimeonEhrig @sk1p Thx for your help and feedback! I've (hopefully) included all your feedback and fixed all the errors and glitches. Just now I tested it successfully on a quad-GPU system with and without installed CuPy. It should now be ready for review as a merge candidate.
> > > Performance is another topic, see #198 for that. As soon as we have such a benchmarking suite, we can start tuning parameters.
> > 
> > 
> > I tested the alpaka bindings on our dev system with two GPUs (Quadro V100 + Tesla K20). The result is correct, but I believe, just one GPU was used, because `nvidia-smi` shows just memory usage on one GPU:
> > image
> > Here is my test notebook:
> > [TestAlpakaBinding.pdf](https://github.com/LiberTEM/LiberTEM/files/4758288/TestAlpakaBinding.pdf)
> > Did I something wrong our can you verify the behaviour?
> 
> Discussed in an VC: when the backend `cuda` is used, the developer have to set the cuda device (e.g. `cuda_set_device()`) in the `UDF`. The interface for setting the cuda device should be public and documented.

I [modified the alpaka binding](https://github.com/Ptychography-4-0/ptychography/pull/28) and tested it successful on a dual-GPU system. Now it runs on both GPUs. When this PR is merged, I add the example notebook to Pty 4.0 repo.

src/libertem/io/dataset/frms6.py
```diff
@@ -452,7 +452,7 @@ def get_diagnostics(self):
         global_header = self._get_hdr_info()
         return [
             {"name": "Offset correction available and enabled",
-             "value": str(self._get_dark_frame() is not None)},
```
uellue, line 455: @sk1p This was probably a merge artifact? What is correct here?

src/libertem/io/dataset/k2is.py
```diff
@@ -641,7 +642,7 @@ def _get_syncer(self, do_sync=True):
 
     def _get_fileset(self, with_sync=True):
         sig_shape = (SECTOR_SIZE[0], NUM_SECTORS * SECTOR_SIZE[1])
-        num_frames = int(np.prod(self._get_scansize()))
+        num_frames = int(np.prod(self._scan_size))
```
uellue, line 645: @sk1p This was probably a merge artifact? What is correct here?

src/libertem/io/dataset/frms6.py
sk1p: No, change looks good - this was a fix for a bug that was caught as part of changing the `dist` tests to use the new cli

src/libertem/io/dataset/k2is.py
sk1p: Change looks good, as `_get_scansize` accesses the gtg file, which means `_get_fileset` could only be used on worker nodes.

src/libertem/io/dataset/frms6.py
uellue: Thx! :-)

src/libertem/io/dataset/k2is.py
uellue: Thx! :-)

uellue:
Note before merge: Add test case for running hybrid (numpy, cupy) UDF on cluster with CUDA, but without cupy.

uellue:
@sk1p Sry but I missed the window before 0.50 release of Numba. Fixing Blobfinder is on my list for tomorrow, for the time being a temporary pin in 56b2616.

sk1p:
> @sk1p Sry but I missed the window before 0.50 release of Numba. Fixing Blobfinder is on my list for tomorrow, for the time being a temporary pin in [56b2616](https://github.com/LiberTEM/LiberTEM/commit/56b261690d883dbc63457b36371e56e185e57413).

No worries, a temporary pin should also be fine.

Issue #628, opened 2020-02-20
Allow final postprocessing on the central node in an UDF


uellue:
Discussion with @sk1p and @Brow71189:

Currently, final postprocessing of an UDF's results is performed in a `run_*()` function that is not part of the UDF. If live updating results are required, this has to be re-implemented by the routine that is consuming the live-updating raw results.

In the future, an UDF should be able to specify a final processing routine that transforms raw results into something useful for the user. `StdDevUDF` would be a prime non-trivial example where this is required.

It should be possible to perform the transformation for each merging step so that incremental postprocessed results are available.

sk1p:
Just for naming: how about `UDF.finalize()`?

One important point: making this a method on the UDF will currently limit us to finalizing only using the result of one UDF. This is not necessarily bad, just something to keep in mind.

uellue:
`finalize()` is the name of a destructor in many languages: https://en.wikipedia.org/wiki/Finalizer.

What about `consolidate()`?

uellue:
It should probably be a pure function that doesn't modify the state of the UDF. In that case we can call it whenever we need consolidated partial results.

sk1p:
> It should probably be a pure function that doesn't modify the state of the UDF. In that case we can call it whenever we need consolidated partial results.

:+1:

> finalize() is the name of a destructor in some languages

Tried my best to forget about Java :grin:

> What about consolidate()?

Hmm, to me this sounds like a merge/reduce-ish operation, but really this could be any kind of transformation. What about the more general `get_results`? It may also make it clear that it is a getter and not really a function where you should put anything computationally intensive

uellue:
> What about the more general `get_results`?

:+1:

PR #764, opened 2020-04-29
WIP: Partitioning based on ROIs

### LINKED ISSUE #651: UDF computations on overlapping ROIs with a common merge step
Discussions with https://github.com/Ptychography-4-0/ptychography/ team:

Ptychography reconstruction requires partial reconstruction and stitching of regions that are close and compact in sample coordinates and not necessarily in the "natural order" and current partitioning of the dataset, which could be a random scan with temporal ordering.

To enable distributed ptychography, it should be possible to run an UDF in such a way that data that belongs to a global ROI is passed as one data block to a reconstruction algorithm, i.e. "`process_partition()`" The local result is then merged into a global result buffer with a non-trivial merge function.

The actual reconstruction and merge can be swapped out as needed and the size and shape of regions can be adapted to the "sweet spot" of various algorithms and hardware as needed.

sk1p:
Closes #651 - this is still only a simple sketch of the API, although the basic functionality is already there.

Refs https://github.com/Ptychography-4-0/ptychography/issues/6

Still to do:

- [ ] The actual implementation needs changes in at least the `UDFRunner`, which depends on having a slice associated to the `Partition`.
- [ ] Maybe add `Partition.shape_for_roi(roi)` method, which can override the current `partition_shape=partition.slice.adjust_for_roi(roi).shape` used when constructing the `UDFMeta` object, which can then also be used in the short-circuit logic that decides not to generate tasks for empty ROIs
- [ ] "nested" ROI support - you should still be able to pass a `roi` argument to `run_udf` on a `RoiDataset`. Needs tests for the case that the `Partition` ROIs actually overlap, and the `roi` selects elements from the overlapping region, etc.
- [ ] Implement `DataSet.meta` property on all `DataSet` impls
- [ ] Documentation

Question about partition sizing: as the user may have requirements for overlapping etc. we don't have control over the size and number of partitions, as we can't use generic logic here. I guess this is the question: where do the partition ROIs come from? Should the user pass in a callable that generates the ROIs? For example, the decisions could depend on the number of GPUs available on the system, and the available memory, so this is actually a per-node decision. I guess the size of each ROI should fit the memory of the "smallest" GPU, for the edge case of having more than one GPU, but with different memory sizes.

## Contributor Checklist:

changes were made to the GUI)


sk1p:
So, in numpy, once you apply fancy indexing, the affected dimensions are collapsed into one flat dimension, which is "compressed" to the number of nonzero entries. This is also how I currently handle the "normal" ROI.

When dealing with overlapping ROIs, assigning coordinates to tiles becomes a bit more difficult. Just passing on the coordinates from the wrapped dataset (as in the API sketch in this PR) does not work, as the coordinates would all start at zero. My proposal here is to also flatten the navigation axis of the whole `RoiDataSet`, and have the navigation axis expand, so its navigation shape would be basically `sum(np.count_nonzero(roi) for roi in rois)`. The origin of each partition would be determined by the number of non-zero entries in the preceding ROIs in the list.

uellue:
Discussion with @sk1p on behavior (WIP):

Two options to implement the navigation dimension:
1. If the ROI partitions don't overlap and cover the entire navigation space, the interface should work with all existing UDFs and yield results as if it was running with default partitions (keep the navigation dimension of the original dataset).
2. The ROI partitions give results in a new navigation dimension, which is a concatenation of all the ROIs.

In the second case, we need tools to translate between the coordinates in the new navigation dimension and the navigation dimension of the underlying dataset.

As a comment, the automatisms of `kind="nav"` result buffers will probably not be too useful for ROI partitions since ROI partitions are introduced for use cases where normal 1:1 merging isn't sufficient. For that reason, `kind="single"` result buffers with suitable `extra_shape` will likely be used often.

* Running an UDF with ROI partitions yields live-updating results as the partition results come in. That means a usable result can not only be generated in a final postprocessing step, but there are ways to generate them at every merging step. Refs #628 

* The merge function needs information on the ROI of the current incoming partition result.

sk1p:
> In the second case, we need tools to translate between the coordinates in the new navigation dimension and the navigation dimension of the underlying dataset.

Maybe the `DataSet` could provide a hook for coordinate transformation, which is then used by `UDFMeta.coordinates` - so the default implementation works like the implementation from #793 by @AnandBaburajan, but the `ROIDataSet` can override this to provide the "original" coordinates, "punching" through to the underlying `DataSet`. As a next step, we can then see if we need to provide additional helper functions, or extensions to the UDF interface.

uellue:
> Maybe the `DataSet` could provide a hook for coordinate transformation, which is then used by `UDFMeta.coordinates` - so the default implementation works like the implementation from #793 by @AnandBaburajan, but the `ROIDataSet` can override this to provide the "original" coordinates, "punching" through to the underlying `DataSet`.

Sounds good! 👍

uellue:
FYI, in the Ptycho 4.0 call we just discussed this item. Since Simeon is picking up work on the CUDA/Alpaka side of this, we should put it on the short list after 0.6 is out. We can already test with SSB and @Sniper2k 's merge function, taking into account that SSB results change if you constrain it differently. :-)

PR #775, opened 2020-05-07
SEQ: UnicodeDecodeError exception

AnandBaburajan:
Exception for the following error which can be reproduced by loading a raw file:
`UnicodeDecodeError: 'utf-16-le' codec can't decode bytes in position 452-453: il
legal encoding`

## Contributor Checklist:

### DIFF
--- a/src/libertem/io/dataset/seq.py
+++ b/src/libertem/io/dataset/seq.py
@@ -240,7 +240,7 @@ class SEQDataSet(DataSet):
                     "scan_size": str(image_count),
                 }
             }
-        except OSError:
+        except (OSError, UnicodeDecodeError):
             return False
 
     @classmethod
--- a/tests/io/datasets/test_seq.py
+++ b/tests/io/datasets/test_seq.py
@@ -1,0 +1,6 @@
+from libertem.executor.inline import InlineJobExecutor
+from libertem.io.dataset.seq import SEQDataSet
+
+def test_detect_unicode_error(default_raw, lt_ctx):
+    path = default_raw._path
+    SEQDataSet.detect_params(path, InlineJobExecutor())


sk1p:
Good catch! It would be good to have a regression test for this; something like this should reproduce the error, right?

```python
import pytest

from libertem.executor.inline import InlineJobExecutor
from libertem.io.dataset.seq import SEQDataSet


def test_detect_unicode_error(default_raw, lt_ctx):
    path = default_raw._path
    SEQDataSet.detect_params(path, InlineJobExecutor())
```

AnandBaburajan:
> Good catch! It would be good to have a regression test for this; something like this should reproduce the error, right?

Yes, thanks! I'll be sure to add tests next time.

sk1p:
Thanks, merged!

uellue:
A little nitpick after the fact: The regression test should make sure the error is triggered reliably. Does the default raw dataset do that and can we be sure it continues doing so in the future?

Anyway, I don't think it matters here, let's leave it as-is. We fix the problem properly if it crops up again. :-)

sk1p:
> Does the default raw dataset do that
> don't know how probable it is to hit invalid unicode vs. valid unicode

I was curious and had a look into it. The way `_mk_random` currently generates test data means we get only the values from the set `{0.0, 1.0, 288.0, 2880.0}` for `default_raw`. The binary representation of those float values all start with a `'\x00'` - so the first string value, which is at offset `4`, will always be truncated to zero length and trigger the error. If we do change the function to generate values that have a larger fractional value in the binary representation (which we might want to do!), this assumption will no longer hold.

> the content of the default might change in the future since it is not designed to trigger unicode errors...

That's true. The test could be updated to reliably trigger the problem with fixed input data, for example all-zeros in the first kb of data.

Issue #335, opened 2019-05-14
Is there an efficient way to share data between workers?


uellue:
MaskJobs that match a dataset against a large template library may benefit from sharing the library between workers on a node. In particular, L3 cache use on machines with many cores should be much improved. Is there an efficient and portable way to achieve this?

I was reading through the dask.distributed documentation, and it seems to be possible to share data with persist() and scatter(). However, I couldn't figure out if the data is copied to the workers or resides in shared memory.

What would be the best way to test the behavior of dask.distributed? Or is there a more efficient and easy way to share the data?

sk1p:
> Is there an efficient and portable way to achieve this?

Hmm, something like [this](https://docs.python.org/2/library/multiprocessing.html#multiprocessing.Array)? 

> I was reading through the dask.distributed documentation, and it seems to be possible to share data with persist() and scatter(). However, I couldn't figure out if the data is copied to the workers or resides in shared memory.

AFAIK, the data is copied (I could be wrong, though! need to check the source), but using `scatter()` should already a first step into the right direction, as currently static auxiliary data (like masks) is sent to the worker process with each task (see also #287).

Did you try to run dask.distributed workers with threads instead of processes? Then, if we scatter the data, it should be shared, or at least possible to share.

uellue:
> Hmm, something like [this](https://docs.python.org/2/library/multiprocessing.html#multiprocessing.Array)?

Yes, that could be an option. One would have to coordinate the workers on a node so that only one creates the shared memory and the others receive handles to it. I was hoping that someone else already implemented that...

> AFAIK, the data is copied (I could be wrong, though! need to check the source), but using `scatter()` should already a first step into the right direction, as currently static auxiliary data (like masks) is sent to the worker process with each task (see also #287).

Yes, I was also assuming copying, but don't know for sure. I haven't tried it yet, I wanted to ask before doing work.

uellue:
Another option could be something like a dataset. That would be a rather natural way to make larger persistent data available at workers.

sk1p:
Just as a note, the [ray distributed AI framework](https://ray.readthedocs.io/en/latest/index.html) uses a [shared-memory object store](https://ray.readthedocs.io/en/latest/plasma-object-store.html) for this purpose.

uellue:
[...]

PR #778, opened 2020-05-07
WIP: Initial generic corrections support

sk1p:
- [x] support in the Python API
- [ ] support for loading correction data from files (moved to  #808)
- [ ] support for uploading correction data (moved to #807)
- [ ] support for corrections settings in the GUI (moved to #807)
- [x] hot/dead pixel corrections
   - [ ] XML metadata from SEQ datasets (moved to #805)
   - [ ] From Gatan service configuration file, thx @ju-bar :-)
- [ ] per-UDF preferences (moved to #804)
   - [ ] Implement corrections in `MaskContainer` and `ApplyMasksUDF` (moved to #806)
- [ ] Performance tuning (moved to #809)

/CC @jhgee

## Contributor Checklist:

changes were made to the GUI)


sk1p:
> We can include a dead/hot pixel map as well

Yeah, I was actually already starting to write a comment about that :grin: We have to think about how the user can best specify those. The SEQ datasets actually includes a `<BadPixels/>` element in an XML sidecar file, so we should probably support something like that. In that case, whole rows are actually marked as bad, so I don't know how well this will work with the current correction code. The XML format also supports single pixels, though. In addition to `DataSet`-provided dead pixel maps, we should allow user-specified fancy-indexing formats, i.e. boolean arrays or coordinate lists.

In what formats should the GUI accept bad pixel maps? Also as binary maps or coordinate lists, in what formats? .mat, .npz, csv...?

> How would that work in this context?

I'm not sure about that yet, I'll add it to the TODO above, though.

uellue:
> > We can include a dead/hot pixel map as well
> 
> Yeah, I was actually already starting to write a comment about that 😁 We have to think about how the user can best specify those. The SEQ datasets actually includes a `<BadPixels/>` element in an XML sidecar file, so we should probably support something like that. In that case, whole rows are actually marked as bad, so I don't know how well this will work with the current correction code. The XML format also supports single pixels, though. In addition to `DataSet`-provided dead pixel maps, we should allow user-specified fancy-indexing formats, i.e. boolean arrays or coordinate lists.

The correction back-end expects a list of coordinates, so that would match well here. The correction back-end can detect and ignore neighboring dead pixels, it just needs at least one good pixel next to each bad pixel.

> In what formats should the GUI accept bad pixel maps? Also as binary maps or coordinate lists, in what formats? .mat, .npz, csv...?

See https://groups.google.com/d/msg/libertem/0u_xCjPVf5Y/v7Aj0tJABwAJ :-)

src/libertem/udf/base.py
```diff
@@ -703,7 +706,8 @@ def __init__(self, udf, debug=False):
     def _get_dtype(self, dtype):
         return np.result_type(self._udf.get_preferred_input_dtype(), dtype)
 
-    def run_for_partition(self, partition: Partition, roi):
+    def run_for_partition(self, partition: Partition, roi, corrections):
```
uellue, line 709: maybe sth. like this for UDF support:

* Include corrections in UDFMeta

```python
if not self._udf.does_corrections():
    partition.set_corrections(corrections)
```

uellue:
Another question: Corrections are most efficient when applied to a stack of frames. If I understand the code correctly, the `DataSet` CAN return a stack of frames, which is then applied one by one within the `UDFRunner`. Do all the DataSet implementations actually do return several frames here? The balance between using the L3 cache for a single frame (frame size in the region of 1/3 MB -- we have two correction maps of the same size as the frame!) vs. performance gains from correcting several frames at once using the correction data in the L1 cache are not so trivial. If one frame is larger than 1/3 MB, a stack of frames is likely more efficient here.

Long story short, tiled processing should be of great advantage when using corrections.

Referring to #335, the correction data would be a perfect use case for shared memory, by the way. That would be a great relief for the L3 cache on machines with many cores.

Another thought about the bad pixel maps: One can construct a boolean sparse matrix from the list, use slicing on that and then convert back to a list for tiled processing. Only disadvantage is that pixels on a tile boundary would not use the entire environment for patching, but only the one within the tile. I am not sure how much of a problem it is, but in some edge cases it might change results depending on the tiling.

uellue:
Referring to the e-mail from @rafaldb, the code already allows to use a different `CorrectionSet` class.

The code to patch masks should actually be part of the `CorrectionSet`, similar to `apply()`, since it is closely linked to the mathematical model used for correction. The current "constant offset and linear gain, average patching" implementation can patch masks. However, with a non-linear gain correction this wouldn't be possible anymore. That means the UDF should be able to probe the `CorrectionSet` for correction capabilities and boundary conditions within its proposed `does_corrections()` method. `ApplyMasksUDF`, for example, should fall back to receiving corrected data if the correction is not linear.

sk1p:
> Another question: Corrections are most efficient when applied to a stack of frames. If I understand the code correctly, the `DataSet` CAN return a stack of frames, which is then applied one by one within the `UDFRunner`.

Correct.

> Do all the DataSet implementations actually do return several frames here?

Not yet - the SER reader being the only exception. I'm right now returning frame-by-frame to avoid copying, but I think that's easily fixable, even if it may require a patch to `ncempy`.

> The balance between using the L3 cache for a single frame (frame size in the region of 1/3 MB -- we have two correction maps of the same size as the frame!) vs. performance gains from correcting several frames at once using the correction data in the L1 cache are not so trivial. If one frame is larger than 1/3 MB, a stack of frames is likely more efficient here.
> 
> Long story short, tiled processing should be of great advantage when using corrections.

Agreed.

> Referring to #335, the correction data would be a perfect use case for shared memory, by the way. That would be a great relief for the L3 cache on machines with many cores.

Also agreed.

> Another thought about the bad pixel maps: One can construct a boolean sparse matrix from the list, use slicing on that and then convert back to a list for tiled processing. Only disadvantage is that pixels on a tile boundary would not use the entire environment for patching, but only a part within the same tile. I am not sure how much of a problem it is, but in some edge cases it might change results depending on the tiling.

Do we need to add the bad pixel list to the negotiation input, so we can generate a tiling scheme that doesn't have bad pixels at boundaries?

> The correction back-end expects a list of coordinates, so that would match well here

The point here being that both are supported, for ease of use (I know it is easy to transform back and forth, but the user shouldn't have to know...)

> That means it can patch an entire row without problems

Perfect!

> Referring to an e-mail from @rafaldb

Didn't get that one.

> The code to patch masks should actually be part of the CorrectionSet, similar to apply(), since it is closely linked to the mathematical model used for correction. The current "constant offset and linear gain, average patching" implementation can patch masks. However, with a non-linear gain correction (edit: or patching) this wouldn't be possible anymore. That means the UDF should be able to probe within its proposed does_corrections() method the CorrectionSet for correction capabilities and boundary conditions. ApplyMasksUDF, for example, should fall back to receiving corrected data if the correction is not linear.

So the interface should allow a UDF to modify the `CorrectionSet`, for example by deactivating a certain correction?

```python

# default case
def adjust_corrections(self, corrections):
    return corrections

# masks
def adjust_corrections(self, corrections):
    if corrections.something is not "good":
        return corrections.without_something()\
                          .without_something_else()
    return corrections

# unconditionally deactivate:
def adjust_corrections(self, corrections):
    return CorrectionSet()  # or possibly corrections.empty() so the user doesn't have to import stuff
```

This is then similar to the `adjust_tileshape` interface of the `DataSet`.

uellue:
> So the interface should allow a UDF to modify the `CorrectionSet`, for example by deactivating a certain correction?

Oh, that's a neat idea, to grab parts of the correction! We just have to make sure that the UDF really understands what the correction does, for example by probing if it is an instance of a supposed `LinearCorrectionSet`, and only takes over if it can make sure the result is equivalent.

With "parts of a correction", do you mean detector correction vs. correction of other physical coordinates in the future, or parts of a detector correction?I'm not sure if deactivating parts of a detector correction would bring much benefit compared to either doing all the detector corrections or none.

uellue:
> Do we need to add the bad pixel list to the negotiation input, so we can generate a tiling scheme that doesn't have bad pixels at boundaries?

Now THAT would be neat! :-) :-) :-)

uellue:
> The point here being that both are supported, for ease of use (I know it is easy to transform back and forth, but the user shouldn't have to know...)

I guess we can just support both since they are easily distinguished by shape. Probably we just convert to the sparse "list of pixels" format immediately in the constructor?

uellue:
> So the interface should allow a UDF to modify the `CorrectionSet`, for example by deactivating a certain correction?

Ah, just for clarity: The current implementation assumes that the gain correction is a constant multiplier, which is a good first approximation in many cases. However, real-world detectors have non-linear gain curves, as @ju-bar and @rafaldb wrote. In the future the user may want to use a different mathematical model for correction, i.e. a completely different implementation of a `CorrectionSet` set that takes nonlinearity into account, for example gamma curves as they are used in photography.

uellue:
> Not yet - the SER reader being the only exception. I'm right now returning frame-by-frame to avoid copying, but I think that's easily fixable, even if it may require a patch to `ncempy`.

An issue for the "good first issue" category?

sk1p:
> > Do we need to add the bad pixel list to the negotiation input, so we can generate a tiling scheme that doesn't have bad pixels at boundaries?
> 
> Now THAT would be neat! :-) :-) :-)

It just won't work (efficiently?) if a whole column of pixels is missing. Another possibility is reading overlapping tiles, which are then cropped by the correction mechanism. I think we can leave that for later, when this becomes a problem in practice.

> I guess we can just support both since they are easily distinguished by shape. Probably we just convert to the sparse "list of pixels" format immediately in the constructor?

Yeah... I had in mind that you could just use fancy indexing into a coordinate array with either format without checking the shape, but that doesn't seem to be as easy as I remember.

> Ah, just for clarity: The current implementation assumes that the gain correction is a constant multiplier, which is a good first approximation in many cases. However, real-world detectors have non-linear gain curves, as @ju-bar and @rafaldb wrote. In the future the user may want to use a different mathematical model for correction, i.e. a completely different implementation of a CorrectionSet set that takes nonlinearity into account, for example gamma as it is used in photography.

We have to take care that the API stays valid in the future. That means that the return value from `adjust_correction` should stay valid even in the face of any changes in the future.

> An issue for the "good first issue" category?

With copy, yes, without, it is an issue in the openNCEM project /cc @ercius

uellue:
> > > Do we need to add the bad pixel list to the negotiation input, so we can generate a tiling scheme that doesn't have bad pixels at boundaries?
> > 
> > 
> > Now THAT would be neat! :-) :-) :-)
> 
> It just won't work (efficiently?) if a whole column of pixels is missing. Another possibility is reading overlapping tiles, which are then cropped by the correction mechanism. I think we can leave that for later, when this becomes a problem in practice.

👍 

> > I guess we can just support both since they are easily distinguished by shape. Probably we just convert to the sparse "list of pixels" format immediately in the constructor?
> 
> Yeah... I had in mind that you could just use fancy indexing into a coordinate array with either format without checking the shape, but that doesn't seem to be as easy as I remember.

As far as I still understand my own Numba code, it is built around a list of pixels and wouldn't be so easy to change.

> > Ah, just for clarity: The current implementation assumes that the gain correction is a constant multiplier, which is a good first approximation in many cases. However, real-world detectors have non-linear gain curves, as @ju-bar and @rafaldb wrote. In the future the user may want to use a different mathematical model for correction, i.e. a completely different implementation of a CorrectionSet set that takes nonlinearity into account, for example gamma as it is used in photography.
 
> We have to take care that the API stays valid in the future. That means that the return value from `adjust_correction` should stay valid even in the face of any changes in the future.

Hm, that would make an "all or nothing" approach easier. If the UDF positively identifies the correction and knows what to do, it takes over. If anything seems unsure, it falls back to consuming corrected data. IMHO it could even make sense to flag a "constant shift, linear gain, linear patching" correction as something special since it will allow folding the correction into any UDF that is a linear transformation: f(u + c*v) = f(u) + c*f(v) https://en.wikipedia.org/wiki/Linear_map

As an example, correcting the result of `SumUDF` would be trivial: Correct and patch it like a single detector frame, just with the dark frame multiplied by the number of frames.

jhgee:
I just tested a few .seq files and the dark corrections work! :-)
As a side note (most probably better located somewhere else) : It would be super helpful if additional frames would be ignored. For example I have scans with 4100 frames for a 64x64 scan. would be nice if the additional 4 frames would be ignored.

AnandBaburajan:
> It would be super helpful if additional frames would be ignored. For example I have scans with 4100 frames for a 64x64 scan. would be nice if the additional 4 frames would be ignored.

I'm working on that, thanks for the feedback!

sk1p:
> Hm, that would make an "all or nothing" approach easier. If the UDF positively identifies the correction and knows what to do, it takes over. If anything seems unsure, it falls back to consuming corrected data.

Sounds good. 

> IMHO it could even make sense to flag a "constant shift, linear gain, linear patching" correction as something special [...]
> As an example, correcting the result of SumUDF would be trivial: Correct and patch it like a single detector frame, just with the dark frame multiplied by the number of frames.

Here are some strawmen examples for discussion:

The very explicit way:

```python
def postprocess(self):
    my_own_apply_corrections_fn(self.results.something_sig_shaped, self.meta.corrections)

def get_correction_preferences(self):
        return {"custom_corrections": self.meta.set_of_corrections == {UDF.CORR_DARK_FRAME, UDF.CORR_GAIN_MAP, UDF.CORR_PATCH_PIXELS}}
``` 

Something a bit more "automatic", for supporting all kinds of linear corrections:

```python
def postprocess(self):
    self.meta.corrections.apply_linear_corrections(self.results.something_sig_shaped, number_of_frames)

def get_correction_preferences(self):
    return {"custom_corrections": self.meta.set_of_corrections == UDF.CORR_SET_LINEAR}
``` 

Support for patching masks:

```python
def get_task_data(self):
    masks = ...  # possibly need to fold this into the mask container somehow, or the mask factory fn
    self.meta.corrections.patch_masks(masks)
    return {"masks": masks}

def get_correction_preferences(self):
    # we know what can be patched into masks, so we can also provide that as a set of identifiers:
    return {"custom_corrections": self.meta.set_of_corrections == UDF.CORR_SET_MASK_PATCH}
```

sk1p:
As this PR is already quite large, I'm trying to get it into a state that is ready for check-in, and to tackle remaining issues in separate independent PRs.

@uellue it would be helpful to get feedback on the `excluded_pixels` and `detector.correct`-related changes - do you think it is reasonable to use a `sparse.COO` array as the common interface?

sk1p:
Trying to revive travis...

sk1p:
Closing and re-opening to trigger a travis build

uellue:
Using a sparse.COO for the excluded pixels sounds good! That makes it easy to change things later, since it builds on the NumPy ndarray interface.

If I understood the code correctly, one would use it like this?

```python
ds = ctx.load('frms6', path=..., )
udf = ...
res = ctx.run_udf(udf=udf, dataset=ds, corrections=ds.get_correction_data())
```

If the dataset doesn't provide corrections or if you'd like to use different ones, one would explicitly create a `CorrectionSet` with the desired parameters?

If a dataset has corrections, they will not be applied automatically, right?

Generally, this looks like a good point to start! We can then add more features step by step. :-)

uellue:
Looking at the code, it will likely generate a significant merge conflict with #760. Let me know if/how I can help if it gets too messy!

sk1p:
> Using a sparse.COO for the excluded pixels sounds good! That makes it easy to change things later, since it builds on the NumPy ndarray interface.

:+1:

> If I understood the code correctly, one would use it like this?
> 
> ```python
> ds = ctx.load('frms6', path=..., )
> udf = ...
> res = ctx.run_udf(udf=udf, dataset=ds, corrections=ds.get_correction_data())
> ```
> 
> If the dataset doesn't provide corrections or if you'd like to use different ones, one would explicitly create a `CorrectionSet` with the desired parameters?

Yes.

> If a dataset has corrections, they will not be applied automatically, right?

They will be automatically applied - currently this is done in `api.py` and `web/jobs.py` - I think it may need to be moved to the `UDFRunner` instead, if possible, so we don't need to set the default behavior in two different places.

> 
> Generally, this looks like a good point to start! We can then add more features step by step. :-)

That's the plan! :grin:

sk1p:
> Looking at the code, it will likely generate a significant merge conflict with #760. Let me know if/how I can help if it gets too messy!

Yes, indeed, I should manage, but otherwise we can pair on this via skype.

uellue:
> They will be automatically applied - currently this is done in `api.py` and `web/jobs.py` - I think it may need to be moved to the `UDFRunner` instead, if possible, so we don't need to set the default behavior in two different places.

Ah found it! Yes, I looked in `UDFRunner`... 😄

uellue:
> Yes, indeed, I should manage, but otherwise we can pair on this via skype.

We can also merge this first and then I rebase the CuPy stuff. :-)

sk1p:
> > Yes, indeed, I should manage, but otherwise we can pair on this via skype.
> 
> We can also merge this first and then I rebase the CuPy stuff. :-)

That could work, too. I'll first finish the review of #760, then I'll add finishing touches here and we can merge this.

PR #779, opened 2020-05-08
Dynamic dimension fields

### LINKED ISSUE #233: Improve dataset open forms
Since LiberTEM is getting better at guessing correct parameters for opening a dataset, users start to develop the expectation that default parameters are correct. For that reason we should remove all default parameters that are not verified against the data set. That includes the scan dimension of K2IS data sets.

AnandBaburajan:
Fixes #233 
Currently more than 2 scan dimensions are allowed. I'll fix it to 2D after review.

>I would suggest to keep keyboard users in mind: it should not be harder for a keyboard-only user to enter a tuple-style value than it currently is. For example, the focus could jump to the next field when hitting comma, possibly adding a new field if the user is on the last one, making sure tabbing between fields works well etc.

:thumbsup:

## Contributor Checklist:

changes were made to the GUI)


sk1p:
The behavior is really nice :+1: some details can be improved, for example the icons could have a `title` tooltip that describes what will happen. Maybe instead of the trash can, we can have a minus icon?

Otherwise, really nice! I think the next step could be to extract this functionality into its own component, what do you think? Let me know if you need help with that!

AnandBaburajan:
> I think the next step could be to extract this functionality into its own component, what do you think?

> The fields will probably only be used for raw data in the future after detection is completely implemented.

@sk1p To confirm, only the RAW format should have the flexible fields, right? And the add-analysis page would have the the fields after reshaping is implemented.

sk1p:
> To confirm, only the RAW format should have the flexible fields, right?

> after detection is completely implemented.

It turned out that for some formats, the user still needs to specify the `scan_size`. That means at least the `scan_size` of MIB and SEQ formats, too (as discussed in #233, currently fixed to 2 dims). In case of SEQ, we don't know the `scan_size` at all, only the number of images - in case of MIB, the `scan_size` can also be changed, for example if the `.hdr` file is missing, or contains wrong information.

The `tileshape` parameter should now be gone completely, so that we don't have to take care of that anymore.

> the add-analysis page would have the fields after reshaping is implemented.

I'm not convinced it makes sense to put the reshaping UI in "add analysis" - at least as long as we don't support other-than-2D visualization, you would not be able to open a, for example, 3D-dataset-that-is-really-4D without reshaping, so you would not even get to the step of pressing "add analysis".

I think it makes more sense to reshape data on opening, or directly after opening, or possibly using a "settings"-like button to fine-tune (for example for the case where we have to ignore extra data at the beginning/end, for manual synchronization).

AnandBaburajan:
Max number of fields is 4 right now, will fix it to 2 after the functionality is approved. About ScanSize's cognitive complexity, please let me know if I missed any best practices, thanks!

AnandBaburajan:
> That means at least the `scan_size` of MIB and SEQ formats, too (as discussed in #233, currently fixed to 2 dims). In case of SEQ, we don't know the `scan_size` at all, only the number of images

Sorry, I just found out that `libertem.io.dataset.seq` actually returns the number of images. So should the initial scan size be just [ 64 ] or [ 64 ] [  ] or [  ] [  ] ? But the actual scan size might be
[ 8 ] [ 8 ], for example.

client/src/dataset/components/ScanSize.tsx
```diff
@@ -0,0 +1,70 @@
...
+        setFieldValue("scan_size", scanSize.toString());
+      }
+    }
+
+    return (
+      <>
+        <Form.Group>
+          {scanSize.map((val, idx) => {
+            const scanRef = (ref:FormFieldProps) => { scanRefsArray.current[idx] = ref; }
+            return <Form.Field key={idx} width={2}><Field name={"scan_size_" + idx} id={"id_scan_size_" + idx} type="number" value={val} innerRef={scanRef} onChange={onScanSizeChange} onKeyDown={onCommaPress} /></Form.Field>
```
sk1p, line 59: Hmm, passing information via the `name` like this is a bit brittle - if for some reason the `name` needs to change, this logic will break. For example, let's say you want to use `ScanSize` also for a field called `some_other_tuple` - now the index won't be at position 2 anymore.

I would suggest to extract the "part" as a separate component, so for each tuple element, you would use this component, like this:
```
<ScanSizePart key={idx} value={val} onChange={...} {...etc} />
```

Then, inside that component, you know the `idx` and can pass it up to the change handler in `ScanSize`, something like this:

```
const ScanSizePart = ({ idx, onChange, value }) => {
   const onPartChange = (e) => {
      onChange(idx, e.target.value);
   }

   return <input type='number' onChange={onPartChange} value={value} />; 
}
```

Then, `ScanSize` can still do its state management in `onScanSizeChange`, just without needing to parse the `name` anymore.

I'm also not sure it is the right thing to use the Formik `<Field />` for the parts - as the documentation states, "<Field /> will automagically hook up inputs to Formik. It uses the name attribute to match up with Formik state." - so you end up with values for the parts in the formik state. Maybe just a plain input would work here? The hooking up to formik state part you do manually anyways, using `setFieldValue`...

client/src/dataset/components/ScanSize.tsx
```diff
@@ -0,0 +1,70 @@
...
+const ScanSize: React.FC<ScanSizeProps> = ({ value, minScan, maxScan, setFieldValue }) => {
+
+    const [scanSize, setScanSize] = React.useState(value.split(",").length>=minScan? value.split(","): new Array(minScan).fill(""));
+
+    const scanRefsArray = React.useRef<FormFieldProps[]>([]);
+
+    const onScanSizeChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+      const idx = parseInt(e.target.name.split("_")[2], 10);
+      scanSize[idx] = e.target.value.toString();
+      setScanSize(scanSize);
```
sk1p, line 21: Hmm, I'm not sure this is right. You are modifying an element of `scanSize`, and then call `setScanSize` with the "same" `scanSize` object as it was before. "same" meaning same object identity here. The `useState` docs say this: "If your update function returns the exact same value as the current state, the subsequent rerender will be skipped completely." As I understand, this is a shallow equality, so it will only check the identity of the object passed, not the values contained inside, like array elements. Then, calling `setScanSize` does not trigger a re-render, which may also be the reason you had to fiddle with the `key` in the forms.

If you instead return a copy of the array, with the new values, you should get a re-render when calling the setter. [Here is a codesandbox that demonstrates the issue:](https://codesandbox.io/s/vigilant-turing-bs9ej?file=/src/App.js) clicking the "set a" button does not immediately cause a re-render, clicking "set value", which creates a new object, triggers a re-render.

So generally, the value returned from `useState` should be considered immutable, even if JavaScript allows you to mutate it.

client/src/dataset/components/ScanSize.tsx
```diff
@@ -0,0 +1,70 @@
...
+      setFieldValue("scan_size", scanSize.toString());
+    };
+
+    const onCommaPress = (e: KeyboardEvent) => {
+        const idx = parseInt((e.target as HTMLInputElement).name.split("_")[2], 10);
+        if(e.keyCode === 188){
+            idx===(scanSize.length-1)? newScanDim() : scanRefsArray.current[idx+1].focus();
```
sk1p, line 28: This is really nice behavior :+1: Note that for conditions, it is always better to use a proper `if` statement, if possible. Inside JSX expressions, it is often not possible, so using the tenary operator is fine, but here, a proper `if` is much better readable.

client/src/dataset/components/ScanSize.tsx
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    const newScanDim = () => {
+      if(scanSize.length < maxScan) {
+        scanSize.push("");
+        setScanSize(scanSize);
```
sk1p, line 35: Same as the above `setScanSize` comment.

docs/source/formats.rst
```diff
@@ -63,8 +59,8 @@ There are some common parameters across data set types:
   The name of the data set, for display purposes. Only used in the GUI.
 `scan_size`
   In the GUI, we generally support visualizing data containing rectangular 2D scans. For
-  some data set types, you can specify a scan_size as a tuple `(y, x)`; for
-  others it is even required. When using the Python API, you are free to use n-dimensional
+  some data set types, you can specify a scan_size as (scan width | scan height); for others
+  it is not even required. When using the Python API, you are free to use n-dimensional
   `scan_size`, if the data set and chosen analysis supports it.
```
sk1p, line 64: Thanks for keeping the documentation in mind!

`(y, x)` is `(height, width)`. This is also what is described in the "we follow numpy convention" paragraph - I think we should keep that!

The "for others it is even required" was referencing formats like raw or seq that absolutely need a scan size argument to properly read the data. Maybe this is misleading and that part of the sentence can be completely removed.

client/src/dataset/components/ScanSize.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+    return (
+      <>
+        <Form.Group>
+          {scanSize.map((val, idx) => {
+            const scanRef = (ref:FormFieldProps) => { scanRefsArray.current[idx] = ref; }
+            return <Form.Field key={idx} width={2}><Field name={"scan_size_" + idx} id={"id_scan_size_" + idx} type="number" value={val} innerRef={scanRef} onChange={onScanSizeChange} onKeyDown={onCommaPress} /></Form.Field>
+          })}
+          <Form.Field>
+            <Button onClick={newScanDim} disabled={scanSize.length === maxScan? true: false} type="button" icon="add" title="Add dimension" basic={true} color="blue" />
```
sk1p, line 62: The `? true: false` part is unnecessary - `scanSize.length === maxScan` is already a boolean value! 

If `maxScan` is equal to `minScan`, consider the possibility of hiding the buttons completely

client/src/dataset/components/ScanSize.tsx
AnandBaburajan: > `setScanSize` does not trigger a re-render, which may also be the reason you had to fiddle with the key in the forms

Yes, that is actually the reason. I'll fix this, thanks!

client/src/dataset/components/MIBParamsForm.tsx
```diff
@@ -31,7 +32,22 @@ const MIBFileParamsForm: React.SFC<MergedProps> = ({
...
+    const onReset = () => {
+      handleReset();
+      setScanSizeKey("scan_size_key_empty");
+    }
+
+    React.useEffect(() => {
+      if(values.scan_size==="") {
+        setScanSizeKey("scan_size_key");
+      }
+    }, [values.scan_size]);
```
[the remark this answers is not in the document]
AnandBaburajan: No, and I'll try to fix this in the setState update part.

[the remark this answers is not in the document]
AnandBaburajan: > Why does the key need to change based on the value?

`values.scan_size===""` wasn't correct, sorry.

> Was it not updating correctly without this?

I'm using/updating the keys just for the reset function to work.

AnandBaburajan:
I really appreciate the detailed feedback! :) I've made the changes as suggested.

client/src/dataset/components/ScanSizePart.tsx
```diff
@@ -0,0 +1,23 @@
...
+    scanRef: (ref:HTMLInputElement) => void,
+}
+
+const ScanSizePart: React.FC<ScanSizePartProps> = ({ name, id, scanKey, scanSizeChangeHandle, value, commaPressHandle, scanRef }) => {
+   const onPartChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+      scanSizeChangeHandle(scanKey, e.target.value);
+   }
+   const onCommaPress = (e: React.KeyboardEvent<HTMLInputElement>) => {
+      commaPressHandle(scanKey, e.keyCode);
+   }
```
sk1p, line 19: You might want to move the keycode check here, otherwise the name of the callback, `commaPressHandle`, is a bit misleading, as it behaves like `onKeyDown`

client/src/dataset/helpers.ts
```diff
@@ -8,6 +8,15 @@ export function parseNumList(nums: string) {
     return nums.split(",").filter(part => part.trim() !== "").map(part => +part);
 }
 
+export function parseNumListWithPadding(nums: string, minLength: number) {
+    let initialList = new Array(minLength).fill("");
+    if(nums) {
+      const value = parseNumList(nums);
+      initialList = (minLength - value.length)>0 ? [...value, ...Array(minLength - value.length).fill("")] : [...value];
```
sk1p, line 15: For readability, the ternary operator can be converted to an `if`-statement.

client/src/dataset/components/MIBParamsForm.tsx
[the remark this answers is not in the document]
sk1p: > just for the reset function to work.

Hm - having to fiddle with `key` or similar to force a component to update is most of the time a sign that there is an underlying bug. In this case, I think it is caused by not having a single source of truth for the scan size - you have both the value that comes in via the `value` prop and the local `scanSize` state variable. If you instead remove the local state variable and only use the value from the prop, reset should be handled correctly.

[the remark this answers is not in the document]
AnandBaburajan: > I think it is caused by not having a single source of truth for the scan size.

Yup! That was the reason, thanks again! :)

client/src/dataset/components/ScanSize.tsx
```diff
@@ -0,0 +1,79 @@
...
+}
+
+const ScanSize: React.FC<ScanSizeProps> = ({ value, minScan, maxScan, setFieldValue }) => {
+
+    const [scanSize, setScanSize] = React.useState(parseNumListWithPadding(value, minScan));
+
+    React.useEffect(() => {
+      const newScanSizeValue = parseNumListWithPadding(value, minScan);
+      setScanSize(newScanSizeValue);
+    }, [value, minScan]);
```
sk1p, line 20: You can also completely get rid of the `scanSize` state variable. Then you can just do something like `const scanSize = parseNumList...(value, minScan)`, which will get executed on each re-render, so every time `value` changes, `scanSize` will adapt. then below, you can remove all the `setScanSize` calls and only call `setFieldValue`. What do you think?

client/src/dataset/components/ScanSize.tsx
```diff
@@ -0,0 +1,79 @@
...
+    return (
+      <>
+        <Form.Group>
+          {parseNumListWithPadding(value, minScan).map((val, idx) => {
+            const scanRef = (ref:HTMLInputElement) => { scanRefsArray.current[idx] = ref; }
+            return <Form.Field width={2} key={idx}><ScanSizePart scanKey={idx} name={"scan_size_" + idx} id={"id_scan_size_" + idx} value={val} scanRef={scanRef} scanSizeChangeHandle={scanSizeChangeHandle} commaPressHandle={commaPressHandle} /></Form.Field>
+          })}
+          <Form.Field hidden={minScan === maxScan}>
+            <Button onClick={newScanDim} disabled={scanSize.length === maxScan} type="button" icon="add" title="Add dimension" basic={true} color="blue" />
+            <Button onClick={delScanDim} disabled={scanSize.length === minScan} type="button" icon="minus" title="Remove dimension" basic={true} color="blue" />
```
sk1p, line 72: A comment about the buttons: keyboard focus is very hard to see on `basic={true}` buttons. You can check this by focusing the last dimension and pressing tab. Currently, the shade of the "+" changes veeery sliiiightly.

Even though it might be a tiny bit more ugly, what do you think about using `basic={false}` buttons here with the default grey color? That one makes the keyboard focus obvious.

Thanks for the great work so far, I think this is almost done!

client/src/dataset/components/ScanSize.tsx
AnandBaburajan: > You can also completely get rid of the scanSize state variable.

That did occur to me, but I wasn't sure if the +/- functionality would work without having states.

> Then you can just do something like const scanSize = parseNumList...(value, minScan), which will get executed on each re-render, so every time value changes, scanSize will adapt.

Nice! So a state isn't needed at all, I've taken note of that, thanks!

sk1p: > I wasn't sure if the +/- functionality would work without having states.

Yes, this should work! As long as what you feed into `setFieldValue` round-trips together with your parse function, +/- should work (clicking "+" should result in a trailing comma)

sk1p:
Hm, now travis is also "hanging", at least the synchronization into GitHub. Let me close and re-open the PR to see if at least travis will be fixed by that.

AnandBaburajan:
@sk1p Should I convert RAW's `detector_size` into fixed 2D [  ] [  ] too?

sk1p:
> Should I convert RAW's `detector_size` into fixed 2D [ ] [ ] too?

I think that would be great! Maybe the `ScanSize` should then be renamed to `TupleInput` or similar?

sk1p:
Thanks, merged!

PR #782, opened 2020-05-08
Better test to trigger UnicodeDecodeError

AnandBaburajan:
> The way _mk_random currently generates test data means we get only the values from the set {0.0, 1.0, 288.0, 2880.0} for default_raw. The binary representation of those float values all start with a '\x00' - so the first string value, which is at offset 4, will always be truncated to zero length and trigger the error. If we do change the function to generate values that have a larger fractional value in the binary representation (which we might want to do!), this assumption will no longer hold.

Thanks for the explanation :)

## Contributor Checklist:

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -262,6 +262,24 @@ def uint16_raw(tmpdir_factory):
     yield ds
 
 
+@pytest.fixture(scope='session')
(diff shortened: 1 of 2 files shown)


uellue:
Nice, well done, merging!

uellue:
Refs #775

PR #786, opened 2020-05-14
Shutdown Button

### LINKED ISSUE #698: Button to make the server stop
An example as in Jupyter notebook :

quit

twentyse7en:
Fix #698 
## Contributor Checklist:

changes were made to the GUI)


twentyse7en:
`
/home/abi/anaconda3/lib/python3.7/multiprocessing/semaphore_tracker.py:144: UserWarning: semaphore_tracker: There appear to be 1 leaked semaphores to clean up at shutdown
  len(cache))
`
Above message is displayed 4 times before the shutdown

sk1p:
> Above message is displayed 4 times before the shutdown

I guess you started four worker processes? Does this have any consequences? Are there any processes left over after shutdown is complete?

twentyse7en:
> 
> I guess you started four worker processes? 

Yes

>
>Does this have any consequences? Are there any processes left over after shutdown is complete?

There is also `BrokenPipeError: [Errno 32] Broken pipe`. All processes are terminated.

twentyse7en:
I think closing the executor solved it.

src/libertem/web/shutdown.py
```diff
@@ -0,0 +1,18 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class ShutdownHandler(tornado.web.RequestHandler):
+    def initialize(self, state: SharedState, event_registry):
+        self.state = state
+        self.event_registry = event_registry
+
+    async def get(self):
```
sk1p, line 14: This should possibly be a non-`GET` method, as `GET` is defined as a "safe method", which [SHOULD NOT have the significance of taking an action other than retrieval.](https://tools.ietf.org/html/rfc2616#section-9) (excuse the shouting, that's part of how RFC specifications are written...) `POST` or `DELETE` may be more appropriate. Also, possibly write a response (and wait until it is completely sent) before shutting down the I/O loop. The [flush method](https://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.flush) may be of help here.

Rest of the `RequestHandler` looks good.

src/libertem/web/server.py
```diff
@@ -79,6 +79,12 @@ def make_app(event_registry, shared_state):
             "state": shared_state,
             "event_registry": event_registry
         }),
+        (r"/api/shutdown/", ShutdownHandler, {
+            "state": shared_state,
+            "event_registry": event_registry
+        }
+
+        ),
```
sk1p, line 87: Style nitpick: closing parenthesis can be on the same line as closing curly brace.

client/src/App.tsx
```diff
@@ -6,11 +6,13 @@ import ChannelStatus from './channel/components/ChannelStatus';
 import DatasetList from './dataset/components/DatasetList';
 import ErrorList from './errors/components/ErrorList';
 import logo from './images/LiberTEM logo-medium.png';
+import QuitButton from './quit/components/QuitButton';
 
 class App extends React.Component {
     public render() {
         return (
             <Container style={{ margin: "5em 1em 5em 1em" }}>
+                <QuitButton/>
```
sk1p, line 15: Hm, maybe there can be a menu where we put this button? Possibly we should also move the "about"-action from the litte "i"-icon to that menu, to make it more visible/accessible.

src/libertem/web/shutdown.py
twentyse7en: >
> This should possibly be a non-`GET` method, as `GET` is defined as a "safe method"

Thx for the information.

> Also, possibly write a response (and wait until it is completely sent) before shutting down the I/O loop.

What is the purpose of sending the response? (Is that to ensure we are inside request handler 
and about to shutdown)

client/src/App.tsx
twentyse7en: Yes, so it can be easily accessed even if we scroll to bottom.

src/libertem/web/shutdown.py
sk1p: > What is the purpose of sending the response?

This is to inform the client that the shutdown request was received and will be processed. If you don't send a response, the client will never know - maybe the network connection just dropped before receiving the request, and the server is still running?

client/src/App.tsx
sk1p: Yes, also adding "top-level" user interface elements can clutter up the interface - it is better to create a hierarchy of user interface elements

src/libertem/web/shutdown.py
twentyse7en: Under what circumstance we use `RequestHandler.finish()`

sk1p: After sending the response, to make sure we wait for the response to be sent before the event loop is stopped. Otherwise we may only send a part of the reponse, or nothing at all. [RequestHandler.write](https://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.write) only writes into an output buffer, so it is not guaranteed that the response ends up on the network after only calling `write`.

client/src/App.tsx
sk1p: Ah, something else: maybe pressing "quit" should also [show a modal dialog](https://react.semantic-ui.com/modules/modal/), where the user can confirm that they want to stop the server? Otherwise, mis-clicks could make people unhappy...

src/libertem/web/shutdown.py
```diff
@@ -0,0 +1,23 @@
+import tornado
+import logging
+
+from .state import SharedState
+
+log = logging.getLogger(__name__)
+
+
+class ShutdownHandler(tornado.web.RequestHandler):
```
sk1p, line 9: `ShutdownHandler` looks good now :+1:

client/src/quit/components/QuitButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+            <Modal.Content>
+                <p>
+                    Do you want to exit ?
+                </p>
+            </Modal.Content>
+            <Modal.Actions>
+                <Button color='red' onClick={this.handleClose}>
+                    <Icon name='remove'/> No
+                </Button>
+                <Button color='green' onClick={this.handleQuit} >
```
twentyse7en, line 40: If the user confirms exit by pressing "Yes" button, the request is sent to the server, and the modal closes. But there is delay in stopping the server.

AnandBaburajan: A suggestion: you might like implementing your Modal with React's functional components along with hooks.

sk1p: > there is delay in stopping the server

That's to be expected, as it takes a while for the worker processes to close down.

sk1p, line 40: About button colors and labels: instead of red/green, the convention of the rest of the UI should be followed: the "yes" button should be marked as "primary" (then it will automatically become blue), the secondary action should stay default (grey). That way the buttons all behave consistently. Instead of putting "yes" and "no" as labels of the button, the labels could be "Shutdown" (or "Exit") and "Cancel" - because user's don't like to read, they can just click on the action they want to execute, instead having to read and understand the text above.

I know this may seem very minor, but things like this add up to a more fluid user experience.

twentyse7en: Is there a chance of confusion for the user as he/she press "Yes" and the server is still running?

sk1p: > Is there a chance of confusion for the user as he/she press "Yes" and the server is still running?

What you can also do: after confirming the shutdown, turn the modal dialog into a spinner, like you get when the GUI runs but the server is shutdown. That way, the user will know that something is happening, and they will also know when the executor has finished shutting down.

twentyse7en: Instead of modal dialog, "shutdown" button will turn into loading indicator.

sk1p: Okay, that works, too! I think the button needs to be disabled when a shutdown is running, too, as otherwise, `handleQuit` is called multiple times

client/src/Menu.tsx
```diff
@@ -0,0 +1,29 @@
...
+        <Menu fixed="top">
+            <Menu.Item>
+            <Modal trigger={
+                        <Button content="About" />
+                    }>
+                        <Popup.Header>About LiberTEM</Popup.Header>
+                        <Popup.Content>
+                            <About />
+                        </Popup.Content>
+                    </Modal>
```
sk1p, line 18: There is an indentation mismatch here, should be re-formatted

client/src/App.tsx
```diff
@@ -1,33 +1,27 @@
...
 import ChannelStatus from './channel/components/ChannelStatus';
 import DatasetList from './dataset/components/DatasetList';
 import ErrorList from './errors/components/ErrorList';
 import logo from './images/LiberTEM logo-medium.png';
+import HeaderMenu from './Menu'
 
 class App extends React.Component {
     public render() {
         return (
+            <div>
```
sk1p, line 13: Instead of the `<div></div>`, if you don't need the `<div>` element in the DOM, you can also use `<></>`. Also the contents then need to be re-indented (can possibly be done by re-formatting in vs code?)

client/src/quit/components/QuitButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+                </p>
+            </Modal.Content>
+            <Modal.Actions>
+                <Button onClick={this.handleClose}>
+                    <Icon name='remove'/> Cancel
+                </Button>
+                <Button primary={true} loading={this.state.shutdown} onClick={this.handleQuit} >
+                    <Icon name='checkmark' /> Shutdown
+                </Button>
+            </Modal.Actions>
```
sk1p, line 45: Hmm, maybe the language should be unified: Quit, Shutdown, Exit? Maybe only use "Shutdown"?

client/src/App.tsx
twentyse7en: Thanks for the suggestion, Should have installed a formatter earlier.

twentyse7en:
>
>alternatively show a "shutdown complete" message and also permanently disable the websocket connection

will go with this option, should we follow this behaviour also for stopping server using ctrl+c ?

>
>Did you try the shutdown behavior in the context of jupyterlab? Does the server reliably shut down?

It's restarting as expected. Will try again after disabling WebSocket request.

sk1p:
> should we follow this behaviour also for stopping server using ctrl+c ?

Hm, I don't think so - at least when developing LiberTEM, I often only restart the server, and let the client reconnect as soon as the server is available again. This makes for a smooth experience, which would be interrupted by having to reload the tab otherwise.

> It's restarting as expected. Will try again after disabling WebSocket request.

:+1:

client/src/shutdown/components/ShutdownButton.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+        if (channel === "disconnected" && shutdown) {
+            this.setState({ shutdown: false });
+            this.modalClose();
+        }
+    }
+
+    public render() {
+
+        this.handleModal();
```
twentyse7en, line 46: is it okay to use function inside `render` ?

twentyse7en:
Stopping the WebSocket request after the shutdown, reloading the page responded with `500 : Internal Server Error`  in Jupyter Lab

client/src/shutdown/components/ShutdownButton.tsx
sk1p: Good call - I think calling `setState` in `render` can be a problem, but doesn't have to be (one needs to be careful not to create infinite loops, as calling `setState` will result in a re-render).

src/libertem/web/shutdown.py
sk1p: Btw. it would be good to have test coverage for `ShutdownHandler`, just to make sure the code keeps working across other changes, for example in tornado, or if the structure of the state changes - see https://github.com/LiberTEM/LiberTEM/tree/master/tests/server for some tests that exercise the HTTP API.

sk1p:
> Stopping the WebSocket request after the shutdown, reloading the page responded with `500 : Internal Server Error` in Jupyter Lab

Interesting. Anything logged on the console of jupyterlab?

I'll have a more detailed look on Tuesday, or if possible, tomorrow.

client/src/shutdown/components/ShutdownButton.tsx
```diff
@@ -0,0 +1,68 @@
...
+    public modalClose = () => {
+        this.setState({ modal: false });
+    };
+
+    public handleShutdown = () => {
+        const timestamp = Date.now();
+        this.setState({ shutdown: true });
+        this.props.shutdownAction(timestamp)
+        handleSubmit();
+    };
```
sk1p, line 39: I gave it a spin - one thing I noticed was that the "currently shutting down" state was never displayed, even if the shutdown took a while. I think there are two things to fix: 1) in this function, the `shutdownAction` should only be dispatched once the answer from the HTTP API is received, and 2) only send the `SERVER_SHUTDOWN` message once the executor has finished closing. You can test that this works by adding an artificial `await asyncio.sleep(some_seconds)` to your `ShutdownHandler`, before sending the response.

For 1), a hint: as the `handleShuwdown`method here is not defined as `async`, you will have to use the Promise/callback API here, so something like this:

```typescript
handleSubmit().then((response) => { this.props.shutdownAction(...); });
```

(NB: `handleSubmit` could be named better - maybe `doShutdown` or similar? Something that indicates that this shuts down the server - `handleSubmit` sounds more like something that handles a form submission to me)

sk1p: Btw. the `CHANNEL_SHUTDOWN` action is dispatched two times - once here, and once from the saga - is that intentional? Maybe it should only be dispatched in the saga? Or only here?

twentyse7en: > is that intentional? Maybe it should only be dispatched in the saga? Or only here?

Yes, It's only temporary. Will be removed from saga after some other tweaks.

twentyse7en:
> Interesting. Anything logged on the console of jupyterlab?

https://pastebin.com/embed_js/bTWD7Mrv

sk1p:
> > Interesting. Anything logged on the console of jupyterlab?
> 
> https://pastebin.com/embed_js/bTWD7Mrv

Okay, so Jupyterlab doesn't expect the server to suddenly disappear - maybe it is somehow possible to "tell" jupyterlab that we are shutting down? As this is unrelated to the basic shutdown button functionality, I would say this can wait until the PR is merged, then we can see if/how we can better integrate into jupyterlab, if you are interested.

twentyse7en:
When the server responds, `closeLoopAction` is dispatched to stop further WebSocket requests. `close` is dispatched when the server stops, then `shutdownAction` to display the message.

client/src/App.tsx
sk1p: Just curious, which formatter are you using? That's something different from the built-in TypeScript formatting of vs code, right?

Also, it would be helpful to only format the newly added/changed code (there should be a setting "format on save" or similar, combined with manually applying "Format Selection"), otherwise, when reviewing, I always have to check if a given change is only a formatting change, or part of the "real" changes.

(sadly, automatically formatting only changed lines is not yet supported in vs code, https://github.com/microsoft/vscode/issues/44075)

twentyse7en: > Just curious, which formatter are you using? That's something different from the built-in TypeScript formatting of vs code, right?

Prettier.Sorry for the trouble, no way to format just the changed lines.(https://github.com/prettier/prettier-vscode/issues/630).

I had no idea about "Format Selection", I will only use that from now on. Thanks.

twentyse7en:
> * After the server is shut down, if the user doesn't close the tab as requested but instead clicks on shutdown again 

Should we improve the shutdown message to force these guys to close the tab :sweat_smile: .Will disable the button.

twentyse7en:
need some help on `test_shutdown.py`, (https://pastebin.com/embed_js/0ELwk8E6)

sk1p:
> Should we improve the shutdown message to force these guys to close the tab 

Haha, no I think the message is fine as-is, just never underestimate the ability of people to not read what is in front of them :grinning: 

> need some help on test_shutdown.py

I'll have a look

tests/server/test_shutdown.py
```diff
@@ -0,0 +1,19 @@
...
+async def test_shutdown(base_url, http_client, server_port):
+    await create_connection(base_url, http_client)
+
+    print("checkpoint 1")
+
+    url = f"ws://127.0.0.1:{server_port}/api/shutdown/"
+    async with http_client.delete(url) as resp:
+        print("checkpoint 2")
+        assert resp.status == 200
+        assert_msg(await resp.json(), "SERVER_SHUTDOWN")
```
sk1p, line 19: There are two parts to [the errors](https://pastebin.com/raw/0ELwk8E6) you are getting:

1) the "ERROR at teardown of test_shutdown": "RuntimeError: cannot schedule new futures after shutdown" - this happening because of the `server_port` fixture - that fixture starts a libertem server in the background, which is stopped again after usage. But as you are already shutting down the server, the second try fails when it tries to `close` the executor.

2) the `http_client`, which is based on `aiohttp`, fails to read the response with a `aiohttp.client_exceptions.ServerDisconnectedError`


I think 1) can be fixed by not stopping the server in [the fixture](https://github.com/LiberTEM/LiberTEM/blob/048c38d0cce687dc16269ff71987ec3e7cfafdba/tests/conftest.py#L66). This can be done even without modifying the fixture code - as you can see, the executor is only closed if there is one in the corresponding state variable - you could set it to `None` in your `ShutdownHandler`.

I'm not sure yet how to fix 2) - maybe in addition to `flush`, you need to call [`finish`](https://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.finish)? I'm not sure this is enough.

src/libertem/web/shutdown.py
```diff
@@ -0,0 +1,25 @@
...
+    async def delete(self):
+        log.info("Handling shutdown button")
+        if self.state.executor_state.executor is not None:
+            await self.state.executor_state.executor.close()
+            self.state.executor_state.executor = None
+        await self.finish({
+            "status": "ok",
+            "messageType": "SERVER_SHUTDOWN"
+        })
+        await asyncio.sleep(1)
```
twentyse7en, line 24: `finish` doesn't fix the issue. Now the test passes with some warning.

sk1p: Hmm, I'll have a look later. Perhaps we have to use a different method of testing for this, as I would prefer not to have a sleep there.

twentyse7en: any hint?

sk1p: Sorry, didn't find time yet. Will have a look later today

tests/server/test_shutdown.py
```diff
@@ -0,0 +1,19 @@
+import pytest
+
+from utils import assert_msg
+from aio_utils import create_connection
+
+pytestmark = [pytest.mark.functional, pytest.mark.flaky]
```
sk1p, line 6: Can you remove the `flaky` mark here? Otherwise, the test won't be run in CI

sk1p:
Thanks, merged!

PR #787, opened 2020-05-14
Add COM curl and fix divergence

uellue:
This has to be verified with real data

## Contributor Checklist:


sk1p:
> This has to be verified with real data

Once that is done, this can be merged I think.

PR #797, opened 2020-06-02
Run LiberTEM test server in a separate event loop

sk1p:
This is needed to properly isolate event loops, and allows to test edge cases, for example shutting down the server.

## Contributor Checklist:

changes were made to the GUI)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -109,12 +109,10 @@ class ServerThread(threading.Thread):
 
 
 @pytest.fixture(scope="function")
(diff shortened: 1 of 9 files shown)


sk1p:
I'm also re-enabling the flaky tests to see how they react - with the new solution, it is possibile that we will get proper error messages if the server decides to hang at shutdown.

They are quite slow, I think this is because of the numba non-caching, which is amplified here, as each new server is run on a new thread, so it is possible that they don't share the numba compilation results. Needs to be verified...

PR #801, opened 2020-06-08
Prototype for notebook generator

twentyse7en:
## Contributor Checklist:

changes were made to the GUI)


uellue:
Thx for the PR! :-)

About your previous question about visualization, plotting the data instead of the visualization (I believe it was `.raw_data`) will in most cases be more user-friendly since it makes it easier to adjust limits, color scale etc. However, we have some complex analyses where `imshow` will fail, for example because the result is a vector field or complex numbers.

As a pragmatic approach, one could just include both and skip plotting the raw data in case it won't work:

```
# Pseudo code
if len(data.shape) != 2:
   print("This data doesn't work directly with imshow since it is not 2D, requires a custom visualization")
elif is_complex(data):
    plt.imshow(np.abs(data))
else:
    plt.imshow(data)
```

uellue:
As a quick question, how can I run the code? Can you perhaps provide a minimal example?

twentyse7en:
Please change the parameters in `notebook_genertor.py` and run it.

twentyse7en:
What is the difference between `libertem.udf.sum.SumUDF` and `create_sum_analysis` ?

uellue:
> Please change the parameters in `notebook_genertor.py` and run it.

Worked out of the box, nice! The notebook looks pretty good already. :-)

uellue:
> What is the difference between `libertem.udf.sum.SumUDF` and `create_sum_analysis` ?

The UDF is the "workhorse" that does the numerical processing, and the analysis deals with interfacing, mostly with the web GUI. The UDF returns raw results, while the analysis also deals with postprocessing and visualization.

We have an open issue to include essential postprocessing in the UDF. The `StdDevUDF` is a prime example where the result buffers of the UDF are not the final result. See #628 for the corresponding issue.

uellue:
About the structure of the code: Probably one should allow per-analysis customization of the output not centrally, but in a template that is linked to a specific Analysis sub-class. That would also allow working around the plotting method issue discussed above: The Analysis can "know" how to plot the data properly.

twentyse7en:
> About the structure of the code: Probably one should allow per-analysis customization of the output not centrally, but in a template that is linked to a specific Analysis sub-class. That would also allow working around the plotting method issue discussed above: The Analysis can "know" how to plot the data properly.

Yep, that would be more convenient. I'm working on it. Thx for the feedback :)

src/libertem/analysis/apply_fft_mask.py
```diff
@@ -6,6 +6,12 @@
 class ApplyFFTMask(BaseAnalysis):
     TYPE = "UDF"
 
+    @classmethod
+    def get_analysis_info(cls):
+        return {
+            "type": "APPLY_FFT_MASK",
+        }
+
```
uellue, line 14: Where is the return value of this used? I can only find a deleted code portion.

src/libertem/analysis/base.py
```diff
@@ -154,7 +154,26 @@ def __iter__(self):
...
+        cls.registry[analysis_info['type']] = {
+            "class": cls,
+        }
+
+    @classmethod
+    def get_analysis_by_type(cls, type_):
+        return cls.registry[type_]
+
+
+class Analysis(metaclass=AnalysisRegistry):
```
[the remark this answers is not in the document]
uellue: ...as soon as it is clear what this does and why, we should discuss if there is maybe a way to achieve the same goal with more common programming patterns, CC @sk1p .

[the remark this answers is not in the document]
sk1p: The alternative is not as elegant, that would be explicit registration (`Aanlysis.register('SOME_NAME', SomeAnalysisClass)`), or a central list of all available analyses, like we have for the `DataSet`s. This registry pattern is actually one of the most common metaclass uses, and is often used as an example in tutorials and books. I also use it for the registry of download formats, for example.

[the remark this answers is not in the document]
uellue: Ok, in that case a comment pointing to a concise explanation would be good! :-)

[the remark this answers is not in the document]
sk1p: There is actually a new feature available since Python 3.6, which is the `__init_subclass__` method. That can be used as a shorter alternative, and we can skip the metaclass explanations: https://stackoverflow.com/a/50099920/540644

Still not a very well known feature, but a bit less code needed, and possibly more straight-forward to understand. If we decide on that version, the download format registry should also be changed, so we have some consistency.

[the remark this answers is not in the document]
uellue: > (...) `__init_subclass__` method (...)

Oh, *that* looks neat and straight to the point. Much more intuitive. :+1:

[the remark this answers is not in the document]
twentyse7en: So I will change it :)

src/libertem/analysis/base.py
```diff
@@ -172,6 +172,22 @@ class Analysis:
     # https://www.python.org/dev/peps/pep-0586/
     TYPE: typing.Union[str, None] = None
 
+    registry = {}
+
+    def __init_subclass__(cls, type_=None, **kwargs):
```
sk1p, line 177: Hmm, can we rename `type_`? Like this, there are two "types" for an analysis, this one and the UDF/JOB one. Maybe `id_` or `name` also fits?

src/libertem/analysis/base.py
```diff
@@ -154,7 +154,7 @@ def __iter__(self):
         return iter(self.results)
 
 
-class Analysis:
+class Analysis():
```
sk1p, line 157: Empty parens can be removed here

twentyse7en:
> What do you think?
Sounds good :+1:

prototypes/notebook-generator/code_template.py
```diff
@@ -0,0 +1,93 @@
...
+            helper = helperCls()
+
+            analysis_api = helper.api
+            short = helper.short_name
+
+            plot_ = helper.get_plot()
+            docs_ = self.format_template(temp_docs, {'analysis': an_type})
+
+            if 'roi' in an_params.keys():
+                data = {'roi_params': an_params}
```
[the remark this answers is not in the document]
twentyse7en: It's the parameters for `get_roi()`, example [notebook](https://github.com/LiberTEM/LiberTEM/blob/01f3569f9166c60550f1a20ddb32a2e2cb890dfe/prototypes/notebook-generator/comp_analysis_notebook.ipynb).

src/libertem/analysis/base.py
twentyse7en: will go with `id_`

prototypes/notebook-generator/code_template.py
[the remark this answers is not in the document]
sk1p: Ok, but if the analysis has more parameters than just the `roi`, how is that handled? I think this may be the case for one of the FFT analyses? `SumfftAnalysis`?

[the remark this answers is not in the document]
twentyse7en: does the `sumfftAnalysis` have `roi` as parameter?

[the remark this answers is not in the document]
sk1p: Yes, as it is a subclass of the `SumAnalysis`. IIRC all analyses that have a `sig`-shaped result support ROIs, and you can check by using the ROI dropdown in the GUI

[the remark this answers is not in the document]
twentyse7en: fft_analysis
ROI dropdown is missing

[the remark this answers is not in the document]
sk1p: Ah, so it was never added to the FFT analysis in the GUI. But you can use your imagination: 

image

These are the analysis parameters; if there was a `roi`, there would also be a `roi` key under parameters, in addition to the `real_*` ones here:

image

twentyse7en:
can I get an example notebook for cluster analysis?

src/libertem/analysis/radialfourier.py
```diff
@@ -9,11 +9,32 @@
...
+        self.params = params
+
+    def convert_params(self):
+        params = [f'dataset=ds']
+        for k in ['cx', 'cy', 'ri', 'ro', 'n_bins', 'max_order']:
+            params.append(f'{k}={self.params[k]}')
+        return ', '.join(params)
+
+    def get_plot(self):
+        # wrong visualization
```
twentyse7en, line 32: need help for plotting :)

sk1p:
> can I get an example notebook for cluster analysis?

Maybe @uellue has an example notebook already, but you can also have a closer look at the analysis class - I think in the `controller` method is some code that could be used for a notebook template.

src/libertem/analysis/radialfourier.py
twentyse7en: rad_to_upload

Which one should I choose `ctx.run()` or `ctx.run_udf()`

twentyse7en:
Now notebook can be generated from the GUI. It's just a button for now, will be replaced later.

uellue:
> can I get an example notebook for cluster analysis?

Here is our example for clustering: https://libertem.github.io/LiberTEM/app/pcmclustering.html

src/libertem/analysis/radialfourier.py
uellue: The UDF is just an `ApplyMasksUDF`. The "magic" is in creating a suitable sparse mask stack and interpreting the result. For that reason I'd first go with the Analysis in the notebook and maybe adding a comment to have a look at `RadialFourierAnalysis.get_mask_factories()` and `RadialFourierAnalysis.get_results()` if someone is interested in the inner workings of this analysis.

sk1p:
> When trying out your branch locally I didn't find the download link in the GUI. It is supposed to be an additional format in "Download", right? At least that's how I interpreted the code...

Did you run the development version of the client? `npm run start` in `client`...

uellue:
> Did you run the development version of the client? `npm run start` in `client`...

Riiiiight.... 🤦‍♂️  Thx!

uellue:
Now it worked! In general, it works really nicely, congratulations! For the general interface and code structure I don't have any comments, looking good to me. @sk1p what is your take?

From my perspective, the next steps could be a detailed test and review of all the analyses with different file types etc. Probably that would make sense as soon as the detailed items from above are addressed?

For the descriptions you can probably use the docstrings of the various methods of `Context()` that create analyses, the Analysis classes and the UDFs.

If we make sure that we have good docstrings at the appropriate places -- which we should do anyway -- we can even use the docstrings directly using introspection at run time: https://stackoverflow.com/questions/10037825/python-introspection-access-function-name-and-docstring-inside-function-definit

Since the `Analysis` classes are not public and therefore don't always have well-maintained docstrings, we can probably decide on a case-by-case basis where we pull the docstring of a particular analysis from.

One point to consider is that notebooks use Markdown, while docstrings are RST. We can just use the RST as plain text or find a converter.

src/libertem/analysis/base.py
```diff
@@ -249,6 +266,29 @@ def get_parameters(self, parameters):
         """
         raise NotImplementedError()
 
+    @classmethod
+    def get_analysis_by_type(cls, type_):
```
sk1p, line 270: Why is this needed? Shouldn't the registry be used for this?

src/libertem/analysis/com.py
```diff
@@ -7,11 +7,38 @@
...
+        params = [f'dataset=ds']
+        for k in ['cx', 'cy']:
+            params.append(f'{k}={self.params[k]}')
+        params.append(f"mask_radius={self.params['r']}")
+        return ', '.join(params)
+
+    def get_plot(self):
+        # any other better way to plot ?
+        plot = ["plt.figure()",
+                "plt.imshow(com_result['intensity'].data / com_result['intensity'].data.max())"]
```
sk1p, line 38: Is this division necessary? `imshow` should already normalize, right? For center of mass, there are a few more interesting plots you can do for the results, like `curl`, `divergence` and a field visualization; possibly could add some of them here, too.

src/libertem/analysis/base.py
twentyse7en: Oops! Download feature was a different branch merged into this. I must have done something wrong while merging.

src/libertem/analysis/com.py
twentyse7en: no_image

>
>possibly could add some of them here, too.

Okay, then I have to switch to `ctx.run()`

sk1p: > then I have to switch to ctx.run()

Or use the appropriate post-processing functions

sk1p: Regarding the plot, what shape does `com_result['intensity']` have? And what dtype?

src/libertem/analysis/base.py
sk1p: Ah, yeah, that happens :)

src/libertem/web/notebook_generator/.gitignore
```diff
@@ -0,0 +1,3 @@
+dask-worker-space/
+__pycache__/
+.ipynb_checkpoints/
```
sk1p, line 3: These should all be already ignored by the main gitignore, right? This probably comes from the move from your own repository to here

src/libertem/analysis/sumsig.py
```diff
@@ -1,9 +1,36 @@
...
+
+    def __init__(self, params):
+        self.params = params
+
+    def get_dependency(self):
+        return "from libertem.analysis import SumSigAnalysis"
+
+    def get_docs(self):
+        docs = ["# SumSig Analysis",
+                "***about sumsig analysis ***"]
```
sk1p, line 19: If you are adding helpful descriptions for each analysis, maybe it would be good to also display those in the GUI, maybe available as an "about analysis" button as very short online help? That would be very nice!

sk1p:
As a more general question of organization, once the notebook generation prototype is useful, should we get it into a mergable state already? I think it would be helpful to split the whole project into logical parts, so this PR doesn't keep growing :grin:
Target would then be to merge this after the 0.6 release

src/libertem/web/notebook_generator/.gitignore
twentyse7en: Yes, there is no need for this `.gitignore`

src/libertem/analysis/com.py
twentyse7en: `<BufferWrapper kind=nav dtype=float32 extra_shape=(3,)>`

prototypes/notebook-generator/helper.py
```diff
@@ -0,0 +1,58 @@
...
+
+class GeneratorHelper:
+
+    short_name = None
+    api = None
+
+    def __init__(self, params):
+        self.params = params
+
+    # common in here and code_template
```
twentyse7en, line 12: This function is repeated, what would be the practice.

src/libertem/analysis/sumsig.py
twentyse7en: I was trying to get docs as @uellue suggested, from `Context`. I will add descriptions for analysis which doesn't have right now.

sk1p: That's good! I mean where the description comes from doesn't really matter - what I'm saying is that it would be very cool to **also** display it in the GUI, if we already have it available for generating the notebooks

src/libertem/analysis/com.py
sk1p: Okay, that means you have something like `(256, 256, 3)` (assuming a signal shape of `(256, 256)` - if you just put that unchanged into `imshow`, it will be interpreted as RGB data, which is not correct (and the reason for the strange error message). There needs to be some post-processing code here!

sk1p: > There needs to be some post-processing code here!

... or work towards the `get_results` UDF method: #628

twentyse7en:
> > can I get an example notebook for cluster analysis?
> 
> Here is our example for clustering: https://libertem.github.io/LiberTEM/app/pcmclustering.html

I was asking about `libertem.analysis.clust.ClusterAnalysis`

uellue:
> I was asking about `libertem.analysis.clust.ClusterAnalysis`

That analysis is not exposed through a method in `Context`. It is a multistep Analysis, which works a bit differently than the normal Analysis classes. The source code of the Analysis looks as if it won't play well with notebooks, i.e. it would need some wrappers.

Probably one would have to re-implement the workflow in `controller()` and `get_udf_results()`  in a way that works in a notebook, i.e. similar to the clustering notebook. Possibly the Analysis can be refactored in such a way that relevant code portions are separated out and can be [accessed through introspection](https://stackoverflow.com/questions/427453/how-can-i-get-the-source-code-of-a-python-function) in order to avoid duplication?

prototypes/notebook-generator/helper.py
sk1p: One possibility would be to make `template` a class instead of a list of strings, and put `format` on it as a method. Then it will be available everywhere you have a template instance.

Or you could extract it as a module-level function you import from wherever you need it

twentyse7en:
Can I have some help regarding tests, I'm confused which parts should I consider for testing.

twentyse7en:
Also verify the docs for analysis, RST is converted to Markdown. 
where can I find the docs for `FFT analysis`, `CLUSTER analysis`, `FFT analysis`

twentyse7en:
> As a more general question of organization, once the notebook generation prototype is useful, should we get it into a mergable state already? I think it would be helpful to split the whole project into logical parts, so this PR doesn't keep growing 
> Target would then be to merge this after the 0.6 release

Right now, notebook can be generated corresponding to all compound analysis. I didn't get "mergable state"?

uellue:
> Can I have some help regarding tests, I'm confused which parts should I consider for testing.

Generally, it is a good idea to cover as much of your code and parameter space as reasonably possible with tests. In that case, generating a notebook for each of the supported analyses with default parameters would be a good start, to make sure it doesn't outright fail.

As a next step, one can make sure that the generated notebook runs without errors. One can use one of the files that are available in CI, possibly one of the HDF5 fixtures? Perhaps [runipy](https://deeplearning.lipingyang.org/2018/03/28/run-jupyter-notebook-script-from-terminal/) can work to confirm that a notebook runs?

uellue:
Regarding using `pypandoc` for converting: ["Pypandoc uses pandoc, so it needs an available installation of pandoc."](https://pypi.org/project/pypandoc/). Since Pandoc can't be installed via pip, it means that users wouldn't be able to install via `pip install libertem`. For that reason we should try to find a solution that doesn't depend on pandoc.

src/libertem/analysis/helper.py
```diff
@@ -0,0 +1,112 @@
...
+)
+
+udf = ApplyMasksUDF(
+    mask_factories=lambda: mask, mask_count=len(y), mask_dtype=np.uint8,
+    use_sparse=True
+)
+
+udf_results = ctx.run_udf(dataset=ds, udf=udf, progress=True)
+cluster_result = cluster_analysis.get_udf_results(udf_results=udf_results, roi=roi)
+"""
```
uellue, line 112: Why is this here? It looks like something specific to the cluster analysis? In that case it should be close to there and not in a general part of the code.

Generally, as already mentioned above, would it perhaps be possible to refactor this "temp_cluster_controller" and the actual controller in clust.py in such a way that they share most of the code and that the source code is pulled from the actual implementation by introspection instead of duplicating it?

uellue:
Ah, "working well" doesn't mean that it some way somehow generates working results, but that it is production-ready and doesn't create debt in the future, such as code duplication. :-)

src/libertem/analysis/clust.py
```diff
@@ -13,9 +14,58 @@
...
+    def get_dependency(self):
+        dep = ["from libertem.analysis import ClusterAnalysis",
+               "from libertem.udf.stddev import StdDevUDF, consolidate_result",
+               "from libertem.masks import _make_circular_mask",
+               "from skimage.feature import peak_local_max",
+               "import sparse",
+               "from libertem.udf.masks import ApplyMasksUDF"]
+        return dep
+
+    def temp_cluster_controller(self):
```
twentyse7en, line 36: Is this what you are suggesting @uellue , we have to care about template when changing the `controller`.

twentyse7en:
> It looks like the normal mask-based Analyses work quite nicely already? The cluster analysis and the FFT analysis seem a bit more problematic.

I will go with completing the analysis which are working well : ) Can you please specify the problem with `FFT analysis`.

>Just FYI, just today I worked with a new LiberTEM user and the notebook download would have been useful. :-)

Good to know, I better speed up my work.

src/libertem/analysis/clust.py
uellue: That's already an improvement since it is more likely to break if `controller()` is changed without synchronizing with `temp_cluster_controller()` as soon as the generated notebooks are run in CI.

Even better would be to refactor `controller()`. The text replacement is already a good start.

Possible ideas:

* Move as much functionality as reasonably possible from `controller()` to individual helper methods of the analysis.
  * Instantiating UDFs
  * Interpreting parameters
  * Interpreting UDF results
* Parametrize aspects that are different, such as `self` vs. a variable for the analysis object
* Turn `controller()` into a wrapper that mostly calls a function that looks similar to the current `temp_cluster_controller()` output, i.e. with `self` as a parameter for the analysis, same for dataset etc.

The main points where the controller and the notebook script contents are inherently different is the method for running UDFs: sync vs. async. It seems that the analysis doesn't do too much interesting stuff here, apparently just running the UDF and getting the result in the end. @sk1p can that perhaps be parametrized or abstracted away somehow, ideally that there is just a `ctx.run_udf(...)` in the source code and we somehow mock what `ctx` is in this case to that it works when we call from `controller()` in the async case?

uellue:
> Can you please specify the problem with `FFT analysis`

Just that [you had a question about the docs](https://github.com/LiberTEM/LiberTEM/pull/801#issuecomment-650572176), I thought you had issues with it. If it works otherwise and you just needed the text for the docstring, good! Perhaps just leave the docstring out and put a FIXME in the code for now since that can easily be added in the next iteration?

uellue:
> Good to know, I better speed up my work.

No need to hurry beyond a normal work pace, just wanted to let you know that the feature will be really useful. :-)

src/libertem/analysis/clust.py
uellue: Anyway, that's a bit fiddly and better left for the second iteration. From my gut feeling this will be solvable in the current framework of your code one way or another, which is the main point: Changes here will not affect the notebook implementation in the other analyses.

twentyse7en:
> In that case, generating a notebook for each of the supported analyses with default parameters would be a good start, to make sure it doesn't outright fail.

Instead of generating notebook, input into `nbformat` is tested.

twentyse7en:
I couldn't find a solution for converting RST to MD. But I got something else, converting RST to HTML then displaying the HTML in the [notebook](https://github.com/LiberTEM/LiberTEM/blob/7b3b78039791345e824fe82a1532235872953dc6/prototypes/notebook-generator/notebook/RSTtoHTML.ipynb). I would be better to take care of adding docs in next iteration.

tests/template/test_com_template.py
```diff
@@ -0,0 +1,32 @@
...
+fig, axes = plt.subplots()
+axes.set_title("field")
+axes.imshow(com_result.field.visualized)
+fig, axes = plt.subplots()
+axes.set_title("magnitude")
+axes.imshow(com_result.magnitude.visualized)
+fig, axes = plt.subplots()
+axes.set_title("curl")
+axes.imshow(com_result.curl.visualized)
+'''
```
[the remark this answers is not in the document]
twentyse7en: > For that reason it is essential to actually run the generated notebooks and verify that they pass without errors.

Thank you, Now I have a better understanding of the testing.

>Did you have a look at https://deeplearning.lipingyang.org/2018/03/28/run-jupyter-notebook-script-from-terminal/ already?

I think `nbconvert` would be good :+1:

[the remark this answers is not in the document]
twentyse7en: > If possible, tests should also verify that the notebook generates the intended result

For implementing this, as you suggested an extra cell for saving the result would be easy. `nbconvert` doesn't offer a way to access the result from notebook. If there is no other problem I will add this in the next iteration.

twentyse7en:
Tests are working fine locally in my system, can you please check @uellue.

tests/template/test_com_template.py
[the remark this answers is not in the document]
uellue: > `nbconvert` doesn't offer a way to access the result from notebook.

In the test environment you can use the fixture `tmpdir_factory` to generate temporary directories. In `conftest.py` you can see it in action, for example. One can then generate a directory from that fixture and add a cell to the notebook that explicitly saves relevant results to files in that directory, for example using [numpy.save()](https://numpy.org/doc/stable/reference/generated/numpy.save.html#numpy.save). In the test routine you can then just load the file contents from that location and compare them with the result of the analysis. That way you don't have to access variables within the notebook.

Since that is (hopefully) not too difficult and at the same time very important to verify the correctness of the generated notebooks, perhaps you can give it a try for this iteration?

uellue:
> Tests are working fine locally in my system

Running the tests in a fresh environment using `tox -e py37 tests/template` reproduces the failures on my system. I tried adding `ipykernel` to `test_requirements.txt`, but that didn't help. More ideas that you could try:

* Add `jupyter` to `test_requirements.txt`
* To add a new kernel, as you suggested, you can add an instruction to explicitly install a kernel in `tox.ini`, `[testenv]` environemnt before the `pytest` commands. `python -m ipykernel install --user` should hopefully do it, otherwise you can have a look at the [IPykernel documentation](https://ipython.readthedocs.io/en/latest/install/kernel_install.html).

uellue:
> tox -e py37 tests/template

...ah, another detail: After you make changes to the environment specification (changes to `test_requirement.txt`, for example), you have to explicitly recreate the environment by adding the `-r` option: `tox -r -e py37 tests/template`

uellue:
Thank you for the update, looks good! I can have a closer look on Monday. The error in Appveyor seems to be unrelated. A check in Travis still complains about something in the TypeScript code. Is that related to your changes?

tests/template/test_com_template.py
```diff
@@ -0,0 +1,66 @@
...
+        assert np.allclose(results["field"], expected["field"].raw_data)
+        assert np.allclose(results["magnitude"], expected["magnitude"].raw_data)
+        assert np.allclose(results["divergence"], expected["divergence"].raw_data)
+        assert np.allclose(results["curl"], expected["curl"].raw_data)
+        assert np.allclose(results["x"], expected["x"].raw_data)
+        assert np.allclose(results["y"], expected["y"].raw_data)
+
+    except CellExecutionError:
+        out = None
+    assert out is not None
```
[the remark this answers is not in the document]
twentyse7en: Yes, No need for the exception.

twentyse7en:
> * We need a better way of translating from GUI ROI parameters to notebook ROIs. The usual way of doing ROIs in notebooks is `roi = np.zeros(ds.shape.nav, dtype=bool); roi[...] = True` with a suitable selector (slices or fancy indexing) for the entries that should be `True`

Slicing is challenging, especially the code corresponding to the [`rectangular`](https://github.com/LiberTEM/LiberTEM/blob/1b372535e303a5fa9ae715033ec5f151626cc099/src/libertem/masks.py#L359).

> The Radial Fourier Analysis visualization should include plotting phase and complex result, each with the proper color scales

I don't know how to plot complex result. Also couldn't find proper color scales.

prototypes/notebook-generator/helper.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def get_plot(self):
+        return None
+
+    def get_docs(self):
+        return None
+
+    def get_roi_code(self):
+
+        if 'roi' in self.params.keys():
```
uellue, line 31: Instead of using the internal `get_roi()` method you can use `masks.circular()` or `masks.rectangular()` to draw the masks. Those two methods are included in the reference, so you can import and use them. At the time you are generating the notebook you know if it is a rectangle or a circle, so you can include the appropriate code.

twentyse7en:
can notebook include `CMAP_CIRCULAR_DEFAULT.rgb_from_vector` and `cmaps`, I couldn't find those on reference.

sk1p:
> can notebook include `CMAP_CIRCULAR_DEFAULT.rgb_from_vector` and `cmaps`, I couldn't find those on reference.

I see two options here:

1) Add `libertem.viz` to the documentation, or
2) Add a dependency to `empyre` and remove most of `libertem.viz`, which was duplicated from that package anyways. Then, also directly use `empyre.vis.colors` in the generated notebook code.

I would prefer option 2, as option 1 would add more duplication, instead of removing it (see [the empyre docs](https://empyre.iffgit.fz-juelich.de/empyre/vis.html#module-empyre.vis.colors)). It would also mean we can do some easy improvements, like [plotting a colorwheel](https://empyre.iffgit.fz-juelich.de/empyre/vis.html#empyre.vis.decorators.colorwheel) in the notebook code.

uellue:
> I would prefer option 2

Sounds reasonable! After a quick check, `empyre` in its base version doesn't introduce new secondary dependencies to LiberTEM and isn't too big.

Should we create a separate PR to swap out the functionality of `libertem.viz` and use `empyre` instead? I could take care of that now.

sk1p:
> Should we create a separate PR to swap out the functionality of `libertem.viz` and use `empyre` instead? I could take care of that now.

Sounds good, go ahead!

twentyse7en:
> (see [the empyre docs](https://empyre.iffgit.fz-juelich.de/empyre/vis.html#module-empyre.vis.colors))

I couldn't access the `empyre` link.

sk1p:
> I couldn't access the `empyre` link. `404 The page you're looking for could not be found`.

Ah, the repository (and with it, the documentation) is not yet public - I'll contact the author and see if I can get it fixed!

uellue:
@twentyse7en #851 replaces much of `viz.py` with imports from `empyre`. The notebooks can import and instantiate `ColormapCubehelix` from  `empyre.vis.colors`. Then one can call `rgb_from_vector()` from that object to get the visualization as RGB values, and `plt.imshow()` that.

src/libertem/analysis/com.py
```diff
@@ -1,17 +1,66 @@
...
+            params.append(f'{k}={self.params[k]}')
+        params.append(f"mask_radius={self.params['r']}")
+        return ', '.join(params)
+
+    def get_plot(self):
+        plot = []
+        for channel in self.channels[:3]:
+            plot.append("fig, axes = plt.subplots()")
+            plot.append(f'axes.set_title("{channel}")')
+            plot.append(f'axes.imshow(com_result.{channel}.visualized)')
```
uellue, line 52: This is the last remaining instance that should plot the data instead of the visualized result. :-)

For the field you can use the Cubehelix plot from èmpyre` like for Radial Fourier Analysis, and the two other ones are just `imshow()`.

twentyse7en:
> All in all, congratulations! The first iteration seems in the bag and turned out really well.

Thx :smile:, It took more time than I expected. Thank you very much for your support @uellue and @sk1p.

uellue:
👍 👍 👍  ready to go from my side! @sk1p what is your take?

uellue:
> It took more time than I expected.

Yes, it's always like this! 🙈 But "Gut Ding will Weile haben". Thx for all your work and patience!

sk1p:
I'll have a last look tomorrow

tox.ini
```diff
@@ -3,6 +3,7 @@ envlist = flake8, py{36,37}, benchmark, benchmark-cuda{101,102}, mypy
 
 [testenv]
 commands=
+    python -m ipykernel install --user --name libertem-kernel
```
sk1p, line 6: Using `--user` here means running the tests will change the user's ipython configuration, and may conflict with an existing ipython kernel called `libertem-kernel` (it creates a directory under `~/.local/share/jupyter/kernels`). Is it maybe possible to use `ipykernel install --sys-prefix` instead? That would only modify the tox environment and put the kernel configuration into `$VIRTUAL_ENV/share/jupyter/kernels`, which would not conflict with the user's configuration.

Running the tests outside of tox (with plain pytest) seems to work for me, too, without having the jupyter kernel installed, or is that just coincidence? Also, you have `kernel='libertem-env'` in the tests, and create `libertem-kernel` here, so there seems to be some mismatch. As a note, when running via tox, I get the following in the logs:

```
2020-07-22 11:56:11 INFO Executing notebook with kernel: python
```

sk1p:
> Also, when I'm running the tests locally, both tests from `test_sd_template.py` fail for me with `ValueError: minvalue must be positive`

That's using Python 3.8 on Linux btw.

uellue:
> fail for me with `ValueError: minvalue must be positive`

Just as a guess, log scaling 0 creates the issue? Maybe setting `vmin=max(1e-30, np,min(data))` can help?

tox.ini
twentyse7en: > Running the tests outside of tox (with plain pytest) seems to work for me, too, without having the jupyter kernel installed, or is that just coincidence?

I already had a jupyter kernel `libertem-env`, so I was able to run plain pytest without any problem. When I tried to run the test using tox, it was failing because of the missing kernel. So I added the instruction to set up a new jupyter kernel and all tests were working fine.

twentyse7en:
> That's using Python 3.8 on Linux btw.

I wasn't able to reproduce the error. I created a new python3.8 virtual env and successfully installed `libertem`, But I wasn't able to import it. No problem with python3.7 virtual env.

tox.ini
twentyse7en: > 2020-07-22 11:56:11 INFO Executing notebook with kernel: python

I too get this log message, even after matching the kernel name. I have no idea how this happens.

sk1p:
> I created a new python3.8 virtual env and successfully installed `libertem`, But I wasn't able to import it. No problem with python3.7 virtual env.

What was the exact error message you got when trying to import libertem?

sk1p:
Btw. I can reproduce this with a clean virtualenv:

```bash
$ mkvirtualenv notebook-generator -p $(which python3.8)
$ pip install -e . -r test_requirements.txt
$ pytest tests/template/test_sd_template.py
```

tox.ini
sk1p: I also get the following warning:


```
tests/template/test_sd_template.py::test_sd_default
tests/template/test_sd_template.py::test_sd_roi
  /home/alex/.virtualenvs/notebook-generator/lib/python3.8/site-packages/traitlets/config/configurable.py:73: DeprecationWarning: Passing unrecoginized arguments to super(ExecutePreprocessor).__init__(kernel='libertem-env').
  object.__init__() takes exactly one argument (the instance to initialize)
  This is deprecated in traitlets 4.2.This error will be raised in a future release of traitlets.
    super(Configurable, self).__init__(**kwargs)
```

From [this example](https://nbconvert.readthedocs.io/en/latest/execute_api.html#example), the parameter should be called `kernel_name`, but then I wonder how installing the kernel in tox.ini fixed things for you. Can you maybe try to remove the kernel installation from tox.ini, remove the kernel from your jupyter config (`~/.local/share/jupyter/kernels` or similar), remove the `kernel='...'` from the `ExecutePreprocessor` instantiation and see what happens, inside of tox and outside? Maybe something unrelated changed and made things work for you.

twentyse7en:
I tried running the test `test_sd_template.py` multiple times, I never encountered the error.
<details>
  <summary>Pytest log</summary>
 ```
============================= test session starts ==============================
platform linux -- Python 3.8.3, pytest-5.4.3, py-1.9.0, pluggy-0.13.1
benchmark: 3.2.3 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=True warmup_iterations=100000)
rootdir: /home/anu/LiberTEM, inifile: pytest.ini
plugins: forked-1.2.0, xdist-1.33.0, cov-2.10.0, asyncio-0.14.0, benchmark-3.2.3
collected 2 items

tests/template/test_sd_template.py ..                                    [100%]

=============================== warnings summary ===============================
tests/template/test_sd_template.py::test_sd_default
  /home/anu/LiberTEM/src/libertem/udf/blobfinder/patterns.py:6: FutureWarning: Blobfinder has moved to its own package LiberTEM-blobfinder with a new structure. Please see https://libertem.github.io/LiberTEM-blobfinder/index.html for installation instructions and documentation of the new structure. Imports from libertem.udf.blobfinder are supported until LiberTEM release 0.6.0.
    warnings.warn(

tests/template/test_sd_template.py::test_sd_default
  /home/anu/LiberTEM/src/libertem/udf/blobfinder/refinement.py:6: FutureWarning: Blobfinder has moved to its own package LiberTEM-blobfinder with a new structure. Please see https://libertem.github.io/LiberTEM-blobfinder/index.html for installation instructions and documentation of the new structure. Imports from libertem.udf.blobfinder are supported until LiberTEM release 0.6.0.
    warnings.warn(

tests/template/test_sd_template.py::test_sd_default
  /home/anu/LiberTEM/src/libertem/udf/blobfinder/correlation.py:8: FutureWarning: Blobfinder has moved to its own package LiberTEM-blobfinder with a new structure. Please see https://libertem.github.io/LiberTEM-blobfinder/index.html for installation instructions and documentation of the new structure. Imports from libertem.udf.blobfinder are supported until LiberTEM release 0.6.0.
    warnings.warn(

tests/template/test_sd_template.py::test_sd_default
  /home/anu/LiberTEM/src/libertem/udf/blobfinder/utils.py:7: FutureWarning: Blobfinder has moved to its own package LiberTEM-blobfinder with a new structure. Please see https://libertem.github.io/LiberTEM-blobfinder/index.html for installation instructions and documentation of the new structure. Imports from libertem.udf.blobfinder are supported until LiberTEM release 0.6.0.
    warnings.warn(

tests/template/test_sd_template.py::test_sd_default
tests/template/test_sd_template.py::test_sd_roi
  /home/anu/libertem-venv/lib/python3.8/site-packages/traitlets/config/configurable.py:73: DeprecationWarning: Passing unrecoginized arguments to super(ExecutePreprocessor).__init__(kernel='libertem-env').
  object.__init__() takes exactly one argument (the instance to initialize)
  This is deprecated in traitlets 4.2.This error will be raised in a future release of traitlets.
    super(Configurable, self).__init__(**kwargs)

-- Docs: https://docs.pytest.org/en/latest/warnings.html
======================== 2 passed, 6 warnings in 13.45s ========================
```
</details>

**Comment by sk1p** [comment]
> I tried running the test `test_sd_template.py` multiple times, I never encountered the error.

Hmm. Is that with a clean virtualenv? Maybe we have different versions of some dependencies installed. From tox output:

> py38 installed: aiohttp==3.6.2,apipkg==1.5,async-timeout==3.0.1,attrs==19.3.0,backcall==0.2.0,bleach==3.1.5,chardet==3.0.4,click==7.1.2,cloudpickle==1.5.0,coverage==5.2,cycler==0.10.0,Cython==0.29.21,dask==2.21.0,decorator==4.4.2,defusedxml==0.6.0,distributed==2.21.0,entrypoints==0.3,execnet==1.7.1,h5py==2.10.0,hdbscan==0.8.26,HeapDict==1.0.1,idna==2.10,imageio==2.9.0,ipykernel==5.3.4,ipython==7.16.1,ipython-genutils==0.2.0,jedi==0.17.2,Jinja2==2.11.2,joblib==0.16.0,jsonschema==3.2.0,jupyter-client==6.1.6,jupyter-core==4.6.3,kiwisolver==1.2.0,libertem @ file:///home/alex/source/upstream-LiberTEM/.tox/.tmp/package/1/libertem-0.6.0.dev0.zip,libertem-blobfinder==0.4.1,llvmlite==0.33.0,MarkupSafe==1.1.1,matplotlib==3.3.0,mistune==0.8.4,more-itertools==8.4.0,msgpack==1.0.0,multidict==4.7.6,nbconvert==5.6.1,nbformat==5.0.7,ncempy==1.6.0,networkx==2.4,numba==0.50.1,numpy==1.19.1,packaging==20.4,pandocfilters==1.4.2,parso==0.7.0,pexpect==4.8.0,pickleshare==0.7.5,Pillow==7.2.0,pluggy==0.13.1,primesieve==2.2.0,prompt-toolkit==3.0.5,psutil==5.7.2,ptyprocess==0.6.0,py==1.9.0,py-cpuinfo==7.0.0,Pygments==2.6.1,pyparsing==2.4.7,pyrsistent==0.16.0,pytest==5.4.3,pytest-asyncio==0.14.0,pytest-benchmark==3.2.3,pytest-cov==2.10.0,pytest-forked==1.2.0,pytest-xdist==1.33.0,python-dateutil==2.8.1,PyWavelets==1.1.1,PyYAML==5.3.1,pyzmq==19.0.1,scikit-image==0.17.2,scikit-learn==0.23.1,scipy==1.5.1,six==1.15.0,sortedcontainers==2.2.2,sparse==0.10.0,tblib==1.6.0,testpath==0.4.4,threadpoolctl==2.1.0,tifffile==2020.7.22,toolz==0.10.0,tornado==6.0.4,tqdm==4.48.0,traitlets==4.3.3,wcwidth==0.2.5,webencodings==0.5.1,websockets==8.1,yarl==1.4.2,zict==2.0.0

**Comment by sk1p** [comment]
Here is a notebook that reproduces the error for me: [reproducer.zip](https://github.com/LiberTEM/LiberTEM/files/4965507/reproducer.zip). Basically, plotting an all-zero array with `LogNorm` fails for some reason, as something sets `vmin` to a negative value (???). Can you check what version of `matplotlib` you were testing with?

**Comment by twentyse7en** [comment]
> matplotlib

3.3.0

**Comment by sk1p** [comment]
[snip]

> 3.3.0

Hmm, if you run the following, do you get the error or not?

```python
plt.figure()
plt.imshow((np.zeros((128, 128))), norm=colors.LogNorm())
plt.colorbar()
```

(without the colorbar, there is no error for me, btw.)

twentyse7en:
> Hmm, if you run the following, do you get the error or not?

Yes I'm getting error, `ValueError: minvalue must be positive`. No error when `colorbar` is removed.

twentyse7en:
Tests work fine with just `ipykernel`, without instruction for installing kernel for venv. `ExecutePreprocessor` will use the default python kernel. It doesn't throw an error when a wrong kernel is provided, in that case it might be using default python kernel in the venv.

sk1p:
> Yes I'm getting error, `ValueError: minvalue must be positive`. No error when `colorbar` is removed.

The stddev of the `hdf5_ds_1` should be 0 (as it is all ones), and should cause the error in the test run. So I wonder if there is any other difference between your test environment and the one where the above causes an error?

> Tests work fine with just ipykernel, without instruction for installing kernel for venv. ExecutePreprocessor will use the default python kernel. It doesn't throw an error when a wrong kernel is provided, in that case it might be using default python kernel in the venv.

:+1:

twentyse7en:
> The stddev of the `hdf5_ds_1` should be 0 (as it is all ones)

Instead of all zeros, I'm getting all `3.10235215e-32`. BTW I'm using `virtualenv` for creating venv.

twentyse7en:
Should we remove the `colorbar` for now and consider this in the next iteration.

twentyse7en:
> The stddev of the `hdf5_ds_1` should be 0 (as it is all ones), and should cause the error in the test run. So I wonder if there is any other difference between your test environment and the one where the above causes an error?

How could be the tox tests run without any error? does that fail for you @sk1p?

sk1p:
> How could be the tox tests run without any error? does that fail for you @sk1p?

Yeah, running tests via tox fail for me, too. 

> Instead of all zeros, I'm getting all 3.10235215e-32

Ah-ha! That's the difference! 

> Should we remove the colorbar for now and consider this in the next iteration.

Maybe a good idea - @uellue what do you think?

uellue:
> Maybe a good idea - @uellue what do you think?

Sure, it is already very useful without it. We should just keep this information for a follow-up issue since it is mostly a problem in testing and sounds fixable. :-)

uellue:
Congratulations and many thanks. Merging! 🥳

uellue:
By the way, with 148 comments the most commented PR in LiberTEM so far, by a wide margin! A serious amount of work from all sides. Congratulations and many thanks again!

uellue:
@twentyse7en Just FYI, in the last days I've been using the code generator several times. REALLY nice to not start from scratch, but from working code with just a few clicks. :-)

twentyse7en:
Glad to hear that @uellue : )

PR #812, opened 2020-06-10
Add GPU configuration to the GUI

### LINKED ISSUE #803: Add GPU configuration to the GUI
As continuation of #760: we should allow users to opt-in to GPU processing. What would be the best place to integrate this? I guess when starting the cluster? Or as a runtime configuration, where the user can enable/disable GPU processing after the fact?

I guess once we have a cluster connection, we can trust that the user created it with the right parameters, and use the GPU workers if available. So the first option would work well.

So integration steps would be:

- [x] add GPU information to the configuration that is returned on connection by the web API
- [x] add GPU checkboxes to the "make local cluster" form
- [x] send selected GPUs back to the web API and pass them to the cluster spec creation code

Does that sound right? If so, I think it is easy enough to do in 0.6, and I'll take a stab at it after #760 is merged, if no one else volunteers.

sk1p:
Fixes #803

## TODO

- [x] Use GPUs by default
- [x] Inform the user if `cupy` is not installed, but GPUs are available, and possibly disable GPU dropdown in that case
- [x] Update for recent changes in #760

## Contributor Checklist:

changes were made to the GUI)


sk1p:
> * Should we activate CUDA devices by default and let the user deactivate? That's currently the behavior of `api.Context()`.

Yes, we should match the behavior of `Context` here, I'll change the default.

> * What happens if `numba.cuda` detects CUDA devices, but there's no cupy installed?

That's a good point - the configuration dict returned by the web API should also include information about installed packages (cupy, torch, hdbscan), so we can warn the user if there is no cupy installation available, and/or even disable the GPU dropdown, as long as we don't have CUDA-but-not-cupy UDFs available in the GUI.

uellue:
> That's a good point - the configuration dict returned by the web API should also include information about installed packages (cupy, torch, hdbscan), so we can warn the user if there is no cupy installation available, and/or even disable the GPU dropdown, as long as we don't have CUDA-but-not-cupy UDFs available in the GUI.

Ok, we should discuss on Monday how to best integrate and support that. :-)

uellue:
In #760 I just pushed support to run hybrid (NumPy, CuPy) UDFs on clusters with CUDA, but without CuPy. They will only be scheduled on the CPU in that case.

For that I updated the `detect()` and `cluster_spec()` methods. Let me know how I can help to adapt your code to the changes! :-)

sk1p:
> In #760 I just pushed support to run hybrid (NumPy, CuPy) UDFs on clusters with CUDA, but without CuPy. They will only be scheduled on the CPU in that case.
> 
> For that I updated the `detect()` and `cluster_spec()` methods. Let me know how I can help to adapt your code to the changes! :-)

:+1: - I'll take care of this once #760 is merged

sk1p:
A warning is now shown when no cupy is available:

image

And GPUs are enabled by default. 

> For that I updated the detect() and cluster_spec() methods. Let me know how I can help to adapt your code to the changes! :-)

I think the `detect` usage doesn't need updates, right? Because we do:

```python
devices = detect()
devices["cudas"] = some_new_cudas
DaskJobExecutor.make_local(spec=cluster_spec(**devices, ...))
```

So `has_cupy` etc. should be passed through in any case.

uellue:
Just needs the rebuilt client now, right?

uellue:
> I think the `detect` usage doesn't need updates, right? Because we do:

That was the hope when introducing the new parameter this way, glad that it worked out! :-)

uellue:
> A warning is now shown when no cupy is available:

Nice, that should point users in the right direction to make use of CUDA support.

sk1p:
> Just needs the rebuilt client now, right?

Yep, will do that now!

uellue:
Nice, merged!

PR #817, opened 2020-06-17
Make the `correct` tests a bit more strict

sk1p:
Don't accidentally catch exceptions

## Contributor Checklist:

### DIFF
--- a/src/libertem/executor/dask.py
+++ b/src/libertem/executor/dask.py
@@ -299,10 +299,7 @@ class DaskJobExecutor(CommonDaskMixin, JobExecutor):
     def close(self):
         if self.is_local:
             if self.client.cluster is not None:
-                try:
-                    self.client.cluster.close(timeout=1)
-                except tornado.util.TimeoutError:
-                    pass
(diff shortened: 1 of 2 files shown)


sk1p:
Hm, the Mac OS X test once again failed - should we maybe allow failures for them? /cc @uellue

uellue:
> Hm, the Mac OS X test once again failed - should we maybe allow failures for them? /cc @uellue

Did you restart it? I can't seem to find a failed test... Do you have a link?

sk1p:
Yes, I restarted - sadly the old logs seem to be erased in that case. IIRC the error was that the scheduler couldn't bind to the given address (already in use) - and then it hit another issue in distributed (https://github.com/dask/distributed/pull/3912).

I have a hunch what the problem could be: the local cluster destruction was a bit impatient and had a short timeout. I'll increase the timeout and actually not swallow the exception in that case, let's see if that improves things.

uellue:
👍 Fingers crossed! :-)

uellue:
Nice, merged!

Issue #814, opened 2020-06-12
Severe performance regression (at least) in K2IS since 0.5.0


uellue:
The K2IS dataset reached the same pixel-per-pixel performance as raw in the past. On Moellenstedt this was about 5 GPix/s. The 202 GB dataset /cachedata/users/clausen/K2IS01_vacuum_.gtg should fit the file system cache on Moellenstedt, and if not, the SSD RAID it is stored on can deliver 6 GB/s. That means this file should be analyzed in just about half a minute under ideal conditions.

A ring analysis on this file in a notebook took way over 10 minutes in tests using the current master at this time and the state of #812. I aborted the tests because they took too long.

The 0.5.0 release version reached the expected performance:

```
# First run
CPU times: user 9.5 s, sys: 1.06 s, total: 10.6 s
Wall time: 35.7 s

# Second run
CPU times: user 8.21 s, sys: 1.01 s, total: 9.22 s
Wall time: 35.1 s
```

Zipped test notebook
[k2is-performance.zip](https://github.com/LiberTEM/LiberTEM/files/4769722/k2is-performance.zip)

Other file types were not tested yet.

We should discuss if we fast-track at least a prototype version of #198 for the 0.6 release since we have to get a handle on this before the release one way or another, in particular testing all the file formats we support, GPU processing, and a good sample of analyses.

Edit: ...plus, we still have #798 on our list for 0.6 which falls in the same category.

AnandBaburajan:
I'm not sure if this is helpful, but I'm getting `distributed.worker - WARNING - Memory use is high but worker has no data to store to disk.  Perhaps some other process is leaking memory?  Process memory: 1.82 GB -- Worker memory limit: 2.09 GB` when I run an analysis on a K2IS dataset.

uellue:
@AnandBaburajan Thx! Not sure if it has the same root cause as the performance regression, but definitely something that shouldn't happen. How do you reproduce it?

AnandBaburajan:
> @AnandBaburajan Thx! Not sure if it has the same root cause as the performance regression, but definitely something that shouldn't happen. How do you reproduce it?

I ran a disk analysis through the GUI (current master) on a ~ 8 GB K2IS dataset.

sk1p:
> Memory use is high but worker has no data to store to disk

[...]

PR #820, opened 2020-06-18
Remove nested function in UDFRunner to facilitate profiling

uellue:
This is a complex and performance-critical part of the UDF interface.

Refs #814 

## Contributor Checklist:


uellue:
Going ahead and merging since it is a simple refactor. :-)

PR #831, opened 2020-06-23
Make sure correction handles integer input data correctly

### LINKED ISSUE #818: Correct function does not fix bad pixel
I recently tried to implement correction of bad pixel using correct function
```sig_shape = (256,256)```
```x =np.array([(240,174,206,143), (157,228,242,41)]).astype(np.int64)```
```data_new = correct(data, dark_image = None, gain_map = None, excluded_pixels = x, inplace = False, sig_shape = sig_shape)```

However, when summing all frames it seems the correct function does not change the specific bad pixel from the surrounding pixel. Is there something wrong with the implementation?
image895

### PR DESCRIPTION
**Pull request by uellue**
Closes #818 

Correction only makes sense for floating point data.

* Inplace: Throws TypeError for non-float data
* Copy: Cast to matching floating point type.

@sk1p How is it handled in the dataset corrections code? Do we make sure it is float when correction is turned on?

## Contributor Checklist:

### DIFF
--- a/src/libertem/corrections/corrset.py
+++ b/src/libertem/corrections/corrset.py
@@ -84,3 +84,58 @@ class CorrectionSet:
             inplace=True,
             sig_shape=tuple(tile_slice.shape.sig),
         )
+
+    def adjust_tileshape(self, tile_shape, sig_shape):
+        excluded_pixels = self.get_excluded_pixels()
+        if excluded_pixels is None:
+            return tile_shape
+        if tile_shape[1:] == sig_shape:
+            return tile_shape
+        excluded_list = excluded_pixels.coords
+        adjusted_shape = np.array(tile_shape)
+        # Map of dimensions that should be shrunk
+        # We have to grow or shrink monotonously to not cycle
+        shrink = (adjusted_shape[1:] >= 4)
+        # Try to iteratively reduce or increase tile size
+        # per signal dimension in case of intersections of tile boundaries
+        # with the patching environment
+        # until we avoid all excluded pixels.
+        # This may fail in case of many excluded pixels or full excluded rows/columns,
+        # depending on the tiling scheme. In that case,
+        # swith to full frames while preserving tile size if possible.
+        for repeat in range(4):
+            clean = True
+            for dim in range(1, len(adjusted_shape)):
+                start = adjusted_shape[dim]
+                stop = sig_shape[dim - 1]
+                step = adjusted_shape[dim]
+                for boundary in range(start, stop, step):
+                    # Pixel on the left side of boundary
+                    if boundary - 1 in excluded_list[dim - 1]:
+                        if shrink[dim - 1]:
+                            adjusted_shape[dim] -= 2
+                        else:
+                            adjusted_shape[dim] += 1
+                        clean = False
+                    # Pixel on the right side of boundary
+                    if boundary in excluded_list[dim - 1]:
+                        if shrink[dim - 1]:
+                            adjusted_shape[dim] -= 1
+                        else:
+                            adjusted_shape[dim] += 2
+                        clean = False
+            if clean:
+                break
+            if np.any(adjusted_shape <= 0) or np.any(adjusted_shape[1:] > sig_shape):
+                # We didn't find a solution
+                clean = False
+                break
+        if clean:
+            return tuple(adjusted_shape)
+        else:
+            # No solution found, switch to full frames
+            # at constant tile size
+            n_pix = np.prod(tile_shape)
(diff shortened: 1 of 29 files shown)

### CODE REVIEW

**Comment by sk1p** [comment]
> How is it handled in the dataset corrections code? Do we make sure it is float when correction is turned on?

That's a good point - I remember the FRMS6 dataset had specific code for this case to make sure of that. Currently, the `UDFRunner` decides the read dtype, so this is probably not solved yet in the current version. Ideally, the `UDFRunner` should get the I/O backend from the `DataSet`, the I/O backend knows the corrections and can then be queried what the preferred dtype should be (`float32` in case of corrections, `None` otherwise?)

**Comment by sk1p** [comment]
Maybe we can construct an easy test case together, using the raw `DataSet`, which shows the problem, and I can see if I can fix this before the release?

**Comment by sk1p** [comment]
> Maybe we can construct an easy test case

I thought about this, how about something like this:

```
data = np.ones((1, 1, 3, 3), dtype=np.uint8)
data[..., 1, 1] = 0
defect_pixels=[(1,), (1,)] # or whatever the right expression for the coords is
```

That should fail to correct using `PickUDF` right now, but work if we force `read_dtype=np.result_type(float, read_dtype)`, right? Should even work with the memory data set...

File: src/libertem/io/dataset/base/tiling.py [hunk 1]
```diff
@@ -511,6 +523,22 @@ def get_scheme(self, udfs, partition, read_dtype: np.dtype, roi: np.ndarray):
 
         # the partition has a "veto" on the tileshape:
         tileshape = partition.adjust_tileshape(tileshape)
+        # The correction has to make sure that there are no excluded pixels
+        # at tile boundaries
+        tileshape_corrected = corrections.adjust_tileshape(tileshape, tuple(partition.shape.sig))
```
sk1p, line 528: Hmm, I think this may need further tweaking - `adjust_tileshape` assumes that the negotiation did take the `base_shape` into account, so for example k2is doesn't adjust the tileshape back to something that is divisible by `(1, 930, 16)`.

`adjust_tileshape` is just a blunt instrument to make formats like HDF5 and SER work, where we don't necessarily have direct control over how the data is read (HDF5 ds could be improved to work better with chunking, but that's another story)

uellue: 👍

uellue:
Hi @sk1p when you have time, can you have a look at the new code, in particular code structure? Codeclimate is not happy for good reasons, and I could use inspiration for improvement. :-)

sk1p:
> when you have time, can you have a look at the new code, in particular code structure? Codeclimate is not happy for good reasons, and I could use inspiration for improvement. :-)

Sure thing, I'll have a detailed look tomorrow.

src/libertem/corrections/corrset.py
```diff
@@ -1,10 +1,54 @@
...
+def min_disjunct_multiplier(excluded, size):
+    '''
+    Calculate a small integer i for which i * n not in "excluded" for any n > 0
+
+    To make sure that the tile shape negotiation retains as much flexibility as possible,
+    it is important to find a small integer and not just any integer that fulfills
+    this condition.
+    '''
+    # This is guaranteed to include "size" for realistic dimension sizes
+    primes = np.array(primesieve.primes(size*2))
```
sk1p, line 34: This could directly use `primesieve.numpy.primes`, which is supposed to be [three times faster](https://github.com/kimwalisch/primesieve-python#numpy-support)

uellue: Thx! :-)

src/libertem/corrections/corrset.py
```diff
@@ -84,3 +128,78 @@ def apply(self, data: np.ndarray, tile_slice: Slice):
...
+                    forbidden = np.concatenate((excluded_list[dim], excluded_list[dim] + 1))
+                    forbidden = forbidden[forbidden < stop]
+                    nonzero_filter = forbidden != 0
+                    m = min(stop, min_disjunct_multiplier(forbidden[nonzero_filter], stop))
+                    # In case we have a zero where the existing logic doesn't work
+                    if not np.all(nonzero_filter):
+                        m = max(m, 2)
+                    if adjusted_shape[dim] != m:
+                        adjusted_shape[dim] = m
+                        clean = False
```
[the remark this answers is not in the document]
uellue: Thx, I hope the last push is going in the right direction? :-)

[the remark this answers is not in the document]
uellue: ...at least codeclimate is less unhappy now! :-D

sk1p:
Thanks, merged!

PR #844, opened 2020-07-10
Overflow fix for np.prod

AnandBaburajan:
Added `dtype=np.int64` to `np.prod` to fix the `RuntimeWarning: overflow encountered in long_scalars` error while negotiating a tiling scheme.

## Contributor Checklist:

### DIFF
--- a/src/libertem/common/buffers.py
+++ b/src/libertem/common/buffers.py
@@ -33,7 +33,7 @@ def bytes_aligned(size):
 
 
 def empty_aligned(size, dtype):
-    size_flat = np.product(size)
+    size_flat = np.prod(size, dtype=np.int64)
     dtype = np.dtype(dtype)
     buf = _alloc_aligned(dtype.itemsize * size_flat)
     # _alloc_aligned may give us more memory (for alignment reasons), so crop it off the end:
@@ -42,7 +42,7 @@ def empty_aligned(size, dtype):
 
 
 def zeros_aligned(size, dtype):
-    if dtype == np.object or np.product(size) == 0:
+    if dtype == np.object or np.prod(size, dtype=np.int64) == 0:
         res = np.zeros(size, dtype=dtype)
     else:
         res = empty_aligned(size, dtype)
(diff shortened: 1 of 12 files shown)


AnandBaburajan:
> My only question is if the test actually fails in case the overflow is triggered

Sorry, it doesn't! Should I change the test so that it deliberately fails if the overflow warning is raised?

> if the chosen size is large enough to trigger the error

Yes, the test dataset's calculated `partition_size` is 3626434560 > `np.int32`'s upper range

uellue:
> Should I change the test so that it deliberately fails if the overflow warning is raised?

Yes, that would be perfect, thanks in advance! 👍 :-)

AnandBaburajan:
> > My only question is if the test actually fails in case the overflow is triggered
> 
> Sorry, it doesn't! Should I change the test so that it deliberately fails if the overflow warning is raised?

Running an analysis in a notebook or with the GUI throws the warning, but not when run within a test. I couldn't figure out why but I think it's because the warning isn't captured as an exception. I could capture the warning in the tiling code but that doesn't make much sense to me as the bug has been fixed, so now I'm curious if there's need of a test at all.

uellue:
This class of bug is actually recurring in LiberTEM, and that's why it could be nice to not only fix this instance, but make sure it can't happen that easily again in other size calculations.

Looking into the issue a bit more, the warning is not reliable and that is a long-standing issue in NumPy: https://github.com/numpy/numpy/issues/8987. One could catch the warning with https://docs.pytest.org/en/stable/warnings.html#recwarn, but that could give a false sense of security.

Furthermore, this is related to a somewhat non-intuitive behavior of the `np.int` dtype, which is 32 bit on 64 bit Windows: https://github.com/numpy/numpy/issues/9464.

Since the results of `prod` grow large very very fast and routinely exceed the value range for 32 bit, any use of `np.prod` on integers smaller than 64 bit without using the `dtype` parameter or using `np.int` for anything that is a size should be considered a bug waiting to happen.

As a proposal, we could just remove the ineffective test for now, merge your fix, and I open an issue to address the issue more profoundly.

AnandBaburajan:
> any use of `np.prod` on integers smaller than 64 bit without using the `dtype` parameter or using `np.int` for anything that is a size should be considered a bug waiting to happen.

I searched the source code and set the dtype as `np.int64` for all the size calculations using `np.prod`

uellue:
Ah, another point: Your code doesn't patch the use of `np.prod` in the Numba portion of `tiling.py`. The Numba implementation doesn't support the `dtype` kwarg. Instead one could cast the array to `int64` before applying `np.prod()`.

uellue:
Many thanks, merging! :+1:

PR #851, opened 2020-07-20
Import color wheels from empyre instead of using a copy

uellue:
Empyre has been released since the copy was created and we can use it
directly now.

LiberTEM contained a custom 2D version. The downstream code is now adapted to use the
3D visualization functions of empyre by setting z to 0.

Since empyre creates an ndarray from the input tuple, having unequal length
or incompatible shapes will trigger "Creating an ndarray from ragged nested
sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with
different lengths or shapes) is deprecated." since NumPy 1.19.

For that reason we use `np.broadcast_arrays()` to make the shapes and sizes match.

Refs #801 

Not sure if it needs a changelog entry since it is an internal API?

CC @jan-car :-)

## Contributor Checklist:


sk1p:
> LiberTEM contained a custom 2D version. The downstream code is now adapted to use the
3D visualization functions of empyre by setting z to 0.

Ah, right, we did change something! Looks good!

> Since empyre creates an ndarray from the input tuple, having unequal length
or incompatible shapes will trigger "Creating an ndarray from ragged nested
sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with
different lengths or shapes) is deprecated." since NumPy 1.19.

Maybe also could be fixed in empyre? It could do the broadcasting internally...

> Not sure if it needs a changelog entry since it is an internal API?

Don't think it needs one, true.

uellue:
> Maybe also could be fixed in empyre? It could do the broadcasting internally...

https://iffgit.fz-juelich.de/empyre/empyre/-/merge_requests/33

uellue:
It looks like `empyre` won't install on Python 3.6?

uellue:
> It looks like `empyre` won't install on Python 3.6?

https://iffgit.fz-juelich.de/empyre/empyre/-/blob/master/setup.cfg#L36

...what to do about that?

twentyse7en:
- [ ] Fix the tests, tests/template/test_radial_fourier_template.py, and tests/template/test_com_template.py, which requires `empyre` as a dependency.
- [ ] Remove the note mentioning Python 3.7 and `empyre` as a requirement for visualization from radialfourier.py and com.py.

uellue:
Bumped to 0.7milestone since @jan-car is on vacation, we likely need time on empyre side, and we don't have time pressure here since the current state works perfectly well. :-)

sk1p:
What's the status on this one - should we do an empyre release on Thursday?

jan-car:
Very nice to see this come to fruition 😄 🎉 !

tests/template/test_com_template.py
```diff
@@ -1,14 +1,13 @@
...
-import pytest
+
 import numpy as np
 import nbformat
 from temp_utils import _get_hdf5_params
 from libertem.web.notebook_generator.notebook_generator import notebook_generator
 from nbconvert.preprocessors import ExecutePreprocessor
 
 
-@pytest.mark.skip(reason="requires `empyre`")
```
sk1p, line 11: The same needs to be done in `test_radial_fourier_template`

sk1p (changes requested):

tests/template/test_com_template.py
uellue: Thx! I'll look into that later. Just saw the note from @twentyse7en above, will follow up on that.

uellue:
Actually, it should be (data.real, data.imag, 0) since empyre is x, y, z and the imaginary part is commonly plotted on the y axis.

sk1p (approved):

sk1p:
Merged, thank you!

PR #864, opened 2020-08-10
Improve the cluster template

twentyse7en:
## Contributor Checklist:

changes were made to the GUI)


twentyse7en:
How to refactor the `get_udf_results`, It's directly copied to the notebook ?

uellue:
Thank you for the PR! Getting the source code by inspection and adapting it for the notebook looks very neat to me. :-) Looks like this addresses the previous comments regarding clustering, well done!

Regarding the visualization, there is actually a little inconsistency in the analysis. It returns the feature vector instead of the classification. Here's a possible change: Instead of 

```python
    return AnalysisResultSet([
            AnalysisResult(raw_data=udf_results['intensity'].data,
                           visualized=visualize_simple(
                               labels.reshape(self.dataset.shape.nav)),
                           key="intensity", title="intensity",
                           desc="result from integration over mask in Fourier space"),
        ])
```

it can perhaps be 

```python
    return AnalysisResultSet([
        AnalysisResult(raw_data=labels.reshape(dataset.shape.nav),
                       visualized=visualize_simple(
                           labels.reshape(dataset.shape.nav)),
                       key="intensity", title="intensity",
                       desc="result from integration over mask in Fourier space"),
    ])
```

As far as I am aware, this is not used anywhere else, so it should be safe to change.

Regarding the CI failure I am not sure where that comes from. The code portion where the error is reported seems unrelated. We'd have to see if this is caused by changes in this PR that "confuse" Sphinx somewhere down the line, or if it is a separate issue.

uellue:
@sk1p do you think we can merge it despite the CI failure and address that separately, or should we get to the bottom of it first? Other than that, it looks ready to merge from my perspective. What is your take?

sk1p:
> do you think we can merge it despite the CI failure and address that separately, or should we get to the bottom of it first? Other than that, it looks ready to merge from my perspective. What is your take?

I'm checking if I can reproduce locally, give me a sec...

uellue:
> I'm checking if I can reproduce locally, give me a sec...

Doing the same atm. :-) With my previous `docs-check` environment it didn't happen. Rebuilding the environment now.

sk1p:
Okay, after making sure the latest sphinx version is used, I also got the error. I think this is related to the new Sphinx release from 3 days ago, so we should fix this in a separate PR, agreed.

sk1p:
About the PR at hand: overall I like it, the only thing where I stumbled a bit was assigning `cancel_id`, `executor` and `job_is_cancelled` to `self`. That's a bit of a hack and I'm looking into it right now...

sk1p:
Something else I just noticed: the indentation in the "copy" tab seems to be off/missing:

image

Maybe needs to be wrapped into a `<pre>` or similar?

sk1p:
Something else I just noticed, I think this was not caused by this PR but anyways: when copying the complete notebook, some parts are joined together using commas instead of newlines, here always after the plotting:

image

sk1p:
> the only thing where I stumbled a bit was assigning `cancel_id`, `executor` and `job_is_cancelled` to `self`. That's a bit of a hack and I'm looking into it right now...

Hmm, it seems those are only used in `run_sd_udf`, which has a complete different implementation for the notebook anyways. So maybe those can be passed directly?

src/libertem/analysis/clust.py
```diff
@@ -24,47 +25,75 @@ class ClusterTemplate(GeneratorHelper):
...
 
     def get_docs(self):
         docs = ["# Cluster Analysis"]
         return '\n'.join(docs)
 
+    def get_run_sd_udf(self):
+        indent = " "*4
+        temp_run_sd = [
+            "def run_sd_udf(roi, stddev_udf):",
+            f"{indent}sd_udf_results = ctx.run_udf(dataset=ds, udf=stddev_udf, roi=roi)",
```
sk1p, line 69: Should probably also add `progress=True` here.

uellue:
Should we clean up the "Copy" part in this PR as well? I had planned to review it in a separate issue since it is already merged, right? But we can also just do it here. :-)

uellue:
Regarding the CI failure: #865

sk1p:
> Should we clean up the "Copy" part in this PR as well? I had planned to review it in a separate issue since it is already merged, right? But we can also just do it here. :-)

I'll let @twentyse7en decide on this - do you want to work general "copy notebook" feedback into this PR, or do it separately?

twentyse7en:
> do you want to work general "copy notebook" feedback into this PR, or do it separately?

separately

twentyse7en:
> I stumbled a bit was assigning `cancel_id`, `executor` and `job_is_cancelled` to `self`.

Is that a bad practice?

twentyse7en:
> Hmm, it seems those are only used in `run_sd_udf`, which has a complete different implementation for the notebook anyways. So maybe those can be passed directly?

Okay

sk1p:
> > I stumbled a bit was assigning `cancel_id`, `executor` and `job_is_cancelled` to `self`.
> 
> Is that a bad practice?

Yes, because it creates a tight coupling between the two methods: you can't call `run_sd_udf` outside of the controller. If you try, you get an attribute error. It also creates some surprising behavior, as you suddenly have state in the `Analysis` class: once you have run the controller method once, you actually _can_ call `run_sd_udf`, but `executor`, `cancel_id` and so on may have values that don't match what you are trying to do.

Then, if you want to change `run_sd_udf` later for some reason, you have to look up what `executor`, `cancel_id` etc. are, because they are not arguments and also they are not set by the class constructor, so you need to search the code to find the assignments in the `controller` method.

twentyse7en:
> Maybe needs to be wrapped into a `<pre>` or similar?

Before `<pre>` :
nooverlap

After `<pre> ` :
overlap

It overlaps with the button and goes outside the cell for long lines. Indentation is fixed as expected.

sk1p:
> It overlaps with the button and goes outside the cell for long lines. Indentation is fixed as expected.

Okay, `<pre>` most of the time needs to be combined with `overflow-x: scroll` or similar construct, this should fix the issue and possibly introduce horizontal scroll bars. If the code lines are consistently too long, it may be a good idea to format them differently.

twentyse7en:
> possibly introduce horizontal scroll bars.

horizontal scroll bar is already available, It would be better to reformat the long lines.

sk1p:
> horizontal scroll bar is already available, It would be better to reformat the long lines.

Okay, another way to to prevent the overlapping elements is to use flexbox for layouting the cell, instead of floating the button

src/libertem/analysis/clust.py
```diff
@@ -24,47 +25,76 @@ class ClusterTemplate(GeneratorHelper):
...
+        ]
+        return "\n".join(temp_run_sd)
+
     def get_analysis(self):
-        temp_analysis = [f"parameters={self.params}",
-                         "cluster_analysis = ClusterAnalysis(dataset=ds, parameters=parameters)"]
-        temp_analysis.append(self.temp_cluster_controller())
-        return '\n'.join(temp_analysis)
+        udf_results = self.format_code(inspect.getsource(ClusterAnalysis.get_udf_results))
+        sd_roi = self.format_code(inspect.getsource(ClusterAnalysis.get_sd_roi))
```
sk1p, line 77: Hmm, maybe `get_sd_roi` could be replaced by a call to `libertem.analysis.getroi.get_roi()`? That should reduce the amount of code a bit. Could also be changed in the analysis itself.

src/libertem/analysis/clust.py
```diff
@@ -24,47 +25,76 @@ class ClusterTemplate(GeneratorHelper):
...
+        run_sd_udf = self.get_run_sd_udf()
+        cluster_controller = self.temp_cluster_controller()
+
+        temp_analysis = [
+                    f"parameters={self.params}\ndataset=ds",
+                    f"{sd_roi}",
+                    f"{run_sd_udf}",
+                    f"{sd_results}",
+                    f"{cluster_udf}",
+                    f"{udf_results}",
```
sk1p, line 89: There should be vertical white space between the function definitions, ideally following PEP8 (two empty lines between top-level functions)

tests/template/test_clust_template.py
```diff
@@ -0,0 +1,75 @@
...
+
+    notebook = notebook_generator(conn, dataset, analysis, save=True)
+    notebook = io.StringIO(notebook.getvalue())
+    nb = nbformat.read(notebook, as_version=4)
+    ep = ExecutePreprocessor(timeout=600)
+    out = ep.preprocess(nb, {"metadata": {"path": datadir}})
+    data_path = os.path.join(datadir, 'cluster_result.npy')
+    results = np.load(data_path)
+
+    analysis = ClusterAnalysisTest(dataset=hdf5_ds_2, parameters={
```
sk1p, line 60: I would prefer if this tested against the "real" `ClusterAnalysis`. I guess the challenge was the `async` methods? It should be possible to use a `Context` with an `async_executor` instead, mark the test as an `asyncio` test and use the original class here. For "faking" `send_results`, you could do something like this:

```python
class ResultContainer:
    def __call__(self, results, finished):
        self.results = results
        self.finished = finished

res_container = ResultContainer()
# ...
await analysis.controller(..., send_results=res_container)
```

Then you can grab the results from `res_container.results`. Or you can use a `unittest.mock.Mock` object and assert on the `call_args_list`. What do you think?

src/libertem/web/notebook_generator/template.py
```diff
@@ -24,6 +25,9 @@ class TemplateBase:
 
     temp_save = ["np.save('${short}_result.npy', ${short}_result['intensity'])"]
 
+    def code_formatter(self, code):
+        return autopep8.fix_code(code)
```
sk1p, line 29: :+1:

src/libertem/analysis/clust.py
```diff
@@ -24,47 +25,76 @@ class ClusterTemplate(GeneratorHelper):
     def __init__(self, params):
         self.params = params
 
+    def format_code(self, code):
+        code = re.sub(r"self(\.|, |)|(, |)executor|, cancel_id|, job_is_cancelled", "", code)
+        # remove indentation
```
sk1p, line 30: There is actually a function available for this in the standard library, as part of the `textwrap` package: [dedent](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) - maybe that can be used instead?

src/libertem/analysis/clust.py
```diff
@@ -24,47 +25,76 @@ class ClusterTemplate(GeneratorHelper):
...
+                    f"{sd_results}",
+                    f"{cluster_udf}",
+                    f"{udf_results}",
+                    f"{cluster_controller}"
+                    ]
+        return '\n\n'.join(temp_analysis)
 
     def get_plot(self):
         plot = ["plt.figure()",
                 "plt.imshow(cluster_result['intensity'].visualized)"]
```
sk1p, line 96: I think here it would be better to `imshow` the raw results, such that the user can inspect the underlying values when plotting interactively.

uellue: Yes in principle. However, the analysis currently has an inconsistency where the raw data is not available. Maybe we can make that a FIXME for now?

tests/template/test_clust_template.py
twentyse7en: Sounds good! I was a little confused earlier.

src/libertem/analysis/clust.py
twentyse7en: > the analysis currently has an inconsistency where the raw data is not available.

I didn't get it, what is the inconsistency here?

src/libertem/analysis/clust.py
twentyse7en: There are currently two blank lines between the functions.

image

sk1p: Hm, I didn't see the blank lines in the "copy notebook" view - maybe it is just an issue with the presentation in that view?

src/libertem/analysis/clust.py
sk1p: > Yes in principle. However, the analysis currently has an inconsistency where the raw data is not available. Maybe we can make that a FIXME for now?

Hm, as far as I can see, the `AnalysisResultSet` returned from the clustering analysis was changed here to return the labels as `raw_data` - wouldn't that work for visualization in the notebook?

uellue: Ah I see, I didn't notice that it was changed already. :-) Yes, that will work!

uellue (approved):
Tested successfully! Ready to merge from my perspective. 👍 well done!

uellue:
Running the notebook in the unit tests and comparing results to the actual analysis is VERY cool IMO, by the way. :-) Many thanks for putting in the effort to make it work!

twentyse7en:
> Many thanks for putting in the effort to make it work!

Thank you, guys!! Every time I'm learning something new.

sk1p (approved):
Looks good to me, too, thanks!

PR #871, opened 2020-09-01
Handle empty rect ROI

AnandBaburajan:
Fix for empty rectangular ROI in the GUI throwing `UnboundLocalError: local variable 'ymin' referenced before assignment`

## Contributor Checklist:

### DIFF
--- a/tests/test_masks.py
+++ b/tests/test_masks.py
@@ -100,3 +100,9 @@ def test_rectmask():
     assert np.allclose(m.rectangular(2, 2, -3, 3, 5, 5), rect3)
     assert np.allclose(m.rectangular(2, 2, 3, -3, 5, 5), rect2)
     assert np.allclose(m.rectangular(2, 2, -3, -3, 5, 5), rect1)
+
+
+def test_empty_rectmask():
+    assert not np.all(m.rectangular(2, 2, 0, 3, 5, 5))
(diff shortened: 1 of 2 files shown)


tests/test_masks.py
```diff
@@ -100,3 +100,9 @@ def test_rectmask():
     assert np.allclose(m.rectangular(2, 2, -3, 3, 5, 5), rect3)
     assert np.allclose(m.rectangular(2, 2, 3, -3, 5, 5), rect2)
     assert np.allclose(m.rectangular(2, 2, -3, -3, 5, 5), rect1)
+
+
+def test_empty_rectmask():
+    assert not np.all(m.rectangular(2, 2, 0, 3, 5, 5))
+    assert not np.all(m.rectangular(2, 2, 3, 0, 5, 5))
+    assert not np.all(m.rectangular(2, 2, 0, 0, 5, 5))
```
sk1p, line 108: `not np.all` means that there is at least one false-y element in the array, is it possible that you mean `not np.any(...)` instead, as all elements are supposed to be `False`?

sk1p (commented):
Good catch! I've added a comment about the test case, which I think can be improved. Thanks!

tests/test_masks.py
AnandBaburajan: Yup, that's what I meant. Thanks!

sk1p (approved):
Looks good to me, thanks!

PR #863, opened 2020-08-04
WIP: Display cluster conn details

twentyse7en:
If GUI is successfully connected to a remote cluster, the connection code will be
generated and an option to copy the code.  Fixes #845 

## Contributor Checklist:

* [x] I have added or updated my entry in [the creators.json file](https://github.com/LiberTEM/LiberTEM/blob/master/packaging/creators.json)
* [x] I have added [a changelog entry](https://github.com/LiberTEM/LiberTEM/tree/master/docs/source/changelog) for my contribution
* [x] I have added/updated documentation for all user-facing changes
* [ ] I have added/updated test cases
* [x] I have included the [rebuilt production build of the client](https://libertem.github.io/LiberTEM/contributing.html?#building-the-client) (only if
changes were made to the GUI)

PR #793, opened 2020-05-21
sync_offset, reshape datasets and get coords for UDFs

AnandBaburajan:
## Google Summer of Code '20 project

Fixes #256, fixes #553 and fixes #726 and

* The Python API and GUI now allows specifying `nav_shape` and `sig_shape` parameters to set a different shape
* `scan_size` and `detector_size` parameters are deprecated in favor of `nav_shape` and `sig_shape` respectively
* All datasets handle missing data gracefully
* The Python API and GUI now allow specifying a `sync_offset` to handle synchronization/acquisition problems
* Users can now access the coordinates of a tile/partition slice through UDFMeta.coordinates

## Todo

* https://github.com/LiberTEM/LiberTEM/issues/441#issuecomment-735218240
* #911

## Contributor Checklist:

* [x] I have added or updated my entry in [the creators.json file](https://github.com/LiberTEM/LiberTEM/blob/master/packaging/creators.json)
* [x] I have added [a changelog entry](https://github.com/LiberTEM/LiberTEM/tree/master/docs/source/changelog) for my contribution
* [x] I have added/updated documentation for all user-facing changes
* [x] I have added/updated test cases
* [x] I have included the [rebuilt production build of the client](https://libertem.github.io/LiberTEM/contributing.html?#building-the-client)

Issue #768, opened 2020-05-04
Parameter sidecar files


sk1p:
From #153:

SEQ files don't contain any information about the scan, so there was a question if we could read this from a "sidecar" file, like we do with other formats with their proprietary implementations. One idea would be to have a simple file, maybe in `.ini` format, or `.yml`, which contains that information. This idea could even be extended to other formats that need additional information, like raw. Something like this:

my_dataset.yml:

```yaml
format: SEQ
path: something.seq
scan_size: [256, 256]
dark_frame: dark.tif
```

my_other_dataset.yml:

```yaml
format: RAW
path: other.bin
dtype: u2
scan_size: [256, 256]
detector_size: [128, 128]
dark_frame: dark.tif
```

These yml files could then provide one-click opening in LiberTEM, even if the underlying raw dataset doesn't provide enough information.

sk1p:
Notes about the `dark_frame` parameter (and other corrections): in addition to the path to the file, there may be more parameters needed, for example when loading raw binary data, the `dtype` needs to be specified, or the array name when loading matlab files.

uellue:
Ah, that resonates with the discussion about specifying bad pixels, gain map, dark frame and also other parameters for cases where the original format doesn't contain that.

jhgee:
I want revive the discussion here and gather some ideas what should be achieved and what open questions are still around.

1. Is there already something similar like parameter sidecar files for the cluster class - https://github.com/LiberTEM/LiberTEM/blob/master/src/libertem/io/dataset/cluster.py#L63-L75 ?
2. Which format do we prefer? YAML sounds suitable. If we don't need a specific feature from one specific configuration file format I think it's best to aim for consistency across the LiberTEM project and choose a format accordingly

sk1p:
> 1. Is there already something similar like parameter sidecar files for the cluster class - https://github.com/LiberTEM/LiberTEM/blob/master/src/libertem/io/dataset/cluster.py#L63-L75 ?

Yeah, those are very specific for that use case. I'm also not sure if the cluster dataset will survive - for caching, we are now using fscache instead, and the cluster ds was mostly meant to be used as a backend for caching...

> 2. Which format do we prefer? YAML sounds suitable. If we don't need a specific feature from one specific configuration file format I think it's best to aim for consistency across the LiberTEM project and choose a format accordingly

I think the main candidates are: YAML, JSON, TOML. 

## JSON
Pro:
[...]

PR #873, opened 2020-09-07
MRC DataSet support

### LINKED ISSUE #152: Support for MRC format
Required by Lothar Houben, Benjamin Bammes <bbammes@directelectron.com>

Important standard format for soft mater / Cryo EM

sk1p:
Fixes #152.

/cc @jhgee

## Contributor Checklist:

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -13,3 +13,4 @@ client
 .dockerignore
 Dockerfile
 docker-compose.yml
+tmp
--- a/README.rst
+++ b/README.rst
@@ -115,6 +115,7 @@ information!
 - Stacks of Gatan DM3 and DM4 files (via `openNCEM <https://github.com/ercius/openNCEM>`_)
 - FRMS6 from PNDetector pnCCD cameras :cite:`Simson2015` (currently alpha, gain correction still needs UI changes)
 - FEI SER files (via `openNCEM <https://github.com/ercius/openNCEM>`_)
+- MRC (via `openNCEM <https://github.com/ercius/openNCEM>`_)
 - HDF5-based formats such as Hyperspy files, NeXus and EMD
 - Please contact us if you are interested in support for an additional format!
 
--- a/docs/source/changelog/features/mrc-support.rst
+++ b/docs/source/changelog/features/mrc-support.rst
@@ -1,0 +1,4 @@
+[Feature] MRC file format support
+=================================
+
+* Support for loading data from MRC files, as supported by ncempy (:issue:`152`, :pr:`873`).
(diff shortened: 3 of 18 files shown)


AnandBaburajan:
@sk1p I think this is the new format you mentioned earlier in my PR. If I'm correct, this PR would be merged only after #793 is merged, right? I'll look into this format asap.

sk1p:
> I think this is the new format you mentioned earlier in my PR. If I'm correct, this PR would be merged only after #793 is merged, right? I'll look into this format asap.

Haha, I'm actually not sure what format that was, but yeah, I think it would be good to have #793 merged, to enable GUI usage for 4D datasets.

The format is very similar to the raw format actually, as we can get a memory map for all data, meaning we can use the generic read ranges infrastructure etc.

> Edit: I just realised this isn't the format you were talking about:
> 
> > Btw. we will probably get support for another file format very soon, which works quite similarly, but should be simpler to understand than K2IS.

sk1p:
> Haha, I'm actually not sure what format that was

I think it was the TVIPS format, which also has some kind of "shutter active" flag.

sk1p:
@jhgee for the test data set you made available this seems to work fine - it seems to be inherently 3D (2D images with one "time" axis), is that correct? Are your production data sets also of this shape? Are they time series, or something like 4D STEM / pixelated STEM / scanning diffraction / ...?

If you have the time, could you test this PR with a "real" data set, also regarding performance? Thanks!

jhgee:
> @jhgee for the test data set you made available this seems to work fine - it seems to be inherently 3D (2D images with one "time" axis), is that correct? Are your production data sets also of this shape? Are they time series, or something like 4D STEM / pixelated STEM / scanning diffraction / ...?

@sk1p, you are right. Normally, this camera mode is used to save a movie but we use it to acquire a pixelated STEM data sets. Thus, it would be nice beeing able to specify the `scan_size` as an optional input.

I can ask the camera manufacturer to add more information to the mrc file but since they are now also offering a converter to hdf5 this might be of low priority.

sk1p:
> you are right. Normally, this camera mode is used to save a movie but we use it to acquire a pixelated STEM data sets. Thus, it would be nice beeing able to specify the `scan_size` as an optional input.

We will get the `scan_size` parameter basically for free (with a few small changes) once #793 is merged - including the `sync_offset` parameter to correct synchronization, if needed. 

> I can ask the camera manufacturer to add more information to the mrc file but since they are now also offering a converter to hdf5 this might be of low priority.

Yeah - I think it is fine having to specify the scanning parameters for now - this will get easier if we implement #768

sk1p:
I've marked `test_copy_notebook` as flaky for now, so work on other PRs can continue, discussion on this is happening in #863

PR #877, opened 2020-09-21
Support for loading stacks of 3D DM files

sk1p:
Before, we only supported loading stacks of 2D images, where each image
was in its own file. This adds support for the case where there is more
than one 2D image per file, stacking the images along the Z axis.

@AnandBaburajan I plan to merge this PR (and possibly also #873) before your GSoC PR is merged, would that be OK? It will probably result in a small conflict, that should be easy enough to resolve.

/cc @FENGSHAN95

## TODO

* [x] Update the `scan_size` check in `check_valid` 

### DIFF
--- a/docs/source/changelog/features/dm-stacks-of-3d.rst
+++ b/docs/source/changelog/features/dm-stacks-of-3d.rst
@@ -1,0 +1,4 @@
+[Feature] Support for loading stacks of 3D DM files
+===================================================
+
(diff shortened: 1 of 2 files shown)


AnandBaburajan:
> I plan to merge this PR (and possibly also #873) before your GSoC PR is merged, would that be OK?

:+1:

AnandBaburajan:
> Before, we only supported loading stacks of 2D images, where each image was in its own file. This adds support for the case  where there is more than one 2D image per file, stacking the images along the Z axis.

@sk1p I've fixed the `DM` issue at #793 (haven't pushed the code yet) so can you send me a `DM` dataset with 'more than one 2D image per file' to test this case? Thanks!

PR #885, opened 2020-10-08
Fix cluster code in notebook

twentyse7en:
Fix to generate TCP cluster connection in the notebook template.

## Contributor Checklist:

### DIFF
--- a/tests/template/test_ring_template.py
+++ b/tests/template/test_ring_template.py
@@ -53,20 +53,28 @@ def test_ring_default(hdf5_ds_2, tmpdir_factory, lt_ctx):
     )
 
 
+@pytest.fixture('function')
+def random_hdf5_1():
+    tmp_dir = '/data/temp_data/tmp'
+    os.makedirs(tmp_dir)
+    ds_path = os.path.join(tmp_dir, 'tmp_random_hdf5.h5')
+    ds = create_random_hdf5(ds_path)
+    yield ds
+    shutil.rmtree(tmp_dir, ignore_errors=True)
+
+
 @pytest.mark.dist
 @pytest.mark.asyncio
-def test_ring_tcp_cluster(lt_ctx):
+def test_ring_tcp_cluster(lt_ctx, random_hdf5_1):
 
     conn = {"connection": {
                     "type": "TCP",
                     "address": "tcp://scheduler:8786"
                     }
             }
-    base_path = '/data/temp_data'
-    tmp_dir = os.path.join(base_path, 'tmp')
-    os.makedirs(tmp_dir)
-    ds_path = os.path.join(tmp_dir, 'tmp_random_hdf5.h5')
-    ds = create_random_hdf5(ds_path)
+    ds = random_hdf5_1
+    ds_path = ds.path
+    tmp_dir = os.path.dirname(ds_path)
     dataset = _get_hdf5_params(ds_path)
 
     analysis = [{
@@ -101,5 +109,3 @@ def test_ring_tcp_cluster(lt_ctx):
         results,
         expected['intensity'].raw_data,
     )
-    # cleanup files after test
-    shutil.rmtree(tmp_dir, ignore_errors=True)
--- a/tests/template/test_ring_template.py
+++ b/tests/template/test_ring_template.py
@@ -1,10 +1,15 @@
 import io
 import os
+import shutil
 import nbformat
+import pytest
 import numpy as np
-from temp_utils import _get_hdf5_params
+from temp_utils import _get_hdf5_params, create_random_hdf5
 from libertem.web.notebook_generator.notebook_generator import notebook_generator
 from nbconvert.preprocessors import ExecutePreprocessor
+
+
+pytestmark = [pytest.mark.functional]
 
 
 def test_ring_default(hdf5_ds_2, tmpdir_factory, lt_ctx):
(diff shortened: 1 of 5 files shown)


tests/template/test_ring_template.py
```diff
@@ -46,3 +50,51 @@ def test_ring_default(hdf5_ds_2, tmpdir_factory, lt_ctx):
...
+def test_ring_tcp_cluster(hdf5_ds_2, tmpdir_factory, lt_ctx):
+    datadir = tmpdir_factory.mktemp('template_tests')
+
+    conn = {
+        'connection': {
+            'type': 'TCP',
+            'address': 'tcp://scheduler:8786',
+        }
+    }
+    path = hdf5_ds_2.path
```
twentyse7en, line 66: The test is failing because of the path, is there any difference for providing dataset path when using cluster in the test?

sk1p: Yes: the workers (and the tests themselves) are started in their own containers, and as such don't have access to each others file systems.

Currently, the `data` subdirectory is bind mounted into each container at `/data/`, such that you can put a variety of test data sets in there. Maybe you can create a temporary directory in the `data` directory, and use that path for these tests? Then it is available on each worker container.

twentyse7en: [Here](https://github.com/LiberTEM/LiberTEM/blob/16011aee5fab65fc2a4094ddb4e1a203f5d51b3e/conftest.py#L38) the dataset is created in the `/data/` right?

twentyse7en: > the data subdirectory is bind mounted into each container at /data/

Is this [dataset](https://github.com/LiberTEM/LiberTEM/blob/16011aee5fab65fc2a4094ddb4e1a203f5d51b3e/tests/io/datasets/test_blo.py#L17) also created in the `/data/` ? `test_blo` is currently being skipped on my local machine.

sk1p: > [Here](https://github.com/LiberTEM/LiberTEM/blob/16011aee5fab65fc2a4094ddb4e1a203f5d51b3e/conftest.py#L38) the dataset is created in the `/data/` right?

No - it is created in a temporary directory, with `data` somewhere in the name. So most of the time this will end up in `/tmp/`

sk1p: > > the data subdirectory is bind mounted into each container at /data/
> 
> Is this [dataset](https://github.com/LiberTEM/LiberTEM/blob/16011aee5fab65fc2a4094ddb4e1a203f5d51b3e/tests/io/datasets/test_blo.py#L17) also created in the `/data/` ? `test_blo` is currently being skipped on my local machine.

To be clear, the `blo` tests (and most other dataset tests, other than raw and hdf5) require the user to put specific, not yet publicly available, datasets at specific paths. It's not really possible to generate the data for these formats on the fly, as we need bit-for-bit matching input data, which can be error-prone. This is a bit painful, but that's how it is currently.

So, for these formats, there is the shared `/data/` bind mount available on the worker processes.

> Maybe you can create a temporary directory in the data directory, and use that path for these tests? Then it is available on each worker container.

Another possibility is to create a dedicated shared volume for these, mount it at `/shared-tmp-data/` or similar in all containers, and create the temporary test files there. That could be a better idea, as it means the `/data`/ mount can be treated as read-only.

twentyse7en:
Hey @sk1p, I'm not sure why the `test_hdf5.py` fails, It's working fine on my system.

sk1p:
> I'm not sure why the `test_hdf5.py` fails, It's working fine on my system.

I haven't seen that one before, either - it's possible this is caused by a new h5py version

sk1p:
Indeed, updating to h5py 3.0 I get the same error. Let me know if you want to investigate, or if I should have a look.

tests/template/test_ring_template.py
```diff
@@ -46,3 +51,55 @@ def test_ring_default(hdf5_ds_2, tmpdir_factory, lt_ctx):
...
+                            ro=8
+                        )
+    expected = lt_ctx.run(analysis)
+
+    assert np.allclose(
+        results,
+        expected['intensity'].raw_data,
+    )
+    # cleanup files after test
+    shutil.rmtree(tmp_dir, ignore_errors=True)
```
sk1p, line 105: This cleanup won't run if the test fails for some reason - it may be a good idea to create a fixture for this (it can even be local to the test file), something like this:


```
@pytest.fixture
def some_name():
    os.makedirs(...)
    ds = ... # create your data
    yield ds
    shutil.rmtree(tmp_dir, ignore_errors=True)
```

The same can be done by wrapping the whole test case into a `try...finally`, but that can make the code a bit harder to read.

twentyse7en:
> Let me know if you want to investigate, or if I should have a look.

Can you please look into it

sk1p:
> Can you please look into it

This is now reported upstream at https://github.com/h5py/h5py/issues/1740 - maybe until a fix is released, you could catch `SystemError` in `H5DataSet.detect_params`, in addition to `TimeoutError`?

twentyse7en:
> maybe until a fix is released, you could catch `SystemError` in `H5DataSet.detect_params`, in addition to `TimeoutError`?

Will do that :+1:

twentyse7en:
> `SystemError` in `H5DataSet.detect_params`, in addition to `TimeoutError`?

Test is still failing, should I also catch `SystemError` in `get_diagnostics` ?

sk1p:
> Test is still failing, should I also catch `SystemError` in `get_diagnostics` ?

Yeah, looks like that is needed, too.

sk1p (approved):
Looks good to me! @uellue do you want to look over it, and merge if you are happy with it, too?

uellue:
Looking good to go! Thx @twentyse7en, congrats! :-)

Just requires the rebuilt client, right?

uellue:
No client required after all! Thx everyone. :-)

Issue #867, opened 2020-08-17
Electronic labbook integration


sk1p:
The goal of this issue is to discuss/document how to implement an integration of a non-privileged version of LiberTEM into electronic labbooks and web-based data management platforms.

## Problem statement

Large data sets are managed by a labbook solution, but currently can't be easily visualized without downloading the data and loading it locally. It would be possible to just run the current LiberTEM version to allow interactive exploration of data, but that would allow access to the data of all users. What is needed is a solution to view a single data set in LiberTEM, without exposing any other data on the system.

## Possible solution

1) Add the concept of (user) sessions to the LiberTEM web API. The default start-up of a LiberTEM server would just create a single (privileged) session, and redirect the user there. In "labbook mode", the labbook would create user sessions via a privileged HTTP API, load one (or more) data sets into the session and mark the session as non-privileged, which would disable all web APIs other than creating analyses, handling jobs etc.

2) Add a "labbook" mode to the GUI, disabling the browse functionality and opening/closing data sets

3) Add some kind of token authentication to authenticate the labbook system with LiberTEM. The token would be sent with each request from the labbook to the LiberTEM server, for example as a header.

## Sequence of events

A typical sequence of events could look like this:

0) on startup, the labbook creates a cluster connection:

- `PUT /api/config/connection/`
  body: `{"type": "LOCAL", "workers": 6, ...}`
  headers: X-Auth-Header: XYZ

  note: Alternatively, this could also be done with a startup script, basically a customized replacement for the current `libertem-server` CLI.

1) the labbook calls an API to create a new user session, enabling
single-dataset mode, for every (user, dataset-id)

- `POST /api/session/`
  body: `{"type": "restricted"}`
  headers: X-Auth-Header: XYZ

  response: `{"status": "ok", "id": "123456"}`

2) the labbook loads a data set by sending over the parameters needed to open it

- `PUT /api/session/123456/datasets/9876543/`
  body: `{"dataset": {"id": 9876543, "params": {...}}}`
  headers: X-Auth-Header: XYZ

  note: The parameters are stored as a metadata item in the labbook, and converted into something that is understood by the `DataSet` classes.

3) the labbook then redirects the user to, for example,
`/libertem/?session=<uuid>` which tells the LiberTEM GUI to always send
the session specifier when performing API requests.

4) At the beginning, a configuration object is requested by the LiberTEM
GUI, which enables the single-dataset mode in the UI. On the server
[...]

PR #893, opened 2020-11-03
Electronic labbook integration

shabihsherjeel:
This is in reference to #867. 

I have added a middle-ware in base.py. And there are some changes in the way the server is configured. I am adding comments to the changes on github to better explain why I made those changes. 

I have only tested this behind an https nginx server with everything running in docker containers. And I have modified vital parts of the code. I tried to keep changes to the minimum. But there are a few things that might need more discussion in how to go about them. 

The middle-ware adds a random session ID if none are given. This should mean every browser connected to LiberTEM should have a separate shared state.

## Contributor Checklist:

changes were made to the GUI)


src/libertem/web/base.py
```diff
@@ -104,3 +106,21 @@ async def send_existing_job_results(self):
             await self.send_results(
                 result_set, job_id, analysis_id, details, finished=True
             )
+
+def handle_session(reqHandler, state, event_registry):
```
shabihsherjeel, line 110: This is a function I added to handle both the websocket and normal middle-ware. It checks if there is a SESSID cookie and sets up a random id if its not found.

src/libertem/web/base.py
```diff
@@ -104,3 +106,21 @@ async def send_existing_job_results(self):
...
+    if  reqHandler.get_secure_cookie("SESSID") is not None and reqHandler.get_secure_cookie("SESSID").decode("utf-8") in state.get_sessions():
+        reqHandler.state = state.get_sessions()[reqHandler.get_secure_cookie("SESSID").decode("utf-8")]["SessionState"]
+    else:
+        # Generate a random session id
+        id = secrets.token_hex(32)
+        reqHandler.state = state.register_session(id)
+        reqHandler.set_secure_cookie("SESSID", id, expires_days=None)
+    reqHandler.event_registry = reqHandler.state.event_registry
+
+class SessionsWebSocket:
```
shabihsherjeel, line 120: I override the initialization function for all the handlers throughout the server.

src/libertem/web/cli.py
```diff
@@ -13,16 +14,19 @@
               help='enable/disable opening the browser', default='True')
 @click.option('-l', '--log-level', help=f"set logging level. Default is 'info'. {log_values}",
               default='INFO')
+@click.option('--instance_type', type=click.Choice(["liberated", "restricted"]), default="liberated")
```
shabihsherjeel, line 17: This allows users to choose what kind of session they want to run the libertem server in. I have setup two. But I believe it would be better to add more "modes" that fill in the gaps for special use cases.

src/libertem/web/cli.py
```diff
@@ -13,16 +14,19 @@
...
+@click.option('--instance_config', type=str)
```
shabihsherjeel, line 18: Currently, the instance_config contains the details on how many workers and cuda devices we need to initiate the server with. This is to replace the connection form that LiberTEM currently offers when the user first connects.

src/libertem/web/connect.py
```diff
@@ -2,22 +2,32 @@
...
+
+from .base import ResultHandlerMixin, SessionsHandler, log_message
 from .messages import Message
-from .base import log_message, ResultHandlerMixin
 from .state import SharedState
-from libertem.utils.devices import detect
 
 log = logging.getLogger(__name__)
 
+def connect_to_local(state, connection, pool):
```
shabihsherjeel, line 15: I separate the connection code to call it from different places as required for the two modes we now have.

src/libertem/web/events.py
```diff
@@ -17,6 +16,9 @@ def add_handler(self, handler):
     def remove_handler(self, handler):
         self.handlers.remove(handler)
 
+    def get_number_of_handlers(self):
```
shabihsherjeel, line 19: I add this to see if we have connections to the server through the current session ID or not. This is to handle deleting the sessions once the user is done working.

src/libertem/web/events.py
```diff
@@ -64,3 +62,5 @@ async def open(self):
 
     def on_close(self):
         self.event_registry.remove_handler(self)
+        if self.event_registry.get_number_of_handlers() < 1: # Still isnt a perfect solution - refreshing a page also means close I guess
+            self.state.deregister()
```
shabihsherjeel, line 66: I added this so that whenever a user closes all connections, we get rid of the session state. But one issue I faced is that if there is only one tab open and the user refreshes it, then it will also delete the session and all changes made to the dataset. One solution I can think of is having a timer for a couple of seconds before the actual delete happens. So when the browser tab is refreshed, the reconnect should cancel the delete timer. What do you think?

src/libertem/web/messages.py
```diff
@@ -7,7 +6,7 @@ class Message(object):
     possible messages - the translation of our python datatypes to json types
     """
 
-    def __init__(self, state: SharedState):
+    def __init__(self, state):
         self.state = state
```
shabihsherjeel, line 10: I did this to get rid of a circular import issue. If you really want the SharedState just for typing, the whole server "base" file will need some rethinking/restructure. I guess that's better for you to handle as you're more experienced with the code base.

src/libertem/web/notebook_generator/notebook_generator.py
```diff
@@ -1,5 +1,7 @@
 import io
+
 import nbformat as nbf
+
 from .code_template import CodeTemplate
```
shabihsherjeel, line 6: These notebook_generator files have changed because I ran isort on everything in the web folder. I thought it doesn't hurt to isort this as well.

src/libertem/web/server.py
```diff
@@ -37,11 +41,8 @@ def get(self):
         self.render("client/index.html")
 
 
-def make_app(event_registry, shared_state):
-    settings = {
-        "static_path": os.path.join(os.path.dirname(__file__), "client"),
-    }
-    return tornado.web.Application([
+def make_app(event_registry, shared_state, instance_type, instance_config):
+    routes = [
```
shabihsherjeel, line 44: I added the routes into a list so we can modify and disable routes when in restricted mode.

src/libertem/web/server.py
```diff
@@ -100,7 +101,30 @@ def make_app(event_registry, shared_state):
             "state": shared_state,
             "event_registry": event_registry,
         }),
-    ], **settings)
+        (r"/api/session/", SessionHandler, {
```
shabihsherjeel: These are unused and I will remove them in the next commit. I had to deal with stashing client changes and docker testing changes every time I commit. That's why this slipped in here. Don't worry about these two routes. I don't use them.

src/libertem/web/server.py
```diff
@@ -100,7 +101,30 @@ def make_app(event_registry, shared_state):
...
+        (r"/api/session/([^/][0-9a-f-]+)/datasets/([^/][0-9a-f]+)/", SessionDatasetHandler, {
+            "state": shared_state,
+            "event_registry": event_registry,
+        }),
+    ]
+
+    shared_state.instance_type = instance_type
+    if instance_type == "restricted":
+        loop = asyncio.get_event_loop()
+        loop.run_until_complete(shared_state.set_local_executor(instance_config))
```
shabihsherjeel, line 108: This is where I set the config from the file in restricted mode. So the user does not get to choose how much resources he can use. That's done by the server admin.

src/libertem/web/server.py
```diff
@@ -100,7 +101,30 @@ def make_app(event_registry, shared_state):
...
+
+        routes_to_disable = ['/api/browse/localfs/']
+        for x in routes:
+            if x[0] in routes_to_disable:
+                routes.remove(x)
```
shabihsherjeel, line 113: A simple loop to allow disabling any other routes we will want to disable.

src/libertem/web/state.py
```diff
@@ -302,6 +327,46 @@ def __init__(self):
 
         self.dataset_for_job = {}
         self.local_directory = "dask-worker-space"
+        self.event_registry = EventRegistry() #Do we need this?
```
shabihsherjeel, line 330: I needed this. I don't know what the discussion behind this is.

sk1p:
Thanks for your work @shabihsherjeel! I've done a first quick read through the code and I'll try to finish my review beginning of next week. Thank you also for the comments in the code, this makes the review much easier :+1:

One question from my first read through the code: how does the labbook server inject the data set parameters? Does it just use the `DataSetDetailHandler`? If so, what stops a user from opening an arbitrary data set? Could you maybe add a small script (prototypes folder?) to demonstrate the labbook <> libertem interaction?

I see that you have methods `{get,add}_allowed_directories` (unused for now?) - we need to be a bit careful with this. This assumes directories are "access boundaries" - so if a user puts a bunch of data into a directory, sharing this via a labbook becomes an all-or-nothing decision.

Also, imagine a user having access to a directory they can write to (via NFS or similar), they can then create symlinks to data of other users which they shouldn't be able to access. This can be nuanced in practice - for example, the opening path specified for some formats is just a header file, whereas the data set consists of multiple data files. So if we only check that the real opening path (with symlinks resolved) is in one of the allowed directories, the data files can still be symlinks which resolve to other user's data.

Maybe we should formalize a minimal threat model, to define the parameters on how LiberTEM can be run securely, and what the security boundaries are?

uellue:
> Maybe we should formalize a minimal threat model, to define the parameters on how LiberTEM can be run securely, and what the security boundaries are?

Sounds good! I've never done a threat model, so I'll just try to follow https://en.wikipedia.org/wiki/Threat_model

Model of the system:

The connection between browser and ElabFTW as well as LiberTEM GUI is SSL encrypted, excluding tampering and eavesdropping on the underlying
 transport layer, right?

* Clicking the "Load in LiberTEM" button in ElabFTW points the user to a LiberTEM instance. The button is not a hyperlink in the classical sense, but seems to trigger some kind of script?
  * Can a user spoof the action (request/JavaScirpt call/...) that triggers opening a LiberTEM window for a file/dataset in ElabFTW without having regular access to that dataset in ElabFTW?
  * Can a user create datasets in ELabFTW that point to arbitrary/unauthorized locations on the backend system?
* A new tab with running LiberTEM opens, having the dataset loaded
  * Which steps are performed when opening a dataset between ElabFTW and LiberTEM? Can the request to open a dataset be spoofed on the client side?
* The LiberTEM web GUI seems to work normally, just the buttons to open new datasets seem to be missing
  * What happens if the user spoofs the request(s) to the LiberTEM GUI to open a dataset, mimicking the action of the usual buttons?
  * Can the web GUI of LiberTEM be used to inject code that runs on the server?
* The LiberTEM GUI seems to load, just without any open dataset, if one points a browser to the target URL.
  * Can the exposed web API be used to perform any actions or disclose information when opened from outside without redirection from ELabFTW?
    * **Yes, it seems so.** I just performed a DoS by pressing the "Shutdown" button from a second browser (no ElabFTW based cookies etc). Clicking on "Load in LiberTEM" in the primary browser doesn't work anymore.

uellue:
Perhaps requests to the sandboxed LiberTEM should contain some form of session ID that is somehow generated from the ElabFTW session, and interaction should be completely disabled for unauthenticated users? That would already make exploits from the open internet a lot harder.

@sk1p, if I understand correctly the security model for JupyterHub  integration is based on LiberTEM running as the Unix user that is logged in to JupyterLab, right? That means until now we had the approach "if the OS lets me do it, it is probably OK". To be honest, I don't have a good feeling how to manage LiberTEM running as a kind of system user shared between different end users.

sk1p:
> Model of the system:
> 
> The connection between browser and ElabFTW as well as LiberTEM GUI is SSL encrypted, excluding tampering and eavesdropping on the underlying
> transport layer, right?

Agreed.

A threat model should also ask and answer questions like these:

- What privilege boundaries exist or should exist? 
  - Example: access restrictions in the labbook should imply access restrictions in LiberTEM
  - Example: in restricted mode, the user should not be able to shut down the LiberTEM server → there should be a boundary between privileged and user API accesses
- Should a single LiberTEM process have read access to data across permission boundaries? 
- Can we trust the parameters for opening data sets, or do we need to take extra care that they are correct? Can a user elevate privileges by manipulating parameters?
- What even is privilege elevation in this case? Remote code execution, but also information disclosure
  - Related: should access via LiberTEM imply that users should be able to access the raw data? (regarding the point analysis plus result download mentioned above, or equivalently just picking all frames individually...)

FYI @shabihsherjeel and I had a call last week about this, and discussed that it would make sense to have privilege separation between the API calls performed by elabFTW (creating a session, opening a data set, shutting down the LiberTEM server, ...) and by the user in the session (running analyses etc.). The elevated calls by elabFTW would be protected by an API token / shared secret that is put into both the LiberTEM server config and the elabFTW integration / plugin. Sorry for not updating here!

> * Clicking the "Load in LiberTEM" button in ElabFTW points the user to a LiberTEM instance. The button is not a hyperlink in the classical sense, but seems to trigger some kind of script?
>   
>   * Can a user spoof the action (request/JavaScript call/...) that triggers opening a LiberTEM window for a file/dataset in ElabFTW without having regular access to that dataset in ElabFTW?

This will probably change soon, to create the session from the server-side of elabftw, see note above.

>   * Can a user create datasets in ELabFTW that point to arbitrary/unauthorized locations on the backend system?

That's a good point, but I think we should draw a boundary here: LiberTEM cannot know this, only the ElabFTW knows what an authorized location would look like. Of course we can try to create some defense in depth (like checking data locations), but this can often be circumvented (see: symlink attacks).

> * A new tab with running LiberTEM opens, having the dataset loaded
>   
>   * Which steps are performed when opening a dataset between ElabFTW and LiberTEM? Can the request to open a dataset be spoofed on the client side?

I think the current implementation relies on the semi-random dataset filenames, but that is subject to change (see above).

>   * Can deeper information about the system be gained from an open dataset?
>     * Possibly the "Dataset info" and the new "Download notebook" can disclose internal storage locations.

Yeah, the diagnostics API provides some information that will contain absolute paths - I would argue it's not a severe issue if LiberTEM is running in a docker container with the data bind-mounted to some generic path.

> * The LiberTEM web GUI seems to work normally, just the buttons to open new datasets seem to be missing
>   
>   * What happens if the user spoofs the request(s) to the LiberTEM GUI to open a dataset, mimicking the action of the usual buttons?
>   * Can the web GUI of LiberTEM be used to inject code that runs on the server? Good candidates would be dataset parameters?

This is related to the question: how trustworthy are the data sets uploaded to the labbook? I can imagine that our readers contain bugs that, in case of numba-based decoding, could result in remote code execution. We should be relatively robust against rogue dataset parameters, 

>   * The "Download result" feature in combination with a collection of "point analyses" and crafted dataset parameters can be used to exfiltrate arbitrary data from the system, as soon as the LiberTEM GUI opens it as a dataset.

I think the key word here is "crafted dataset parameters" - untrusted API calls must not be allowed to open data sets.

>   * Can a malicious website somehow inject requests into the LiberTEM GUI (XSS)?

React makes it hard to inject raw markup and thus scripts - you have to use the aptly named [dangerouslySetInnerHTML](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml) function, which can be audited against, or you have to go out of your way and access raw DOM nodes.

We don't generate any dynamic HTML on the server side. 

Together, these two things should make XSS in principle less likely. In general, it depends on React not being buggy, and changes to the GUI being properly reviewed.

>   * What happens if a user sets parameters that overload the target system (memory, CPU)? In particular clustering analysis has the potential to set large number of reference points and clusters. Opening a large RAW file with parameters for very small signal or navigation dimension could do the same.

This is a good point - the first part relating to resource sharing and throttling by viewers, the second relating to abuse from the side of data uploaders.

> * The LiberTEM GUI seems to load, just without any open dataset, if one points a browser to the target URL.

This will change when the session is created by the elabFTW backend - then, in restricted mode, it should be an error if you visit without a valid session id set.

>   * Can the exposed web API be used to perform any actions or disclose information when opened from outside without redirection from ELabFTW?
>     
>     * **Yes, it seems so.** I just performed a DoS by pressing the "Shutdown" button from a second browser (no ElabFTW based cookies etc). Clicking on "Load in LiberTEM" in the primary browser doesn't work anymore.
> * What happens if/when the LiberTEM GUI is updated? How do we make sure it doesn't introduce new interfaces that are harmless in the usual "per user" case, but can be exploited in the ElabFTW integration scenario?
> * LiberTEM uses Dask.Distributed
>   
>   * Is the Dask.Distributed scheduler exposed somehow?

When deploying LiberTEM, there should be a network barrier between the dask scheduler and the external network.

>   * The "Download Notebook" feature gives information about any external Dask scheduler used.

We have to decide if we consider this privileged information and want to disable it in restricted mode

sk1p:
> Perhaps requests to the sandboxed LiberTEM should contain some form of session ID that is somehow generated from the ElabFTW session, and interaction should be completely disabled for unauthenticated users? That would already make exploits from the open internet a lot harder.

Indeed, see above

> if I understand correctly the security model for JupyterHub integration is based on LiberTEM running as the Unix user that is logged in to JupyterLab, right? That means until now we had the approach "if the OS lets me do it, it is probably OK".

Yeah, JupyterLab integration is a different beast - it means:

- You can launch a libertem instance per user, because either there is such a thing as a unix user for the jupyter user, or the container/vm is "owned' by the jupyter user
- Typically in a jupyterlab instance, there should be enough RAM available that we have enough resources for starting per-user LiberTEM instances
- The user already has full access to the context where we are launching the libertem instance, so there are no privilege boundaries to cross
- JupyterLab takes care of authorization/authentication, similar to the proposed session system

>To be honest, I don't have a good feeling how to manage LiberTEM running as a kind of system user shared between different end users.

I think it can work - although it could be possible that the result of our discussion is that we only recommend deploying LiberTEM in a scenario where the data is open access, or "flat" access, such that anyone who can access any data, can access all data. This should match the use case of @shabihsherjeel (please correct me if I'm wrong)

shabihsherjeel:
Sorry for the delayed response. I thought of the points you put out. I decided to try another approach. I will have a middleware process launcher that will control Linux file access for different client instances of LiberTEM. I am planning to have a separate GID for every data file I wish to provide an instance of LiberTEM access to. And the file permissions as default for all the unnecessary files and other data files would be no read, write, or execute. 

The same security issues will exist on the middleware as well. And I will setup privileged APIs for lab notebooks to spawn a LiberTEM process. But this means that LiberTEM will not have to handle creating a "jail" for different instances.

This is still work in progress. I have tested a very basic version of this. I will test out a few details and also design how I will manage concurrent users on the same file. Then I will update this thread with more info.

The only changes I am planning to do now to LiberTEM are to set default cluster configuration and dataset to load through the CLI. All those session details will be handled by a separate middleware process spawner. This way LiberTEM will not need to worry about being in a lab notebook mode and cause users confusion as this is more of a client side application.

I will love to discuss this more with you after I have a basic implementation up and running.

sk1p:
> I thought of the points you put out. I decided to try another approach. I will have a middleware process launcher that will control Linux file access for different client instances of LiberTEM. I am planning to have a separate GID for every data file I wish to provide an instance of LiberTEM access to. And the file permissions as default for all the unnecessary files and other data files would be no read, write, or execute.
> 
> The same security issues will exist on the middleware as well. And I will setup privileged APIs for lab notebooks to spawn a LiberTEM process. But this means that LiberTEM will not have to handle creating a "jail" for different instances.
> 
> This is still work in progress. I have tested a very basic version of this. I will test out a few details and also design how I will manage concurrent users on the same file. Then I will update this thread with more info.
> 
> The only changes I am planning to do now to LiberTEM are to set default cluster configuration and dataset to load through the CLI. All those session details will be handled by a separate middleware process spawner. This way LiberTEM will not need to worry about being in a lab notebook mode and cause users confusion as this is more of a client side application.

Thanks for the update! This also sounds like a good approach, and a bit less complexity on the side of LiberTEM, delegating the security issues to the OS. Some questions about this:

- Will data loading / browsing also be disabled if a dataset was loaded through the CLI?
- I think the plan was to terminate LiberTEM after some inactivity. How does this work with the middleware? Does the `libertem-server` just quit on its own after some time of inactivity?
- I'm not so clear about the GID scheme, I have to think a bit about this. But maybe as an idea of an alternative approach, what about launching the `libertem-server` instances in a container, and r/o bind-mount the file(s) that are needed in there? Then the files only need to be read-accessible by the user that LiberTEM is running as in the container, or LiberTEM could even be run as the UID/GID of the specific files.

I would be happy if the loading via CLI could use something in the direction of #768, if that works for you.

> I will love to discuss this more with you after I have a basic implementation up and running.

Sounds good!

shabihsherjeel:
> Thanks for the update! This also sounds like a good approach, and a bit less complexity on the side of LiberTEM, delegating the security issues to the OS. Some questions about this:
> 
> * Will data loading / browsing also be disabled if a dataset was loaded through the CLI?

I am thinking of leaving that intact. I can see how it works out. But currently, I am managing file access on the OS level so even if you browse you can't open the other datasets.

> * I think the plan was to terminate LiberTEM after some inactivity. How does this work with the middleware? Does the `libertem-server` just quit on its own after some time of inactivity?

I am not at that stage of my implementation yet. But as the process will be managed by the middleware, having a time limit on the process is easy. I need to figure out how the middleware will know if the user still has the tab open. Currently, I am planning on using nginx as a proxy with directives to dynamically add routes for different instances. I guess I could monitor connections from the user's IP or somehow inject a simple script that tells the middleware when the browser tab has been closed.

> * I'm not so clear about the GID scheme, I have to think a bit about this. But maybe as an idea of an alternative approach, what about launching the `libertem-server` instances in a container, and r/o bind-mount the file(s) that are needed in there? Then the files only need to be read-accessible by the user that LiberTEM is running as in the container, or LiberTEM could even be run as the UID/GID of the specific files.

Currently, the way I have it setup is that whenever I want to give a new user/instance access to that file, I chgrp on the file and give that process the same GID. This gives me a bit of liberty in controlling access than the UID would.

As of now, my implementation runs in a docker instance with all the datasets mounted with read/write access to control their permissions through a tornado server that just launches processes within a single container. The reasons I am using docker containers and do not want to go with launching a separate container for every instance are the following:
- Ease of setup for users already deploying ElabFTW as docker containers. 
- Isolation of the system in a container to add another security layer. (I don't think this is a great argument. But it helps with the infancy of the implementation.)
- Less overhead every time a user wants to run a Libertem instance
- If I want to spawn docker containers from my main docker container, I will have to give this main container access to the system docker daemon. And this doesn't feel secure to me.

Please let me know what you think of this.

> 
> I would be happy if the loading via CLI could use something in the direction of #768, if that works for you.
> 

This is great for the actual loading of files. But I only implemented pre-loading the cluster info through the CLI. I do not mind having this too. It is a great idea.

> > I will love to discuss this more with you after I have a basic implementation up and running.
> 
> Sounds good!

sk1p:
> I am thinking of leaving that intact. I can see how it works out. But currently, I am managing file access on the OS level so even if you browse you can't open the other datasets.

Okay, sounds good.

> But as the process will be managed by the middleware, having a time limit on the process is easy. I need to figure out how the middleware will know if the user still has the tab open. Currently, I am planning on using nginx as a proxy with directives to dynamically add routes for different instances. I guess I could monitor connections from the user's IP or somehow inject a simple script that tells the middleware when the browser tab has been closed.

Sounds like a solution. Did you already find a way to implement the dynamic configuration? There seems to be [a module](https://nginx.org/en/docs/http/ngx_http_api_module.html) for dynamic configuration, but that's only available in the commercial version of nginx.

If nginx doesn't work out, there are also some alternatives like [traefik](https://doc.traefik.io/traefik/operations/api/), [caddy](https://caddyserver.com/docs/api-tutorial), [haproxy](https://www.haproxy.com/documentation/hapee/1-9r1/reference/dataplaneapi/) etc. that should also have the possibility to create configuration via API.

> Currently, the way I have it setup is that whenever I want to give a new user/instance access to that file, I chgrp on the file and give that process the same GID. This gives me a bit of liberty in controlling access than the UID would.

Okay - maybe we can later adapt this to different setups, as the data access will probably be different in each setup.

> As of now, my implementation runs in a docker instance with all the datasets mounted with read/write access to control their permissions through a tornado server that just launches processes within a single container. The reasons I am using docker containers and do not want to go with launching a separate container for every instance are the following:
> - Ease of setup for users already deploying ElabFTW as docker containers.
> - Isolation of the system in a container to add another security layer. (I don't think this is a great argument. But it helps with the infancy of the implementation.)
> - Less overhead every time a user wants to run a Libertem instance
> - If I want to spawn docker containers from my main docker container, I will have to give this main container access to the system docker daemon. And this doesn't feel secure to me.

True, especially the last point: access to the system docker daemon is equivalent to root access on the system (example: an attacker can spawn a new container with a bind mount of /etc/shadow and /etc/passwd and add a new root user for themselves). So the middleware would have to be regarded the same as running as root on the host system itself.

uellue:
@shabihsherjeel just re-assessing older PRs. Is this still relevant or is it superseded by token-based auth? Are there other ways to address the integration? :-)

shabihsherjeel:
This has indeed been superseded by a token-based auth. Now that Libertem supports auth with JupyterLab, I can work with that instead. I'm closing this pull request. I hope to have the new integration ready soon. :)

PR #898, opened 2020-11-06
Lazily import some dependencies

sk1p:
Mostly this prevents importing pytorch, matplotlib, skimage, sklearn at
import time. Local/lazy imports should still perform well (some ns), as they
reduce to something like a dict lookup.

This improves initial import time, measured as

    python -c 'import libertem.api'

by a factor of two. Running some of our slowest tests (tests/executor
and tests/server) is reduced from 264s to about 220s on my workstation. Let's see how the
CI time is improved, hopefully there the impact is even larger.

Something like the following was used to profile import times:

```
PYTHONPROFILEIMPORTTIME=1 python -c 'import libertem.api' 2>&1 | sort -k 3 -n
```

@uellue thoughts?

## TODO

- [ ] measure impact of `import torch` in `ApplyMasksUDF` - does it take time in the profile? Maybe move to `get_task_data`?

## Contributor Checklist:

changes were made to the GUI)

### DIFF
--- a/src/libertem/udf/masks.py
+++ b/src/libertem/udf/masks.py
@@ -1,7 +1,3 @@
-try:
-    import torch
-except ImportError:
-    torch = None
 import numpy as np
 
 from libertem.udf import UDF
@@ -117,6 +113,10 @@ class ApplyMasksUDF(UDF):
 
     def get_task_data(self):
         ''
+        try:
+            import torch
+        except ImportError:
+            torch = None
         m = self.meta
         use_torch = self.params.use_torch
         if (torch is None or m.input_dtype.kind != 'f' or m.input_dtype != self.get_mask_dtype()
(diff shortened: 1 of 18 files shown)


uellue:
Nice! Reducing the ever-growing startup time is definitely a good idea. :-) Good to go from my perspective!

sk1p:
Comparing https://travis-ci.com/github/LiberTEM/LiberTEM/builds/198721103 and https://travis-ci.com/github/LiberTEM/LiberTEM/builds/198751228, it seems to shave off 1m to 1m30s from our Linux builds, 3-5 mins from the  Mac OS builds (don't know how much variance there usually is...)

src/libertem/udf/masks.py
```diff
@@ -144,10 +144,7 @@ def process_tile(self, tile):
         ''
         flat_data = tile.reshape((tile.shape[0], -1))
         if self.task_data.use_torch:
-            try:
```
uellue: The purpose of `try ... except` is to work without installed torch since that is an optional dependency.

uellue: Is that still required if use_torch is set, or is use_torch only set if torch can be imported?

sk1p: True, but when we enter that branch, `use_torch` is `True`, which is only the case if the import in `get_task_data` worked

PR #902, opened 2020-11-17
Fix AUX data failure with single frame partitions

### LINKED ISSUE #791: Dataset dimension squeezed out
How to reproduce:

Process data with nav shape (1,), possibly with aux data

Observed behavior:
```
c:\users\weber\documents\libertem\libertem\src\libertem\common\buffers.py in set_buffer(self, buf, is_global)
    503             extra = (1,)
    504         shape = (-1,) + extra
--> 505         self._data = buf.reshape(shape).squeeze()
    506         self._data_coords_global = is_global
    507 

ValueError: cannot reshape array of size 2 into shape (49,2)
```
squeeze() is likely squeezing out the ds dimension.

uellue:
If a partition consists of a single "sig" item, aux data shapes don't match. The singular dimension is "swallowed" somehow, apparently. I tried to find out where, but didn't get far.

This is mostly an issue for testing since normally partitions contain more than one sig element. For memory partitions, the "num_partitions" parameter can be set in such a way that no partitions with singular nav shape are created as a work-around.

Fixes #791 

## Contributor Checklist:

### DIFF
--- a/docs/source/changelog/bugfix/aux_failure_single_frame_part.rst
+++ b/docs/source/changelog/bugfix/aux_failure_single_frame_part.rst
@@ -1,0 +1,4 @@
+[Bugfix] Handle single-frame partitions in combination with aux data
+====================================================================
+
+* Instead of squeezing the aux buffer, reshape to the correct shape (:issue:`791`, :pr:`902`)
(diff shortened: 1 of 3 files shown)


sk1p:
> If a partition consists of a single "sig" item, aux data shapes don't match. The singular dimension is "swallowed" somehow, apparently. I tried to find out where, but didn't get far.

As far as I remember, there is a `np.squeeze` call somewhere, that should be converted to an explicit reshape. Maybe in the `BufferWrapper` code, or related to that?

sk1p:
I think I have a fix for this, can I take over your branch?

uellue:
Yes sure, go ahead! :-)

sk1p:
This should be ready to merge once CI finishes

PR #906, opened 2020-11-27
Add pipeline with data access

### LINKED ISSUE #86: Test files for datasets

sk1p:
Fixes #86. The test data is provided as a bind mount at `/data/` (currently pointing to a directory owned by my user, where I put the private test data to mirror my local setup, but that can be changed).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ data
 *.idea
 .mypy_cache/
 .benchmarks
+junit.xml
--- a/azure-pipelines-data.yml
+++ b/azure-pipelines-data.yml
@@ -6,41 +6,21 @@ stages:
     jobs:
     - job: data_tests
       pool: DataAccess
-      strategy:
-        matrix:
-          Python36:
-            python.version: '3.6'
-            TOXENV: 'py36'
-          Python37:
-            python.version: '3.7'
-            TOXENV: 'py37'
-          Python38:
-            python.version: '3.8'
-            TOXENV: 'py38'
       variables:
-        TOXENV: '$(TOXENV)'
+        TOXENV: 'py36'
         LT_TEST_DATA_PATH: '/data/'
       steps:
-      - task: UsePythonVersion@0
-        displayName: 'Use Python $(python.version)'
-        inputs:
-          versionSpec: '$(python.version)'
+      - bash: python3 -m venv venv
+        displayName: 'create venv'
+    
+      - bash: ./venv/bin/pip install tox
+        displayName: 'install tox'
 
-      - bash: pip install -U tox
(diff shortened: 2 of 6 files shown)


sk1p:
I'm going to merge this now, to test if the build still doesn't run on `master` - @uellue please let me know if you are happy with the comment-based trigger for running I/O tests.

uellue:
The comment-based method is supposed to prevent running arbitrary code on our agent via a PR, right? How is this authenticated and authorized? Can I trigger this, too?

I tried to access the Azure Pipelines details of LiberTEM, but couldn't:

```
{"$id":"1","innerException":null,"message":"TF400813: The user 'Windows Live ID\\d.weber@materials-scientist.com' is not authorized to access this resource.","typeName":"Microsoft.TeamFoundation.Framework.Server.UnauthorizedRequestException, Microsoft.TeamFoundation.Framework.Server","typeKey":"UnauthorizedRequestException","errorCode":0,"eventId":3000}
```

sk1p:
> The comment-based method is supposed to prevent running arbitrary code on our agent via a PR, right?

Right.

> How is this authenticated and authorized?

Authentication: the comment metadata includes the github username, so your login into github is the authentication

Authorization: the comment functionality is limited to team members, which are defined as users with Write permissions. The user [needs to set his membership to public](https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#run-pull-request-validation-only-when-authorized-by-your-team) though.

> I tried to access the Azure Pipelines details of LiberTEM, but couldn't: [...]

I'll add you as an admin to the Azure DevOps organization. The project is set to public visibility, but I guess that doesn't make all settings public, too.

sk1p:
I think your comment above didn't trigger a pipeline run, because the branch was already deleted. Could you re-try on #912?

uellue:
> I think your comment above didn't trigger a pipeline run, because the branch was already deleted. Could you re-try on #912?

Works! 👍 😄

sk1p:
Great! :+1:

PR #919, opened 2020-12-07
Check parameter `executor` in api.Context; fix tests

uellue:
Give a meaningful error message if a bad value is passed as an executor.

Refs #918 

Drive-by change: Fix tests to not lock up when I run them on moellenstedt with remote VSCode. Not sure if the lock-ups occur under other circumstances as well.

## Contributor Checklist:

* [x] I have added or updated my entry in [the creators.json file](https://github.com/LiberTEM/LiberTEM/blob/master/packaging/creators.json)
* [ ] I have added [a changelog entry](https://github.com/LiberTEM/LiberTEM/tree/master/docs/source/changelog) for my contribution
* [ ] I have added/updated documentation for all user-facing changes
* [x] I have added/updated test cases

## Reviewer Checklist:

* [x] `/azp run libertem.libertem-data` passed

PR #684, opened 2020-03-17
Limit to distributed < 2.9

uellue:
The notebook prototypes/reproducers/test-dask-remote.ipynb
reliably locked up or errored when using it with a remote dask
cluster of version 2.9.0 or newer.

Tested on moellenstedt in a new clean conda environment with LiberTEM
master, Python 3.7.

@sk1p could you perhaps test if you can reproduce it locally on your machine? I couldn't reproduce it on my Windows machine.

Issue #61, opened 2018-08-10
CPU load from libertem-server when idle


uellue:
Observed behavior:
The libertem-server process creates significant CPU load on the moellenstedt machine even when no processing is performed and the GUI window is closed. The load seems to increase with the number of workers -- 4 cores on @sk1p 's machine not noticeable, 18 cores on moellenstedt significant.

Expected behavior:
LiberTEM only creates CPU load when it is doing useful work.

Suspicion:
The dask scheduler might be the culprit (@sk1p )

sk1p:
See also: https://github.com/dask/distributed/issues/2079

uellue:
I tested with distributed==1.25.0 released about two weeks ago, and libertem-server still creates significant background load (one process with about 10 % CPU usage) on moellenstedt as soon as the cluster is running. It produces the same load with distributed==1.24.0 which was released before merging dask/distributed#2339.

In conclusion, I'd say that this change didn't improve things for us. 10 % still feels like a bit much for an idling cluster.

sk1p:
There is actually a new workaround posted on the dd issue: https://github.com/dask/distributed/issues/2156#issuecomment-503735503 - we just have to try it and then figure out the right way to set the configuration. It can be done with [dask.config.set](https://docs.dask.org/en/latest/configuration.html#directly-within-python), but it kind of feels dirty to just globally set the dask config when creating an executor (or when starting a local cluster?)

sk1p:
I just re-checked the idle CPU usage. Just running `libertem-server` and creating a local cluster creates 15-20% idle usage on the main process, and additionally 2-3% per worker process.

Setting all four environment variables as in #921 reduces idle load to about 12% in the main process, and on average no load at all per worker process. 

```sh
DASK_DISTRIBUTED__ADMIN__TICK__INTERVAL=1000
DASK_DISTRIBUTED__SCHEDULER__WORK_STEALING_INTERVAL=1000
DASK_DISTRIBUTED__WORKER__PROFILE__CYCLE=60000
DASK_DISTRIBUTED__WORKER__PROFILE__INTERVAL=30000
```

uellue:
Checking on Ptycho, the idle load with default settings is at 10 % for the main process and about 1 % for the workers. The environment variables above lower it to 6 % for main and 0.3-0.7 % for workers.

On a quad-core Windows machine the load is very low as well.

Closing since this doesn't seem to be a big issue anymore.

PR #921, opened 2020-12-10
Use the right loop in dd.as_completed

sk1p:
This is needed if there are multiple distributed `Client`s running in the same process, as the `Client` that was created last is registered as the global `Client` and then used for functions like `dd.as_completed`. That means we create futures on a different loop than we try to retrieve their results, which results in hangs (can be debugged using `PYTHONASYNCIODEBUG=1`), as setting the result for a future is not threadsafe.

As a drive-by fix, workers are now started in parallel in `libertem-worker`.

Refs #919 #684 #482

## Contributor Checklist:

changes were made to the GUI)

## Reviewer Checklist:

### DIFF
--- a/docs/source/tips.rst
+++ b/docs/source/tips.rst
@@ -197,3 +197,25 @@ extra info as well to aid later data evaluation. Note that the compilation tests
 will have poor statistics since it only runs once. If you have an idea on how to
 collect better statistics, please `let us know
 <https://github.com/LiberTEM/LiberTEM/issues/new>`_!
+
+
+Simulating slow systems with control groups
+-------------------------------------------
+
+Under Linux, it is possible to simulate a slow system using control groups:
+
+.. code-block:: shell
+
+    sudo cgcreate -g cpu:/slow
+    sudo cgset -r cpu.cfs_period_us=1000000 slow
+    sudo cgset -r cpu.cfs_quota_us=200000 slow
+    sudo chown root:<yourgroup> /sys/fs/cgroup/cpu,cpuacct/slow
+    sudo chmod 664 /sys/fs/cgroup/cpu,cpuacct/slow
+
+Then, as a user, you can use :code:`cgexec` to run a command in that control group:
+
+.. code-block:: shell
+
+    cgexec -g cpu:slow pytest tests/
+
+This is useful, for example, to debug test failures that only seem to happen in CI.
--- a/tests/io/test_cached.py
+++ b/tests/io/test_cached.py
@@ -23,12 +23,16 @@ def default_cached_ds(tmpdir_factory, default_raw, lt_ctx):
     yield ds
 
 
-# A running Dask client can introduce a timing issue
-# between automatic closing of a numpy.memmap object and renaming
-# the underlying file
-def test_start_client():
+@pytest.fixture(scope='module', autouse=True)
+def client_in_background():
+    # A running Dask client can introduce a timing issue
+    # between automatic closing of a numpy.memmap object and renaming
+    # the underlying file
     with dd.LocalCluster() as cluster:
         client = dd.Client(cluster, set_as_default=False)
+        yield
+        # to fix "distributed.client - ERROR - Failed to reconnect to scheduler after 10.00 seconds, closing client"  # NOQA
+        client.close()
 
 
 def test_simple(default_cached_ds):
--- a/tox.ini
+++ b/tox.ini
@@ -1,9 +1,9 @@
 [tox]
-envlist = flake8, py{36,37,38,39}, benchmark, benchmark-cuda{101,102}, mypy
+envlist = flake8, py{36,37,38,39}, benchmark, benchmark-cuda{101,102,111}, mypy
 
 [testenv]
 commands=
-    pytest --durations=10 --cov=libertem --cov-report=term --cov-report=html --cov-report=xml --junitxml=junit.xml {posargs:tests/}
+    pytest -s -vvv --durations=10 --cov=libertem --cov-report=term --cov-report=html --cov-report=xml --junitxml=junit.xml {posargs:tests/}
     # win_tweaks.py depends on modules that are only available on Windows
     pytest --doctest-modules --ignore=src/libertem/win_tweaks.py src/libertem/
 deps=
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = flake8, py{36,37,38,39}, benchmark, benchmark-cuda{101,102,111}, mypy
+envlist = flake8, py{36,37,38,39}, benchmark, benchmark-cuda{101,102,110}, mypy
 
 [testenv]
 commands=
@@ -12,6 +12,7 @@ deps=
     -roverride_requirements.txt
     cuda101: cupy-cuda101
     cuda102: cupy-cuda102
+    cuda110: cupy-cuda110
 extras=
     hdbscan
 setenv=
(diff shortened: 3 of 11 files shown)


sk1p:
@uellue could you have a look at [the failure of `test_tileshape_adjustment_fuzz`](https://dev.azure.com/LiberTEM/LiberTEM/_build/results?buildId=110&view=logs&j=9d73d1fc-3a9e-5ba7-6ad9-e21fc4cd7ca4&t=166bcb7a-b1a3-550e-b07c-c68401b8ea86)? Looks like sig shape of `(1, ...)` is the problem here?

uellue:
> @uellue could you have a look at [the failure of `test_tileshape_adjustment_fuzz`](https://dev.azure.com/LiberTEM/LiberTEM/_build/results?buildId=110&view=logs&j=9d73d1fc-3a9e-5ba7-6ad9-e21fc4cd7ca4&t=166bcb7a-b1a3-550e-b07c-c68401b8ea86)? Looks like sig shape of `(1, ...)` is the problem here?

Ah, fuzzing works! It didn't find a repair solution, and faithfully raised an error. We can suppress this error here since the test is only about tile shape negotiation. In this case perhaps it is easiest if you include it here as a drive by change?

```
diff --git a/tests/corrections/test_corrset.py b/tests/corrections/test_corrset.py
index f8f17c4..6137ce6 100644
--- a/tests/corrections/test_corrset.py
+++ b/tests/corrections/test_corrset.py
@@ -383,7 +383,7 @@ def test_tileshape_adjustment_fuzz():
         ])
         print("excluded_coords", excluded_coords.shape, excluded_coords)
         excluded_pixels = sparse.COO(coords=excluded_coords, shape=sig_shape, data=True)
-        corr = CorrectionSet(excluded_pixels=excluded_pixels)
+        corr = CorrectionSet(excluded_pixels=excluded_pixels, allow_empty=True)
         adjusted = corr.adjust_tileshape(
             tile_shape=tile_shape, sig_shape=sig_shape, base_shape=base_shape
         )
```

uellue:
Btw congrats for getting to the bottom of this, amazing detective work! An elusive criminal that bugged tests for the longest time is finally brought to justice. 🕵️

sk1p:
> Ah, fuzzing works! It didn't find a repair solution, and faithfully raised an error. We can suppress this error here since the test is only about tile shape negotiation. In this case perhaps it is easiest if you include it here as a drive by change? [...]

Sounds good, will apply!

> Btw congrats for getting to the bottom of this, amazing detective work! An elusive criminal that bugged tests for the longest time is finally brought to justice. :detective:

Thanks! It was really bugging me, and your comments about using multiple executors in parallel being a problem pointed in the right direction.

It's possible that there is still a problem remaining, as sometimes some tests in [test_start seem to fail or hang](https://dev.azure.com/LiberTEM/LiberTEM/_build/results?buildId=114&view=logs&j=19e54898-1cf3-5b2c-6468-85f8c2fd9739). I'll see if I can get to the bottom of these, too.

tox.ini
```diff
@@ -1,9 +1,9 @@
 [tox]
-envlist = flake8, py{36,37,38,39}, benchmark, benchmark-cuda{101,102}, mypy
+envlist = flake8, py{36,37,38,39}, benchmark, benchmark-cuda{101,102,110}, mypy
 
 [testenv]
 commands=
-    pytest --durations=10 --cov=libertem --cov-report=term --cov-report=html --cov-report=xml --junitxml=junit.xml {posargs:tests/}
+    pytest -s -vvv --durations=10 --cov=libertem --cov-report=term --cov-report=html --cov-report=xml --junitxml=junit.xml {posargs:tests/}
```
sk1p, line 6: Should we keep the very verbose `-s` output for a while? I added this because the build sometimes hung without having any indication how far it got and why it hung. If things stabilize a bit, we should switch this back off, as it can be very overwhelming.

sk1p:
Reported upstream: https://github.com/dask/distributed/issues/4357

tox.ini
uellue: If we must... 🙈 😆  Ok, if it is temporary!

docs/source/tips.rst
```diff
@@ -197,3 +197,25 @@ extra info as well to aid later data evaluation. Note that the compilation tests
...
+    sudo chown root:<yourgroup> /sys/fs/cgroup/cpu,cpuacct/slow
+    sudo chmod 664 /sys/fs/cgroup/cpu,cpuacct/slow
+
+Then, as a user, you can use :code:`cgexec` to run a command in that control group:
+
+.. code-block:: shell
+
+    cgexec -g cpu:slow pytest tests/
+
+This is useful, for example, to debug test failures that only seem to happen in CI.
```
uellue, line 221: Oh, neat! :-)

tests/io/test_cached.py
```diff
@@ -23,12 +23,16 @@ def default_cached_ds(tmpdir_factory, default_raw, lt_ctx):
...
+@pytest.fixture(scope='module', autouse=True)
+def client_in_background():
+    # A running Dask client can introduce a timing issue
+    # between automatic closing of a numpy.memmap object and renaming
+    # the underlying file
     with dd.LocalCluster() as cluster:
         client = dd.Client(cluster, set_as_default=False)
+        yield
+        # to fix "distributed.client - ERROR - Failed to reconnect to scheduler after 10.00 seconds, closing client"  # NOQA
+        client.close()
```
uellue, line 35: So that means we always have a second client in the background (`autouse=True`)? How much time does that consume per module?

tox.ini
```diff
@@ -12,15 +12,21 @@ deps=
...
     PY_IGNORE_IMPORTMISMATCH=1
-    # Debug asyncio problems:
-    PYTHONASYNCIODEBUG=1
+    # Debug asyncio problems - has some perf overhead, so only enable if there is a problem
+    # PYTHONASYNCIODEBUG=1
+    # Dask configuration to reduce background load:
+    DASK_DISTRIBUTED__ADMIN__TICK__INTERVAL=1000
+    DASK_DISTRIBUTED__SCHEDULER__WORK_STEALING_INTERVAL=1000
+    DASK_DISTRIBUTED__WORKER__PROFILE__CYCLE=60000
+    DASK_DISTRIBUTED__WORKER__PROFILE__INTERVAL=30000
```
uellue, line 29: Ref #61 -- would it make sense to use that by default in LiberTEM?

uellue (approved):

tests/io/test_cached.py
sk1p: As the fixture is defined in a single module, it should only be used there, right? I can double check if that is true.

tox.ini
sk1p: Yeah, we could possibly set a few of these when creating a local cluster. We should try to measure this, to check where the tweaking actually makes sense (there are some stats in `Worker.digests`, need to look at how to access them)

tests/io/test_cached.py
uellue: Ok, should be easy enough by trying to use it in a test in another module, right? Good point that it is not in conftest.py, so probably you are right. :-)

sk1p: > Ok, should be easy enough by trying to use it in a test in another module, right?

Yeah - I just tried by adding a debug print, and running another test after `test_cached` doesn't start the cluster.

sk1p:
I think all comments above were addressed, I'm going to merge this now.

PR #924, opened 2020-12-12
Update configuration for new sphinxcontrib-bibtex release

sk1p:
This is required because of some backward-imcompatibilities, namely the requirement for setting `bibtex_bibfiles`. Include the `bibtex.json` file in source control as suggested by upstream, if this is kept up to date this means we could remove the second build pass for updating the references.

Also:

 - use parallel building to speed up the build.
 - extract requirements for building docs into its own requirements file, which can be shared between tox environments

## Contributor Checklist:

changes were made to the GUI)

## Reviewer Checklist:

### DIFF
--- a/docs/source/bibtex.json
+++ b/docs/source/bibtex.json
@@ -1,0 +1,37 @@
+{
+    "cited": {
+        "app/amorphous": [
+            "Gibson1997",
+            "6980942",
+            "Panova2019"
+        ],
+        "app/holography": [
+            "Lichte2008"
+        ],
+        "index": [
+            "doi:10.1002/9783527808465.EMC2016.6284",
+            "Ophus_2019",
+            "Krajnak2016",
+            "Tate2016",
+            "Simson2015"
+        ],
+        "reference/app/amorphous": [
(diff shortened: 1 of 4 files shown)


sk1p:
Merged to keep master green. @uellue let me know if you think we can remove the double build now, and I'll do a follow-up PR.

uellue:
> Merged to keep master green. @uellue let me know if you think we can remove the double build now, and I'll do a follow-up PR.

How would be the process of keeping bibtex.json up to date? Probably we should add that to the documentation? Maybe just sticking with a double build would be easier and more user-friendly than managing that file in Git?

sk1p:
> How would be the process of keeping bibtex.json up to date? Probably we should add that to the documentation? Maybe just sticking with a double build would be easier and more user-friendly than managing that file in Git?

Doing a single build locally should update the file, then it just needs to be committed as part of the changes to the docs. I agree that keeping the double build should be more user-friendly for now - the "interactive" rebuild only builds once, so it should be fast enough already. Thanks for the feedback!

PR #927, opened 2020-12-14
Use same rounding and snapping in point selection as in pick

### LINKED ISSUE #926: Make displayed and actual point analysis coordinates match
## How to reproduce

* LiberTEM current master, possibly older versions as well
* Open a low-resolution dataset, for example 32x32x32x32, with clearly identifiable points (in this case a double diagonal with outliers).
* Perform a point analysis around the outliers. Disk analysis and frame picking are not affected, others not tested.

## Expected results

* A pixel on the left is selected if the center of the crosshair is within it.
* As a bonus, the selector position could snap to pixel centers like it does for frame picking.

## Observed result

* The selection area for a pixel seems to be shifted by 1/2 to the lower right. Based on the numerical values, one should use "round()" instead of "floor()" to get the integer point position from float coordinates.

See these pictures:

### Centered: correct pixel
Capture-centered
### Lower left corner: Wrong pixel selected
Capture-lower-left
### Lower right corner: Correct pixel selected
Capture-lower-right
### Upper right corner: Wrong pixel selected
Capture-upper-right
### Upper left corner: Pixel to the upper left selected (wrong pixel, other outlier)
Capture-upper-left
### Just below pixel: Still within selection
Capture-below
### Just right of pixel: Still selected
Capture-right

sk1p:
Fixes #926

## Contributor Checklist:

changes were made to the GUI)

## Reviewer Checklist:


uellue:
Fix confirmed, thx! :-)

Probably it should get a changelog entry since it closes a bug that can cause wrong analysis results in the GUI?

sk1p:
> Probably it should get a changelog entry since it closes a bug that can cause wrong analysis results in the GUI?

True, I'll work on it!

sk1p:
Should be ready to merge now.

uellue:
Thank you for fixing this so quickly! Merging. :-)

PR #929, opened 2020-12-15
Allow choosing between lin and log visualization

### LINKED ISSUE #925: Visualization: remove outliers before normalization
If there are unusually large outliers, the default visualization is significantly impacted, as they throw off the normalization:

image

Something like [these functions in libertem-holo](https://github.com/LiberTEM/LiberTEM-holo/blob/43363ebdb749ec3e792c58ada98e723d09ec4070/src/libertem_holo/base/filters.py#L14-L31) could be useful.

uellue:
Closes #925 with a small modification to make sure large outliers
don't generate a completely blank plot.

See https://github.com/LiberTEM/LiberTEM/issues/925#issuecomment-744483882
for more discussions and an idea for a more comprehensive solution

Full implementation would be #134

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/analysis/masks.py
+++ b/src/libertem/analysis/masks.py
@@ -138,7 +138,10 @@ class SingleMaskResultSet(AnalysisResultSet):
     intensity_log : libertem.analysis.base.AnalysisResult
         Log-scaled sum of the selected region for each detector frame, with shape of
         the navigation dimension. Absolute of the result if the dataset or mask contains
-        complex numbers. Added in 0.6.0.dev0
+        complex numbers.
+
+        .. versionadded:: 0.6.0.dev0
+
     intensity_real : libertem.analysis.base.AnalysisResult
         Real part of the sum of the selected region. This is only available if the dataset
         or mask contains complex numbers.
--- a/src/libertem/analysis/masks.py
+++ b/src/libertem/analysis/masks.py
@@ -81,8 +81,11 @@ class SingleMaskAnalysis(BaseMasksAnalysis):
             )
         return SingleMaskResultSet([
             AnalysisResult(raw_data=data, visualized=visualize_simple(data),
-                           key="intensity", title="intensity",
-                           desc=self.get_description()),
+                           key="intensity", title="intensity [lin]",
+                           desc=f'{self.get_description()} lin-scaled'),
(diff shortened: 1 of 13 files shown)


sk1p:
TODO:

- [ ] review code
- [ ] check GUI behavior

src/libertem/analysis/masks.py
```diff
@@ -132,6 +135,10 @@ class SingleMaskResultSet(AnalysisResultSet):
         Sum of the selected region for each detector frame, with shape of
         the navigation dimension. Absolute of the result if the dataset or mask contains
         complex numbers.
+    intensity_log : libertem.analysis.base.AnalysisResult
+        Log-scaled sum of the selected region for each detector frame, with shape of
+        the navigation dimension. Absolute of the result if the dataset or mask contains
+        complex numbers. Added in 0.6.0.dev0
```
sk1p, line 141: Maybe `.. versionadded` should be used instead? Should make the presentation a bit more consistent, and also makes it greppable

uellue: `..versionadded::`doesn't work in the Parameters section since that uses special formatting. For that reason I already didn't use it in other instances, for example the docstring of `create_com_analysis()` The only way I found working was placing it in the description text above, like this. What is the better option in your opinion?

image

sk1p: I would prefer the version using `versionadded`. Actually, it should be possible to use `versionadded`/`versionchanged` in parameters, it just needs to be indented further. See this example:

https://github.com/numpy/numpy/blob/8e038f4a930c0c899e599ee58a56391e026a608f/numpy/f2py/__init__.py#L29-L63

And how it is rendered:

https://numpy.org/doc/stable/f2py/usage.html#numpy.f2py.compile

If that doesn't work, I would still prefer having `versionadded` outside the parameters section. Maybe we can make this consistent in our next doc review.

sk1p (commented):
Looks good to me! Once docs comments are resolved, this can be merged. Thanks!

src/libertem/analysis/masks.py
uellue: Oh, Thank you! It seems to need the empty lines before and after. I'll fix the instances that I know of here.

> Maybe we can make this consistent in our next doc review.

Yes, sounds good! :-)

sk1p (approved):

Issue #798, opened 2020-06-02
Ensure jitted functions can be cached


sk1p:
If we can cache them, we can get rid of of some nasty latencies introduced by #706, which is important for a smooth user experience.

sk1p:
As an update, I filed https://github.com/numba/numba/issues/6264

I also checked the potential savings using a patched numba that can cache functions captured in a closure. Measured via the GUI on an 8GiB MIB dataset, LiberTEM-stable (v0.5.0) takes about 3.8s to compute the initial sum of all frames, where LiberTEM-master+patches takes 4.6s. I'll retry on a larger data set in a bit, there the throughput improvements should outweigh the initial additional latency (by a noticeable margin)

Maybe the remaining latency can be hidden by heating up the cache directly after loading a data set, as we discussed before - it could possibly also be reduced in the cache miss case, as it is enough to run one compilation job per host to populate the cache, instead of having multiple redundant compilations (for each core) competing for resources.

uellue:
In case numba/numba#6284 is not included in the next Numba release, we could perhaps implement a workaround to get the 0.6 release out in reasonable time.

We only need this caching in a well-defined subset of LiberTEM that has a well-defined interface with user code: the I/O routines.  Furthermore, the caching in Numba is implemented in a central place and your patch seems to work. We could implement a context manager that monkey-patches the Numba cache with your patch *only* when LiberTEM enters I/O routines. That means user code will not be affected by the monkey patching.

Heating up the cache directly after loading sounds like a good idea! It felt more acceptable to wait when opening instead of waiting for the results to come in after adding an analysis or clicking "Apply".

sk1p:
> We only need this caching in a well-defined subset of LiberTEM that has a well-defined interface with user code: the I/O routines. Furthermore, the caching in Numba is implemented in a central place and your patch seems to work. We could implement a context manager that monkey-patches the Numba cache with your patch only when LiberTEM enters I/O routines. That means user code will not be affected by the monkey patching.

This monkey-patched hack would need to make sure that the cache key includes some unique element that makes sure the cache is busted once the hack is removed. I'll have a look into this.

> Heating up the cache directly after loading sounds like a good idea! It felt more acceptable to wait when opening instead of waiting for the results to come in after adding an analysis or clicking "Apply".

[...]

PR #931, opened 2020-12-16
JIT caching continuation

sk1p:
Refs #798

## Contributor Checklist:

changes were made to the GUI)

## Reviewer Checklist:

### DIFF
--- a/benchmarks/common/test_numba_warmup.py
+++ b/benchmarks/common/test_numba_warmup.py
@@ -1,0 +1,13 @@
+import pytest
+
+from libertem.common.numba import prime_numba_cache
+
+
+@pytest.mark.benchmark()
+def test_numba_prime_cache(lt_ctx, benchmark, large_raw):
+    """
+    Primes the cache for the raw dataset. this only makes sense to be run individually, really,
+    or we need to find a way to dynamically set the numba cache path somehow, and/or run the
+    test in a subprocess.
+    """
+    benchmark.pedantic(prime_numba_cache, kwargs=dict(ds=large_raw), rounds=5, warmup_rounds=0)
--- a/src/libertem/common/numba.py
+++ b/src/libertem/common/numba.py
@@ -230,7 +230,7 @@ class MyFunctionCache(FunctionCache):
         data = super().load_overload(*args, **kwargs)
         t1 = time.time()
         if data is None:
-            logger.info("CACHE MISS CUSTOM!! %s %s" % (self._name, self._py_func))
+            logger.info("numba cache miss %s %s" % (self._name, self._py_func))
         else:
             logger.info("cache hit for %s, load took %.3fs" % (self._name, (t1 - t0)))
         return data
--- a/src/libertem/common/numba.py
+++ b/src/libertem/common/numba.py
@@ -207,7 +207,7 @@ class MyFunctionCache(FunctionCache):
     def load_overload(self, *args, **kwargs):
         data = super().load_overload(*args, **kwargs)
         if data is None:
-            print("CACHE MISS CUSTOM!! %r %s %s" % (sig, self._name, self._py_func))
+            print("CACHE MISS CUSTOM!! %s %s" % (self._name, self._py_func))
         # else:
         #    print("cache hit for %r %r" % (args, kwargs))
         return data
--- a/src/libertem/common/numba.py
+++ b/src/libertem/common/numba.py
@@ -222,7 +222,7 @@ class MyFunctionCache(FunctionCache):
         else:
             cvarbytes = b''
 
-        return (sig, "libertem-dirty-hack", codegen.magic_tuple(), (hasher(codebytes),
+        return (sig, "libertem-numba-cache", codegen.magic_tuple(), (hasher(codebytes),
                                              hasher(cvarbytes),))
 
     def load_overload(self, *args, **kwargs):
--- a/src/libertem/common/numba.py
+++ b/src/libertem/common/numba.py
@@ -1,6 +1,16 @@
+import time
+import hashlib
+import logging
+
 import numba
+from numba.core.registry import dispatcher_registry, CPUDispatcher
+from numba.core.caching import FunctionCache
+from numba.core.serialize import dumps
 import numpy as np
 import scipy.sparse
+
+
+logger = logging.getLogger(__name__)
 
 
 @numba.njit(boundscheck=True)
@@ -155,15 +165,27 @@ def _rmatmul_csr(left_dense, right_data, right_indices, right_indptr, res_inout_
                     res_inout_t[right_column, left_row] += tmp
 
 
+_cached_njit_reg = []
+
+
 def cached_njit(*args, **kwargs):
-    kwargs.update({'target': 'custom_cpu', 'cache': True})
-    return numba.njit(*args, **kwargs)
+    """
+    Replacement for numba.njit with custom caching. Only supports usage
+    with parameters, i.e.
+
+    @cached_njit()
+    def fn():
+        ...
+    """
+    def wrapper(fn):
+        kwargs.update({'_target': 'custom_cpu', 'cache': True})
+        _cached_njit_reg.append(fn)
+        return numba.njit(fn, *args, **kwargs)
+    return wrapper
 
 
-import hashlib
-from numba.core.registry import dispatcher_registry, CPUDispatcher
-from numba.core.caching import NullCache, FunctionCache
-from numba.core.serialize import dumps
+def hasher(x):
+    return hashlib.sha256(x).hexdigest()
 
 
 class MyFunctionCache(FunctionCache):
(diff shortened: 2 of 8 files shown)


uellue:
Now I measured 1:03 min in the second run. 

## Warumup

Start libertem-server, open dataset, perform a point analysis, shut down everything again

## Actual measurement

Start libertem-server, open dataset, measure time from "load dataset" until an analysis can be added.

Once the cluster is started and opened the dataset once, opening the dataset again on the same cluster is nearly instantaneous. Just to clarify what "warming up" means in this case.

sk1p:
> Once the cluster is started and opened the dataset once, opening the dataset again on the same cluster is nearly instantaneous. Just to clarify what "warming up" means in this case.

Thanks for clarifying. Yeah, without actually being able to load the cached compilations, the warm-up won't make much difference yet. I'll try to hack together a custom caching version and push it here.

sk1p:
@uellue if you have time, could you re-run your test? On-disk caching is now implemented (or rather: hacked in), and I would be interested if the total warm-up time, and also the first compile time is now reduced on your system. If you want to re-try the cold case, i.e. including compilation, you can remove all `__pycache__` directories with something like `find src/ -name __pycache__ -exec rm -r {} \;`

sk1p:
Oh, and warm-up times are now also logged, so you don't have to time it externally.

uellue:
First run: 40 s until dataset is opened. Some (short) delay until sum analysis shows.

Second run: 14 s until dataset is opened. Sum analysis result after 5 s.

After opening the dataset first, adding an analysis gives a near instantaneous response on a small dataset.

All in all, this seems to solve it! Congratulations! :-)

uellue:
As a user, I feel that these times are very acceptable once the initial compilation is done. :+1:

sk1p:
> First run: 40 s until dataset is opened. Some (short) delay until sum analysis shows.

> Second run: 14 s until dataset is opened. Sum analysis result after 5 s.

> After opening the dataset first and running the first analysis, adding another analysis gives a near instantaneous response on a small dataset.

Sounds great, thanks for the feedback. As a comparison, on my system, when loading from disk cache, warmup takes a bit less than 5 seconds, with another 2 or so seconds overhead until the sum result is done (2.8s cold vs 0.6s hot). When compiling (so "first-use after update" scenario, per data set type), the warmup takes about 12s.

uellue:
How should we test that the "hacked" Numba cache doesn't affect regular use of Numba? Can we construct an example that fails in that case? Maybe poison the cache with a bad entry?

sk1p:
> How should we test that the "hacked" Numba cache doesn't affect regular use of Numba? Can we construct an example that fails in that case? Maybe poison the cache with a bad entry?

Well, it's hard to prove a negative :) The caching works like this - you have 1 to N cache files for each cached function. For each cached function, there can be multiple instances, depending on a cache key, which are constructed from the following fields:

- the signature
- a hash of the actual Python bytecode of the function
- a hash of the transitive dependencies (closure-captured constants, the bytecode of captured functions and their dependencies)
- a magic value (I think representing the architecture for which the compiled result is valid, including detailed features like AVX2)

The cache is also cleared if the source file `mtime` changes. In our case, we add a constant string value to this key. Let's have a look at a concrete example:

```python
import numpy as np
import numba

from libertem.common.numba import cached_njit


def fn(a, b):
    return a ** b


cached_fn = cached_njit(fastmath=True)(fn)
numba_fn = numba.njit(fastmath=True, cache=True)(fn)


if __name__ == "__main__":
    cached_fn(np.float32(1.2345), 7)  # call 1
    numba_fn(np.float32(1.2345), 7)  # call 2
```

Now, on call 1, a compilation result with our modified key is stored, on call 2, a compilation result with the original numba cache key is stored. After the two calls, there are two files in the `__pycache__` directory (and an index file):

```
-rw-r--r-- 1 alex alex 14203 Dec 17 17:15 ol3.fn-7.py38.1.nbc
-rw-r--r-- 1 alex alex 14203 Dec 17 17:15 ol3.fn-7.py38.2.nbc
-rw-r--r-- 1 alex alex  1943 Dec 17 17:15 ol3.fn-7.py38.nbi
```

When loading from the cache, the cache key must match, otherwise you will get a cache miss. With the current numba version, it would not be possible to get a conflicting key - even the tuple length of all the components of the key doesn't match our custom key.

This is the content of the `nbi` file after running the example file above:

```python
((1608221730.1192634, 333),
 {((float32, int64),
   'libertem-numba-cache',
   ('x86_64-unknown-linux-gnu',
    'ivybridge',
    '+64bit,-adx,+aes,-avx,-avx2,-avx512bf16,-avx512bitalg,-avx512bw,-avx512cd,-avx512dq,-avx512er,-avx512f,-avx512ifma,-avx512pf,-avx512vbmi,-avx512vbmi2,-avx512vl,-avx512vnni,-avx512vpopcntdq,-bmi,-bmi2,-cldemote,-clflushopt,-clwb,-clzero,+cmov,+cx16,+cx8,-enqcmd,+f16c,-fma,-fma4,+fsgsbase,+fxsr,-gfni,-invpcid,-lwp,-lzcnt,+mmx,-movbe,-movdir64b,-movdiri,-mwaitx,+pclmul,-pconfig,-pku,+popcnt,-prefetchwt1,-prfchw,-ptwrite,-rdpid,+rdrnd,-rdseed,-rtm,+sahf,-sgx,-sha,-shstk,+sse,+sse2,+sse3,+sse4.1,+sse4.2,-sse4a,+ssse3,-tbm,-vaes,-vpclmulqdq,-waitpkg,-wbnoinvd,-xop,+xsave,-xsavec,+xsaveopt,-xsaves'),
   ('dfaa35a01cfd5851c85e0db1ecbb7fdbbaf09090cfc9fb90ff66dcf62d381f71',
    'b2b452258982ac957835d77e11f8e57919737afe81c2a9559774a5ad7049a68f')): 'ol3.fn-7.py38.1.nbc',
  ((float32, int64),
   ('x86_64-unknown-linux-gnu',
    'ivybridge',
    '+64bit,-adx,+aes,-avx,-avx2,-avx512bf16,-avx512bitalg,-avx512bw,-avx512cd,-avx512dq,-avx512er,-avx512f,-avx512ifma,-avx512pf,-avx512vbmi,-avx512vbmi2,-avx512vl,-avx512vnni,-avx512vpopcntdq,-bmi,-bmi2,-cldemote,-clflushopt,-clwb,-clzero,+cmov,+cx16,+cx8,-enqcmd,+f16c,-fma,-fma4,+fsgsbase,+fxsr,-gfni,-invpcid,-lwp,-lzcnt,+mmx,-movbe,-movdir64b,-movdiri,-mwaitx,+pclmul,-pconfig,-pku,+popcnt,-prefetchwt1,-prfchw,-ptwrite,-rdpid,+rdrnd,-rdseed,-rtm,+sahf,-sgx,-sha,-shstk,+sse,+sse2,+sse3,+sse4.1,+sse4.2,-sse4a,+ssse3,-tbm,-vaes,-vpclmulqdq,-waitpkg,-wbnoinvd,-xop,+xsave,-xsavec,+xsaveopt,-xsaves'),
   ('dfaa35a01cfd5851c85e0db1ecbb7fdbbaf09090cfc9fb90ff66dcf62d381f71',
    'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855')): 'ol3.fn-7.py38.2.nbc'})
```

(note the `libertem-numba-cache` string included in one key, missing from the other)

At runtime, `cached_fn` and `numba_fn` are different `CPUDispatcher` objects, and have independent `FunctionCache` instances, without one having a reference to the other. So our own cache instance can't interfere with the original numba instance.

It would be possible to save our own cache with a different naming scheme - but that would mean we further couple our code to the numba implementation details (example: we subclass [`IndexDataCacheFile`](https://github.com/numba/numba/blob/0bac18af44d08e913cd512babb9f9b7f6386d30a/numba/core/caching.py#L457) and set our own `version`, `index_name` and `data_name_pattern`). Our current implementation is already coupled to details in the `njit` decorator, the `CPUDispatcher` and the `FunctionCache` classes, and I would suggest to keep this as simple as possible.

Because of the coupling mentioned above, it's quite likely that this hack will stop working with one of the next numba releases, but I don't think the on-disk cache keys can be brought into conflict.

I don't have other ideas how our caching could interfere with 'vanilla' numba, but please do let me know if you have further ideas!

uellue:
Thank you for explaining it in such detail! That is quite convincing.

That example already looks a lot like a test case that shows how the hacked decorator generates cache keys that can under no circumstances overlap with other cache keys that the regular decorator creates, even for the same function. It should be possible to wrap this into test. The `__pycache__` folder is per source code folder, right? Maybe we can have a single test file in a separate sub-folder of `tests/`, purge that directory, run the function with the hacked decorator and with the normal one, make sure the results are all correct, and confirm that the *.nbi file in the `__pycache__` folder now contains two cache keys of different length? I can take care of that if you like. Maybe I am just a bit paranoid. 😆 

The second question is that we do generate separate keys for the dataset reading functions where we have to. The dataset tests seem to pass, already a good sign! If the hacked cache was to return the same function in cases where it should be different: Would that crash, cause obvious errors, or would it silently return results that are different in subtle ways?

sk1p:
> The `__pycache__` folder is per source code folder, right? Maybe we can have a single test file in a separate sub-folder of `tests/`, purge that directory, run the function with the hacked decorator and with the normal one, make sure the results are all correct, and confirm that the *.nbi file in the `__pycache__` folder now contains two cache keys of different length? I can take care of that if you like.

Maybe a good idea get inspiration from [the numba test cases](https://github.com/numba/numba/pull/6284/files) which work similarly - IIRC they copy the test module to a temporary directory, import that and then check the `__pycache__`.

> If the hacked cache was to return the same function in cases where it should be different: Would that crash, cause obvious errors, or would it silently return results that are different in subtle ways?

I think subtly different results are unlikely, it's pretty likely that it would either crash or cause obviously wrong results - either strange patterns when the wrong "pixel format" decoder is used, or completely scrambled if the "read ranges" are from another format.

>  Maybe I am just a bit paranoid.

Which is good in this case - I'm a bit worried about compatibility with numba versions - I think there are a few things planned to change in the caching layer for the next release. Maybe we should pin numba at least to `<0.53`? We could still do minor releases just for dependency updates, like 0.5.1. (or we could be optimistic and plan ahead for 0.53 having the caching fixed, and just bypass our code in that case :D)

uellue:
> I think subtly different results are unlikely, it's pretty likely that it would either crash or cause obviously wrong results - either strange patterns when the wrong "pixel format" decoder is used, or completely scrambled if the "read ranges" are from another format.

Ok, that should be noticeable latest in our release QA, probably earlier! 👍

uellue (approved):
Looks great! 🚀 The reduced waiting time is a relieve.

Thx for taking the time with the test case. I was just getting paranoid since it could have pretty far reaching and hard to debug consequences if it went sideways.

Issue #63, opened 2018-08-14
K2 IS binary data reader


sk1p:
- [x] figure out guarantees (is block ordering deterministic?, are frame IDs for blocks really monotonic or can they "interlace" so to speak?, ...)
- [x] figure out what data we need to discard ("desynced" frames at the beginning and end)
- [x] integrate into LiberTEM as DataSet implementation
- [x] split sectors / binary files into smaller partitions
- [x] don't require direct mapping of 
  - [x] sectors to partitions
  - [x] blocks to tiles
- [ ] crop frames to user specified area (see comment by @woozey below, and #28 for the general case)
- [x] optimize conversion uint12 -> uint16 (see #108)
- [x] read tags from gtg file (for us most important: scan dimensions, see #109)

sk1p:
Depending on how well we can handle the weird 12bit format and the shuffling of data blocks, we may need to convert a sensible format (hdf5, some kind of proto-nexus?) before processing.

sk1p:
I got something slow but seemingly correct working. Still need to figure out the detailed guarantees of the format, and implement appropriate optimizations. Then, integrate into LiberTEM.

uellue:
@sk1p Here are some CPU instructions that might help to make it fast: 

General: https://en.wikipedia.org/wiki/Bit_Manipulation_Instruction_Sets
Testing if CPU has it: https://stackoverflow.com/questions/44827477/does-avx-support-imply-bmi1-support

AVX-512: https://software.intel.com/en-us/node/523829
BMI2: https://www.felixcloutier.com/x86/PEXT.html
BMI1: https://www.felixcloutier.com/x86/BEXTR.html

What would be better, using the intrinsics directly or writing it in pure C/C++ and checking if the compiler actually generates efficient machine code?

uellue:
A second point, efficiently unpacking n-bit (u)ints packed into a bitfield to 8-bit, 16-bit, 32 bit, 64 bit (u)ints could be a useful function for numpy and such. Apparently it doesn't exist yet, at least I couldn't find it. A quick Google only found https://stackoverflow.com/questions/37797779/python-reading-12-bit-packed-binary-image which looks horrendously slow.

uellue:
How fast is this one? https://pythonhosted.org/bitstring/constbitarray.html#bitstring.Bits.unpack

sk1p:
[...]

PR #932, opened 2021-01-20
Improve file format test stability; use process_tile() in SumSigUDF

### LINKED ISSUE #937: Validate file readers against alternative implementations
In #932, some tests that use a form of snapshot testing are replaced with tests that validate against alternative file reader implementations to make the results more stable against numerical variations and consistent with other relevant packages.

A prototype notebook with alternative readers for all  formats except K2IS is included in that PR. However, only selected tests for DM, FRMS6 and MRC are converted yet which were affected by the numerical instability. The alternative readers should be included for other file format tests as well:

- [x] FRMS6 full file, not only beginning
- [x] DM full file and 3D DM
- [x] MRC full file
- [x] BLO
- [x] EMPAD
- [x] MIB
- [x] SEQ
- [x] SER
- [x] K2IS

uellue:
Make sure that file format tests are stable against numerical variations in the UDFs that generate reference results. Validate with `allclose()` against results that are generated with an alternative reader and calculation method instead of comparing exact snapshots.

Closes #937 

Original reason: `process_tile()` in `SumSigUDF` is significantly faster, about 1.5x better in some tests

## Contributor Checklist:

changes were made to the GUI)

## Reviewer Checklist:


uellue:
Regarding the failure of the "data" pipeline, it seems that we have a slight difference in the numerical result, as to be expected when changing the summation scheme. @AnandBaburajan, @sk1p, is there a particular reason why hashes are compared, and not results via `np.allclose()`? 

How should we fix the tests? Using hard-coded hashes will run into this issue again since the hash may even change if the implementation of `np.sum()` is changed or if NumPy is compiled with a different compiler or compiler settings.

* Compare against a calculated references using `allclose()` instead of hashes. Since `SumUDF` and `SumSigUDF` are linear, we can also include internal consistence checks by making sure that the sum of partial sums match up with different subdivisions.
* Compare against pre-calculated raw data for the UDF results using `allclose()` instead of comparing hashes.
* Have a RAW version for each test file which can be plain memory mapped to calculate reference results using only NumPy.

sk1p:
The hashes here are basically a poor man's snapshot testing...

> is there a particular reason why hashes are compared, and not results via `np.allclose()`?

Basically two reasons: 1) performance and keeping total test run time low, 2) this will also catch problems that change the result for both code paths - there could be cases where an `allclose` comparison would still work, but the results are wrong, consistently for `sync_offset=0` and `sync_offset=X`.

I'm leaning towards 

> Have a RAW version for each test file which can be plain memory mapped to calculate reference results using only NumPy.

But instead, use [snapshottest](https://pypi.org/project/snapshottest/), so it's easier to update the expected results. What do you think?

uellue:
> The hashes here are basically a poor man's snapshot testing...
> 
> > is there a particular reason why hashes are compared, and not results via `np.allclose()`?
> 
> Basically two reasons: 1) performance and keeping total test run time low, 2) this will also catch problems that change the result for both code paths - there could be cases where an `allclose` comparison would still work, but the results are wrong, consistently for `sync_offset=0` and `sync_offset=X`.
> 
> I'm leaning towards
> 
> > Have a RAW version for each test file which can be plain memory mapped to calculate reference results using only NumPy.
> 
> But instead, use [snapshottest](https://pypi.org/project/snapshottest/), so it's easier to update the expected results. What do you think?

Sounds good! :-) Perhaps we can use an independent file reader where possible to generate the reference data? Where is the reference, actually?

sk1p:
> Perhaps we can use an independent file reader where possible to generate the reference data?

We could - it just won't integrate well with the snapshot testing. Maybe for formats that have existing readers we could have an additional test suite that actually installs those readers and directly compares the output to ours?

> Where is the reference, actually?

You mean the test data? `iff588:/cachedata/users/clausen/libertem-test-data/`

uellue:
> it just won't integrate well with the snapshot testing.

I meant to generate the RAW version, sry! :-)

AnandBaburajan:
> > Have a RAW version for each test file which can be plain memory mapped to calculate reference results using only NumPy.
> 
> But instead, use [snapshottest](https://pypi.org/project/snapshottest/), so it's easier to update the expected results. What do you think?

I'll try to get a better understanding of snapshot testing and implement this!

uellue:
FYI, currently I am writing routines that convert the input files to raw data independent of LiberTEM, as far as that is possible.

uellue:
The conversion notebook is finished now. Should I integrate the generation of verification data into the unit tests, for example as a module-scope fixture for the data, or do we want to keep raw data around? For the former we'd have to add a dependency for Hyperspy and PyXem. Does that go into test-requirements.txt?

uellue:
> I'll try to get a better understanding of snapshot testing and implement this!

@AnandBaburajan Do you have access to our moellenstedt workstation that runs the data pipeline, by the way? I can't remember if we gave you an account at some point... @sk1p somehow the snapshots will have to be persisted, right? By running them manually on moellenstedt with, I guess, `--snapshot-update`, and then committing to git?

uellue:
Essentially we want to test the results of a number of UDFs on the reference datasets against "known good" results with `allclose()` instead of a precise match, right?

As a reference, we can use results calculated with NumPy based on data that is generated with an alternative reader. The "snapshotting" could come from canning the alternative raw data, and/or canning the alternative NumPy result.

For using this with snapshottest, we'd need two capablities:
1. Comparing current and previous results with `allclose()` instead of a precise match
2. Comparing a result against an entry that was generated with a different method.

Looking at snapshottests, I couldn't find documentation on support for either of the two. How should we handle this? The easiest way would perhaps be to keep the raw version of reference data around as an "anchor" in time, calculate NumPy results on the fly (this is fast on moellenstedt), and compare with `allclose()`.

uellue:
Quick discussion with @sk1p: We should generate the reference data "on the fly" using the alternative reading methods and install the necessary packages in the tox environment for the data pipeline.

Snapshotting can perhaps be done later after figuring out how to do it properly in this case?

@AnandBaburajan thx for offering to help! Let me know if you have access to moellenstedt and/or the reference files and are still interested to work on this. :-)

AnandBaburajan:
> @AnandBaburajan thx for offering to help! Let me know if you have access to moellenstedt and/or the reference files and are still interested to work on this. :-)

I don't have access to moellenstedt. I do have the files and I'm interested but I think I'll complete my pending issues first, so I'll skip.

>  @sk1p somehow the snapshots will have to be persisted, right? By running them manually on moellenstedt with, I guess, --snapshot-update, and then committing to git?

Doubt: the snapshot tests would be run locally too, right?

uellue:
For now I've included the packages in the test environments since I couldn't find an obvious way to install them only for the data pipeline.It seems that Hyperspy and the CI environment don't get along on Windows. 🙄 Workaround pushed recently.

@sk1p it seems that the environment for the data pipeline should be rebuilt? How do I trigger this since it didn't seem to happen automatically?

uellue:
> so I'll skip

OK, no problem, thx for the quick reply!

> Doubt: the snapshot tests would be run locally too, right?

So you mean the tests that generate the reference for the snapshot, or the tests that confirm that the snapshot is still the same?

sk1p:
> For now I've included the packages in the test environments since I couldn't find an obvious way to install them only for the data pipeline.It seems that Hyperspy and the CI environment don't get along on Windows.  Workaround pushed recently.
> 
> @sk1p it seems that the environment for the data pipeline should be rebuilt? How do I trigger this since it didn't seem to happen automatically?

I would suggest to create a new tox testenv for this purpose (and change [the environment in the pipeline](https://github.com/LiberTEM/LiberTEM/blob/b6490efb3f795801a234db7a2ebb1a10ce181315/azure-pipelines-data.yml#L14)), then the impact on other environments (like Windows) is limited. Then, you can add an `-r` flag in [the pipeline definition](https://github.com/LiberTEM/LiberTEM/blob/b6490efb3f795801a234db7a2ebb1a10ce181315/azure-pipelines-data.yml#L35)

AnandBaburajan:
> So you mean the tests that generate the reference for the snapshot, or the tests that confirm that the snapshot is still the same?

I actually meant both. So the tests that generate the reference for the snapshot would be run only in CI and the tests that confirm that the snapshot is still the same is run both locally and CI. Did I get it correct now? Thanks!

uellue:
> I actually meant both. So the tests that generate the reference for the snapshot would be run only in CI and the tests that confirm that the snapshot is still the same is run both locally and CI. Did I get it correct now? Thanks!

Hm, I thought it would be generating the reference locally at the request of the user, and then verify it automatically in CI?

uellue:
Ok, skipping like that doesn't seem to work...

uellue:
@sk1p what do you think about including the raw version for the K2IS dataset in the test data portfolio since there are no other readers available in Python, as far as I know?

sk1p:
Just as a reference, I've bisected the change in K2IS behavior using these two small helpers:

To create a reference frame file from the stable LiberTEM version:

```python
import numpy as np
from libertem.udf.raw import PickUDF


if __name__ == "__main__":
    ctx = Context()
    ds = ctx.load("auto", path=sys.argv[1])
    roi = np.zeros(ds.shape.nav, dtype=np.bool)
    roi.reshape((-1,))[0] = True

    res = ctx.run_udf(dataset=ds, udf=PickUDF(), roi=roi)
    np.array(res['intensity']).tofile("./frame0.raw")
```

To compare the pick result with the reference:

```python
import sys

import numpy as np
from libertem.udf.raw import PickUDF


if __name__ == "__main__":
    ctx = Context()
    ds = ctx.load("auto", path=sys.argv[1])
    roi = np.zeros(ds.shape.nav, dtype=np.bool)
    roi.reshape((-1,))[0] = True

    res = ctx.run_udf(dataset=ds, udf=PickUDF(), roi=roi)

    picked = np.array(res['intensity'])
    loaded = np.fromfile(sys.argv[2], dtype=np.uint16).reshape((1860, 2048))
    if not np.allclose(picked, loaded):
        print("not allclose, diff=%d" % (np.sum(picked - loaded.astype("float32"))))
        sys.exit(1)
    else:
        print("allclose")
        sys.exit(0)
```

It appears that the 0th frame in 0.5.1 is the 1st frame in libertem-master. As I somehow expected, it was introduced in 14c5144bbd2285e9a460bada8320c8f05ad189a0. I'll check which behavior we actually want next, this is possibly related to the `skip_frames` handling. 

(btw. `git bisect run [...]` is really cool: just using the above script, git could automatically find the first "bad" commit, without having to babysit and run script/`bisect bad`/`bisect good` all the time)

sk1p:
As far as I can see, the new behavior is correct.

As reference, please see this comment: https://github.com/LiberTEM/LiberTEM/issues/63#issuecomment-424028403

This shows off-by-one behavior, and is actually exactly what 0.5.1 shows:

image

Now, in LiberTEM-master, the off-by-one is fixed, and the result is the same as [manually skipping 57 frames](https://github.com/LiberTEM/LiberTEM/issues/63#issuecomment-418090259):

image

So, 0.5.1 syncs directly to the "shutter active" flag (even though it was supposed to already correct for this via the `skip_frames=-1`! Seems like there was a bug, which I fixed in the new tiling and I/O system), whereas LiberTEM-master syncs one frame before, which is more correct and behaves like the manual synchronization a user would apply.

tests/io/datasets/test_blo.py
```diff
@@ -30,6 +31,14 @@ def default_blo():
     return ds
 
 
+@pytest.fixture(scope='module')
+def default_blo_raw():
+    import hyperspy.api as hs  # avoid importing top level
```
sk1p, line 36: Could we can make `hyperspy` et. al optional? I think these tests have a higher probability to fail, as we don't explicitly keep up with changes in all of those projects. A `try..except` with a `pytest.skip` should do the trick, as the comparison tests are now all independent from the other, right?

Same goes for the other formats.

tests/io/datasets/test_k2is.py
```diff
@@ -29,6 +32,16 @@ def default_k2is():
     return ds
 
 
+@pytest.fixture(scope='module')
+def default_k2is_raw():
+    return np.memmap(
```
sk1p, line 37: Could we make the raw reference data optional, too? I do have the K2 reference data on my laptop, for example, but I don't think I have space to also keep raw references around - what do you think?

tests/io/datasets/test_k2is.py
```diff
@@ -13,10 +13,13 @@
...
 from libertem.io.dataset.base import TilingScheme
 from libertem.common import Shape
+from libertem.common.buffers import reshaped_view
 
-from utils import dataset_correction_verification, get_testdata_path
+from utils import dataset_correction_verification, get_testdata_path, ValidationUDF
 
 K2IS_TESTDATA_PATH = os.path.join(get_testdata_path(), 'Capture52', 'Capture52_.gtg')
+#K2IS_TESTDATA_PATH = os.path.join(get_testdata_path(), 'Capture52', 'Capture52_.gtg_(34, 35, 1860, 2048)_uint16.raw')
+K2IS_TESTDATA_RAW = '/cachedata/users/weber/libertem-test-data-raw/Capture52/Capture52_.gtg_(34, 35, 1860, 2048)_uint16.raw'
```
sk1p, line 22: Can be changed back now, I think

tox.ini
```diff
@@ -37,6 +37,14 @@ commands=
 setenv=
     NUMBA_DISABLE_JIT=1
 
+[testenv:data]
+deps=
+    -rtest_requirements.txt
+    -rdata_requirements.txt
```
sk1p, line 43: If we inline `data_requirements.txt` here, the tox environment will be automatically re-created when they change. What do you think?

tests/io/datasets/test_frms6.py
```diff
@@ -197,50 +296,48 @@ def test_read_invalid_tileshape(default_frms6):
         next(p.get_tiles(tiling_scheme=tiling_scheme))
 
 
-def test_positive_sync_offset(lt_ctx):
+def test_positive_sync_offset(default_frms6_raw, lt_ctx):
```
sk1p, line 299: Can we make the test that depends on the raw version independent of the other? Then we get at least some coverage if stemtool is not installed.

sk1p (changes requested):
Thanks for taking care of this, I've added some comments. In general, I think it would be helpful if as many tests as possible could be run without 3rd-party readers installed - I can imagine installing hyperspy could be error-prone for our contributors and create a higher barrier of entry, for example.

In general, the approach looks good :+1: do I need to take care of updating files on moellenstedt? Should I give you write access to the relevant directory? With the K2 reference updated to the one from master, we should be good to go, right?

sk1p:
As a reference, the "active shutter" synchronization was fixed in ef66b290b25b8a06541f50bc1f1b20b30c300420, which was then later squashed into 14c5144. Sadly I didn't add a comment in the commit message back then, sigh.

tox.ini
uellue: Sounds good! 👍

tests/io/datasets/test_frms6.py
uellue: Ok, I've made the verification packages optional and skip only the tests that require it if they are not available.

tests/io/datasets/test_k2is.py
uellue: Done!

tests/io/datasets/test_k2is.py
uellue: Done! :-)

tests/io/datasets/test_blo.py
uellue: Done!

uellue:
If we run into flakiness with the alternative implementations, we can always switch to "plan B" and keep a raw hardcopy like K2 IS. IMO it is actually quite nice that the alternative implementations are tested as well since that may help catch regressions in these packages on top of catching regressions in LiberTEM. :-)

sk1p (approved):
Looks good to me! Also, only takes about 3 more CI minutes; I think this is acceptable for what we gain by these tests. As long as the data tests take less time than the others we should be fine IMHO. I just tested locally (only thing missing is the K2IS comparison data), and went from 40s to 110s, which is also acceptable. That was with parallel testing and 6 workers, which we could also use in CI btw (18 workers should shred through those tests...)

PR #938, opened 2021-01-25
Handle user-provided positional and keyword arguments for autoloading

### LINKED ISSUE #936: Context.load('auto', ...) silently swallows keyword arguments, at least for FRMS6
How to reproduce:

Open an FRMS6 dataset with embedded dark frames and try to disable offset correction with the option `enable_offset_correction=False`.

## Observed behavior

With `Context.load('frms6', path=..., enable_offset_correction=False)`, the dataset is loaded and offset correction is disabled.

With `Context.load('auto', path=..., enable_offset_correction=False)`, the dataset is loaded, but offset correction is still enabled.

## Expected behavior

The offset correction is disabled also for type `'auto'`. Alternatively, an error can be raised. In any case, keyword arguments should not be silently ignored.

uellue:
Closes #936

This change allows to pass arbitrary positional and keyword arguments to
datasets loaded with 'auto' and with a specified type. This is tested with
FRMS6 since it triggered the original bug and has a number of
optional arguments which are convenient for testing.

* Turn `executor` and `enable_async` into keyword-only arguments to avoid
  collisions with user-provided positional arguments
* Make sure that *args and **kwargs are passed forward in all cases
* Avoid auto-detected parameters overriding or colliding with user-
  provided arguments. That is achieved by ignoring all
  detected values except the type
* Consequence: Make sure that dataset initialization can autodetect
  all optional or detectable parameters since they will not be passed
  to load() anymore. Only `ds_path` for HDF5 seemed to be affected.

## Contributor Checklist:

changes were made to the GUI)

## Reviewer Checklist:

<!--

### DIFF
--- a/src/libertem/io/dataset/hdf5.py
+++ b/src/libertem/io/dataset/hdf5.py
@@ -165,7 +165,7 @@ class H5DataSet(DataSet):
     def detect_params(cls, path, executor):
         try:
             executor.run_function(cls._do_detect, path)
-        except (IOError, OSError, KeyError, ValueError, DataSetException):
+        except (IOError, OSError, KeyError, ValueError, TypeError, DataSetException):
             # not a h5py file or can't open for some reason:
             return False
 
--- a/src/libertem/io/dataset/hdf5.py
+++ b/src/libertem/io/dataset/hdf5.py
@@ -96,7 +96,7 @@ class H5DataSet(DataSet):
         Minimum number of partitions, set to number of cores if not specified. Usually
         doesn't need to be specified.
     """
-    def __init__(self, path, ds_path, tileshape=None,
+    def __init__(self, path, ds_path=None, tileshape=None,
                  target_size=512*1024*1024, min_num_partitions=None, sig_dims=2, io_backend=None):
         super().__init__(io_backend=io_backend)
         if io_backend is not None:
@@ -123,6 +123,11 @@ class H5DataSet(DataSet):
         )
 
     def _do_initialize(self):
+        if self.ds_path is None:
+            datasets = _get_datasets(self.path)
+            datasets_list = sorted(datasets, key=lambda i: i[1], reverse=True)
+            name, size, shape, dtype = datasets_list[0]
+            self.ds_path = name
         with self.get_reader().get_h5ds() as h5ds:
             self._dtype = h5ds.dtype
             self._shape = Shape(h5ds.shape, sig_dims=self.sig_dims)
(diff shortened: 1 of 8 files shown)


src/libertem/io/dataset/hdf5.py
```diff
@@ -121,6 +121,11 @@ def get_reader(self):
         )
 
     def _do_initialize(self):
+        if self.ds_path is None:
+            datasets = _get_datasets(self.path)
+            datasets_list = sorted(datasets, key=lambda i: i[1], reverse=True)
+            name, size, shape, dtype = datasets_list[0]
+            self.ds_path = name
```
sk1p, line 128: Why was this necessary? `type="auto"` should work without, right?

If this is to make `type="hdf5"` without `ds_path` work, I agree that it is a good idea to have - maybe the duplication with the same code in `detect_params` can be resolved by extracting into a helper function, something like `_guess_ds_path` or similar? I guess for `detect_params`, that helper also needs to return the whole list of datasets etc.

sk1p (commented):
Looking good! Just added a comment about the HDF5-related changes.

Note that in #896 I've added an xfailing test case that should now work with this PR, please do remove the xfail decorator there:

https://github.com/LiberTEM/LiberTEM/blob/03f843ccd5e7dd4b36dc07f0732aab719e2b05f1/tests/io/datasets/test_backend.py#L57-L69

src/libertem/io/dataset/hdf5.py
uellue: Yes precisely, it makes sure `ds_path` can still be omitted when autodetecting, and additionally can guess it when specifying HDF5. I looked at writing a generic function, but, as you write, both places require different info. Plus, one calls the executor, the other already runs on the executor. I figured it would be easier to just duplicate in this case. If you have a better design idea, let me know!

sk1p: Ah, didn't see the executor difference there. Don't have a better idea, I think this is ready to be merged then. The conflicts still need to be resolved though, please let me know if I should go ahead and take care of that

sk1p (approved):

uellue:
Dealing with the conflict as we speak! It requires some fiddling to get the argument across correctly.

tests/io/datasets/test_backend.py
```diff
@@ -54,9 +40,9 @@ def test_load_uses_correct_backend(lt_ctx, default_raw):
         )
 
 
-@pytest.mark.xfail
-def test_auto_uses_correct_backend(lt_ctx, hdf5):
-    with pytest.raises(RuntimeError):
+def test_auto_uses_correct_backend_hdf5(lt_ctx, hdf5):
+    # H5DataSet currently doesn't support alternative I/O backends
```
sk1p, line 44: *headdesk* yeah, you are right :grinning:

sk1p:
Looks good, merging!

PR #950, opened 2021-02-04
Handle misconfigured pywin32 gracefully

### LINKED ISSUE #949: Proper error message or warning, and documentation for pywin32 install on Win
When installing pywin32 on Windows for the first time, a post-install script has to be run with administrator privileges. Otherwise it will fail to load.

Since the error is a bit cryptic and hard to understand, we should handle it better and inform users.

- [x] Make pywin32 et al. optional, if possible, and warn instead of fail
- [x] Catch the error and provide a more helpful error message with info on how to fix it
- [ ] Include setting up pywin32 in install documentation

uellue:
Show a warning pointing to the probable fix,
disable the affected feature and continue running
instead of failing hard.

Closes #949 

## Contributor Checklist:

changes were made to the GUI)

## Reviewer Checklist:

### DIFF
--- a/src/libertem/win_tweaks.py
+++ b/src/libertem/win_tweaks.py
@@ -3,21 +3,42 @@ import msvcrt
 import ctypes
 from ctypes import windll, wintypes, byref
 import logging
+import warnings
 
-import win32security
-import pywintypes
 
 logger = logging.getLogger(__name__)
 
 
-def get_owner_name(full_path, stat):
-    try:
-        s = win32security.GetFileSecurity(full_path, win32security.OWNER_SECURITY_INFORMATION)
-        sid = s.GetSecurityDescriptorOwner()
-        (name, domain, t) = win32security.LookupAccountSid(None, sid)
-        return "%s\\%s" % (domain, name)
-    except pywintypes.error as e:
-        raise IOError(e)
+def get_owner_name_dummy(full_path, stat):
+    return "Unknown"
+
+
+try:
+    import win32security
+    import pywintypes
+
+    def get_owner_name(full_path, stat):
+        try:
+            s = win32security.GetFileSecurity(full_path, win32security.OWNER_SECURITY_INFORMATION)
+            sid = s.GetSecurityDescriptorOwner()
+            (name, domain, t) = win32security.LookupAccountSid(None, sid)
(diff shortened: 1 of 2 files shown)


uellue:
@sk1p the data pipeline seems to be stuck -- did sth change on Moellenstedt?

uellue:
Regarding test coverage, it would be nice to be able to mock a missing pywin32 and a broken pywin32. However, I couldn't quite figure out how to do that properly. Here are my thoughts so far:

- We'd have to mock win32security to throw the correct exception when importing
- We'd have to make sure that importing sth from win_tweaks actually tries to import the mocked module, although it probably has been already imported before.

sk1p:
> the data pipeline seems to be stuck -- did sth change on Moellenstedt?

I've kicked the azure agent, it should continue running now.

sk1p (approved):
Looks good!

sk1p:
> Regarding test coverage, it would be nice to be able to mock a missing pywin32 and a broken pywin32. However, I couldn't quite figure out how to do that properly. Here are my thoughts so far:
> 
> * We'd have to mock win32security to throw the correct exception when importing
> * We'd have to make sure that importing sth from win_tweaks actually tries to import the mocked module, although it probably has been already imported before.

We can modify `sys.modules` for this - could use the `monkeypatch.setitem(mapping, name, value)` [fixture](https://docs.pytest.org/en/stable/monkeypatch.html). So: remove `win_tweaks` from modules if it is loaded, use `setitem` to set the `win32security` module to the mock replacement, then import `win_tweaks`.

uellue:
@sk1p thx for the pointers! It is actually quite easy to fake a broken module with monkeypatch, although this isn't really documented anywhere -- see the new test case. I've also written it as a blog post since that would have helped me: http://materials-scientist.com/blog/2021/02/11/mocking-failing-module-import-python/

sk1p (approved):

sk1p:
>It is actually quite easy to fake a broken module with monkeypatch, although this isn't really documented anywhere -- see the new test case.

Yup - test cases look good!

> I've also written it as a blog post since that would have helped me: http://materials-scientist.com/blog/2021/02/11/mocking-failing-module-import-python/

:+1:

sk1p:
Thanks, merged!

Issue #954, opened 2021-02-05
Opening a large chunked HDF5 file with gzip compression freezes LiberTEM for a long time


uellue:
Opening `reference\NeXus\MPSD/Reg5_00001.nxs` pushes the CPU load of LiberTEM to 100 % and makes it unresponsive for minutes. Closing the dataset, trying to browse or trying to shut down the cluster only lead to spinners. It eventually does shut down in the background, but the GUI stays spinning forever.

## Assumed reason

Gzip compression in HDF5 is inefficient and processing a partition takes much longer than usual for that reason.

## Expected behavior

- LiberTEM should at least shut down cleanly and quickly if workers get stuck on long-running tasks. Ideally, closing a dataset or removing an analysis shuts down the offending tasks quickly.
- The GUI should respond to the backend being shut down and not show eternal spinners, even if desperate users start mashing buttons.

## Workaround

Ctrl-C or killing the process; reloading the browser window. Avoid chunking or compression in HDF5.

sk1p:
Discussion with @uellue:

The canceling behavior is mostly inherited from `distributed`, we could improve on this by implementing #382, which makes the "unit of work" smaller in these cases. When shutting down a local cluster, we could also be a bit more forceful and kill the worker after a given timeout (need to look into dask API if this is easily implemented)

uellue:
Closed by #972

PR #972, opened 2021-02-16
HDF5: use chunking as base tileshape

### LINKED ISSUE #971: Read performance with HDF5, Windows 10, HDD
## How to reproduce

Run the "center of mass" example with the EMD file from a HDD on Windows.

## What happens

Read speed is about 17 MB/s on my system, while the HDD is capable of about 117 MB/s using LiberTEM with the buffered reader.

## What should happen

Read perfromance of HDF5 should be in the same order of magnitude as other file formats.

sk1p:
Improves cold read performance on my system with chunked HDF5 files, maybe fixes #971 

## Contributor Checklist:

changes were made to the GUI)

## Reviewer Checklist:

### DIFF
--- a/benchmarks/io/test_hdf5.py
+++ b/benchmarks/io/test_hdf5.py
@@ -1,0 +1,90 @@
+import os
+
+import numpy as np
+import pytest
+import h5py
+
+from libertem.api import Context
+from libertem.udf.masks import ApplyMasksUDF
+from libertem.executor.inline import InlineJobExecutor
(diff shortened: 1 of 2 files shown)


uellue:
Fix confirmed: It now reaches just above 60 MB/s, which is acceptable. 👍 

Also closes #954, apparently that was indeed an amplification issue!

uellue:
Thx, merging!

PR #975, opened 2021-02-16
Use `python -m pip`

### LINKED ISSUE #892: Update install documentation to use `python3 -m pip`
See https://snarky.ca/why-you-should-use-python-m-pip/

For some reason, I have to do it like that on moellenstedt after Python was updated. Using `pip install <...>` tries to install with Python 2 in the wrong environment. That means this issue is real and updating our documentation can help to steer newer users away from potential trouble.

uellue:
Closes #892

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -49,10 +49,10 @@ The short version:
 
     $ virtualenv -p python3 ~/libertem-venv/
     $ source ~/libertem-venv/bin/activate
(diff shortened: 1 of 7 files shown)


sk1p (approved):
Looks good!

PR #978, opened 2021-02-16
Remove deprecated Job interface

### LINKED ISSUE #549: Remove deprecated Job functionality

sk1p:
Fixes #549

* libertem.api:
    * Remove `Context.create_*_job`
    * Remove capability to run `Job` in `Context.run`
* libertem.analysis:
    * Remove `get_results` and `get_job` from `Analysis` class
    * Default to `Analysis.TYPE = 'UDF'`
    * throw error when a `BaseAnalysis` with `TYPE = 'JOB'` is instantiated
* Remove `get_results` and `get_job` from `Analysis` implementations
* Clean up tests: remove job-only tests, port others to UDF interface

## Contributor Checklist:

changes were made to the GUI)

## Reviewer Checklist:


sk1p:
> * [ ]  port radial fourier analysis to natively use results from the UDF

@uellue what do you think about keeping the radial fourier analysis as-is? If we remove the transpose, we change the `raw_results` of the `ResultSet`, so we would either have to accept that the result is different, or transpose anyways. Thoughts?

uellue:
> @uellue what do you think about keeping the radial fourier analysis as-is?

+1 keeping it as-is. We can re-evaluate in case the current way creates a real problem for users.

sk1p:
Great! I'll remove the `FIXME` then - from my side this should be ready.

uellue (approved):
Looks great and still seems to work. Thx for taking care of it! 👍

PR #979, opened 2021-02-17
Remove deprecated blobfinder and FeatureVecMakerUDF

### LINKED ISSUE #776: Remove deprecations for 0.7
Title says all: blobfinder, tileshape, ...

sk1p:
Fixes #776

blobfinder is now available as a separate package at https://github.com/liberTEM/LiberTEM-blobfinder

Instead of FeatureVecMakerUDF, you can use a sparse matrix and the `ApplyMasksUDF`.

## Contributor Checklist:

## Reviewer Checklist:


uellue:
@sk1p maybe you can also remove changelog/misc/nothing.rst? I overlooked that one when reviewing the last PR.

sk1p:
> maybe you can also remove changelog/misc/nothing.rst? I overlooked that one when reviewing the last PR.

Sure. What do you think about removing the `pyfftw` extra dependency? Or do we still need it?

uellue:
> What do you think about removing the `pyfftw` extra dependency?

Sounds good! What about the code in utils/threading.py? It works without pyfftw and would just make sure the thread count is correct in case someone uses it in a UDF.

sk1p:
> Sounds good! What about the code in utils/threading.py? It works without pyfftw and would just make sure the thread count is correct in case someone uses it in a UDF.

Yeah, I would keep it, so our users (and libertem-blobfinder) don't have to think about it

sk1p:
Actually I just saw that we already removed the pyfftw already from setup.py, so nothing to do :grin:

sk1p:
We lost some code coverage by removing the blobfinder tests, the largest offender being `libertem.masks.sparse_template_multi_stack` and 
`libertem.masks.radial_gradient` - should we add separate unit tests?

uellue:
> We lost some code coverage by removing the blobfinder tests, the largest offender being `libertem.masks.sparse_template_multi_stack` and
> `libertem.masks.radial_gradient` - should we add separate unit tests?

That would make sense, probably!

uellue:
So, is this ready to be merged? :-)

sk1p:
> So, is this ready to be merged? :-)

Still needs updated tests, didn't find time for that yet. If you have an idea how to write those feel free to take over the PR

uellue:
OK, I'll quickly whip sth up. :-)

sk1p:
Thanks, merging!

PR #982, opened 2021-02-24
Consider ROI size in tiling; drive-by cleanup

uellue:
Previously, the tile shape was always calculated for the full partition size regardless of ROI. That can lead to a relatively large number of very small tiles if the ROI is almost empty in the area of a partition. With this change, the tile shape is calculated based on the actual size that will be processed.

Additional changes:

* Vestigial remnant of the job interface excised
* Handle https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated

## Contributor Checklist:

changes were made to the GUI)

## Reviewer Checklist:

### DIFF
--- a/src/libertem/analysis/gridmatching.py
+++ b/src/libertem/analysis/gridmatching.py
@@ -37,7 +37,7 @@ class PointSelection:
     def __init__(self, correlation_result: CorrelationResult, selector=None):
         self.correlation_result = correlation_result
         if selector is None:
-            self.selector = np.ones(len(correlation_result.centers), dtype=np.bool)
+            self.selector = np.ones(len(correlation_result.centers), dtype=bool)
         else:
             assert len(correlation_result.centers) == len(selector)
             self.selector = selector
@@ -223,7 +223,7 @@ class Matcher:
         scaled_diffs = diffs / (np.maximum(1, np.abs(indices))**0.5)
         errors = np.linalg.norm(scaled_diffs, axis=1)
         matched_selector = errors < self.tolerance
-        matched_indices = rounded[matched_selector].astype(np.int)
+        matched_indices = rounded[matched_selector].astype(int)
         # remove the ones that weren't matched
         new_selector = point_selection.new_selector(matched_selector)
         result = Match.from_point_selection(
@@ -286,10 +286,10 @@ class Match(PointSelection):
         Match
             A :class:`Match` instance with empty selector and all-'nan' attributes
         '''
-        nanvec = np.array([np.float('nan'), np.float('nan')])
+        nanvec = np.array([np.nan, np.nan])
         return cls(
             correlation_result=correlation_result,
-            selector=np.zeros(len(correlation_result), dtype=np.bool),
+            selector=np.zeros(len(correlation_result), dtype=bool),
             zero=nanvec,
             a=nanvec,
             b=nanvec,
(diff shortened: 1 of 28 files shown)


src/libertem/analysis/gridmatching.py
```diff
@@ -223,7 +223,7 @@ def _match_all(self, point_selection: PointSelection, zero, a, b):
         scaled_diffs = diffs / (np.maximum(1, np.abs(indices))**0.5)
         errors = np.linalg.norm(scaled_diffs, axis=1)
         matched_selector = errors < self.tolerance
-        matched_indices = rounded[matched_selector].astype(np.int)
+        matched_indices = rounded[matched_selector].astype(int)
```
sk1p, line 226: Does it maybe make sense to explicitly use a sized `dtype` like `np.int64`? Just using `int` means the reader needs to know the conversion rules of `numpy`, which I also would have to look up.

For the `bool` cases etc. it should not be a problem, as they are unambiguous.

src/libertem/common/buffers.py
```diff
@@ -42,7 +42,7 @@ def empty_aligned(size, dtype):
 
 
 def zeros_aligned(size, dtype):
-    if dtype == np.object or np.prod(size, dtype=np.int64) == 0:
+    if dtype == object or np.prod(size, dtype=np.int64) == 0:
```
sk1p, line 45: Is `np.object` an alias of `object`? I.e. if a caller uses the old way, does this equality check hold?

src/libertem/io/dataset/base/tiling.py
```diff
@@ -540,9 +540,15 @@ def get_scheme(
         full_base_shape = (1,) + tuple(base_shape)
         min_factors = (depth,) + tuple(min_factors)
 
+        if roi is None:
+            containing_shape = partition.shape
+        else:
+            roi_slice = partition.slice.get(roi.reshape(-1), nav_only=True)
+            containing_shape = (np.count_nonzero(roi_slice), ) + tuple(partition.shape.sig)
```
sk1p, line 547: Could also use `partition.slice.adjust_for_roi(roi).shape` for the same effect, I think

sk1p (commented):
Looks good overall, some comments inline.

> Handle https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated

Thanks for taking care of this :+1: - it also looks like this still works with numpy 1.19 on Py36, which was one of the things I was wondering about this transition.

We do have some tests that check the tile shape negotiation - can you maybe add one that shows the effect(s) of taking the `roi` into account? Doesn't have to be fancy, a small variant of one of the existing tests should be enough.

src/libertem/common/buffers.py
uellue: Yes it does. `np.object == object` holds. I tried that before changing. :-)

uellue: Actually, even `np.object is object`. :-)

src/libertem/analysis/gridmatching.py
uellue: > Does it maybe make sense to explicitly use a sized `dtype` like `np.int64`?

In cases where the number can get very high it would make sense to explicitly use int64. In this particular case it doesn't make a difference as long as it is not int8. :-)

uellue: ..also in all other cases where np.int was changed.

src/libertem/io/dataset/base/tiling.py
uellue: Thx, that's shorter! :-)

sk1p:
Thanks, merged!

Issue #922, opened 2020-12-10
Clean solution for interaction with other code that uses Dask


uellue:
In #921 it became evident that multiple Dask clusters and clients are easily created, but can lead to significant side effects and are cumbersome to manage. The default use pattern relies on registering one global Dask client and then implicitly using that one for all operations. That means one can accidentally switch the cluster or client for a user by starting a new cluster or instantiating a client somewhere down the line. That leads to weird errors that are hard to debug.

For that reason we should think through the interaction between LiberTEM and other projects that use Dask, like Hyperspy.

- [x] Avoid registering a default Dask client from LiberTEM code as far as possible -- we use `set_as_default=False`, not sure if there is more to be done. #921 shows that some side effects will remain, though.
- [x] Support running LiberTEM on an existing Dask client or cluster. That should already work in principle, it just needs a good interface, documentation and tests, in particular for threaded Dask workers.
- [x] Allow to register LiberTEM's Dask client/cluster as global client in a controlled fashion to make it available to be used for other purposes than LiberTEM.
- [x] Allow LiberTEM and other Dask clusters and clients started by a user to co-exist peacefully.
- [x] Make sure that integrations and user workflows work in all these scenarios
  - [x] Open Datasets as Dask arrays
  - [x] In the future, open Dask arrays as Datasets #384 
  - [ ] Use LiberTEM for map() functionality on lazy signals in Hyperspy (future)
  - [ ] How would the interactions with the "advanced executor" project #473 look like?

uellue:
As a quick feedback, implementing `map()` for Hyperspy using LiberTEM would be a significant improvement for lazy signals.

Test: Summing each frame in a 220 MB BLO file by mapping `np.sum()` on my laptop.

* LiberTEM using a default Dask.Distributed client (not local cluster, plain vanilla with Dask.Distributed defaults): 1.47 s
* Hyperspy using a regular Signal: 2.19 s (second run without creating a Distributed client: 1.73 s)
* Hyperspy using a lazy signal on Dask.Distributed: 3 min 30 s
* Hyperspy using a lazy signal on regular Dask: ca. 41-46 s

As a side note, the much slower processing on a Dask.Distributed client might point in the direction that lazy loading may have some kind of issue with accidentally sending large amounts of data across the cluster instead of loading and reducing data locally, similar to what we experienced in the past. LiberTEM file readers might help with constructing good Dask arrays on a Dask.Distributed cluster.

uellue:
[...]

PR #983, opened 2021-02-26
Executor that uses dask.delayed

uellue:
Refs #922

This executor can already run the "simple LiberTEM API" example. The next step would be a proper testing strategy. I'm not sure if one should create dedicated tests or run parts of the existing test suite that use the inline executor against this one as well or even use it instead of the inline executor for some tests.

The progress bar doesn't work with this one since it can't yield partial results like the inline one that just walks through them sequentially or the distributed one that yields futures `as_completed()`. So far I couldn't find an API that can access partial delayed results.

## Contributor Checklist:

## Reviewer Checklist:


uellue:
After some consideration, I am not sure if using dask.delayed for an executor is an approach that makes sense. For Dask without distributed I could not find an API for submitting something for immediate computation or to collect results as they come in. For `as_completed()` on a local system, the easiest way would be to simply use `concurrent.futures` directly to compute the tasks. That might be useful to make Dask in general an optional dependency. This could make sense to further separate "core" aspects of LiberTEM from implementation details and particular applications. However, it would probably not help with #922.

In that sense, I'll try to pivot towards trying some points in https://github.com/LiberTEM/LiberTEM/issues/922#issuecomment-787844743

uellue:
The core points described in https://github.com/LiberTEM/LiberTEM/issues/922#issuecomment-787844743 are working now. However, it currently introduces dependencies on Dask in the `UDFRunner` and `BufferWrapper`. Before moving forward, perhaps we can discuss how to structure this in a way that separates Dask-specific aspects from core LiberTEM aspects?

Performance is very promising, by the way: There is little to no overhead from all the wrapping and "result mangling" compared to `run_for_dataset()`, at least for a single host.

uellue:
Superseded by #1170

Issue #838, opened 2020-06-29
Opening an MIB file from HDD is slow with current master


uellue:
How to reproduce:

* LiberTEM current master
* Windows 10 (Merlin PC)
* Stored on a HDD that can reach about 100 MB/s write speed in a 1:1 copy, as an approximate performance measure.
* MIB file 256x256x256x256 with all frames in a single file
* D:\Merlin_data\Janghyun\20200619 - SNBD Test\04_256X256_1ms_TV160/Test.hdr
* File is not in the FS cache
* Open with LiberTEM GUI and add basic analysis

What happens:

* Read performance 7-8 MB/s, unusably slow

Expected: Reaches at least the read performance of LiberTEM 0.5.0, about 70 MB/s

If the file is in the FS cache, read performance is normal (good).

Possibly related to #814?

sk1p:
> If the file is in the FS cache, read performance is normal (good).

This points to the seek patterns for tiled reading being the problem - apparently Windows doesn't recognize those as "linear" and doesn't perform any read-ahead. I think we are at a point where we should make a general decision how to handle this - if we want to have usable performance on slow-seek media, we need to read in a completely different way than if we want to have optimal performance on fast SSDs. On an NVMe SSD, we theoretically can increase performance by dispatching even more parallel read commands (using liburing etc.), possibly reading deeper tiles, which would completely tank HDD perf.

I think it may very well be worthwhile to explore this, as other I/O "backends" can have similar issues, like reading from parallel file systems, NFS, or via HTTP. I'll see that I do some local testing soon, and then we should discuss how we prioritize a fix for this issue.

uellue:
> we should make a general decision how to handle this - if we want to have usable performance on slow-seek media

On the one hand, since performance was acceptable before and people including us still use such media, we should at least be on par with 0.5.0. However, we should also keep the massive performance gains of optimized tiled processing.

Having different code paths for slow-seek and fast-seek media could get messy, though. More code volume, and how to decide which one to use? Are there OS interfaces that can tell on what kind of storage a file is kept? Should we let users decide?

uellue:
By the way, can this be helpful for mmapped data, perhaps?

* https://linux.die.net/man/2/fadvise
* https://docs.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-prefetchvirtualmemory?redirectedfrom=MSDN

uellue:
[...]

PR #984, opened 2021-03-01
HDF5: special case for 2D-chunked nav axes

sk1p:
This seems to work, but certainly can be improved. Needs test and/or benchmark.

Data to reproduce: https://zenodo.org/record/4312960/files/fe60al40_stripe_pattern.hspy (chunked and compressed)

## Contributor Checklist:

changes were made to the GUI)

## Reviewer Checklist:


uellue:
We could add this dataset to https://github.com/LiberTEM/LiberTEM/blob/master/docs/source/sample_datasets.rst#public-datasets :-)

uellue:
Discussion with @sk1p: When adjusting tile shapes we have to make sure that the adjustment of the corrections is respected, i.e. that tile boundaries are not adjacent to dead pixels. We should have a number of test case for HDF5 files with unfortunate chunking and dead pixel correction that is adjacent to chunk boundaries. This should cover a sufficient number of scenarios to make sure that this case would actually be triggered if it is at all possible since the tile shape negotiation is a bit complex in this case.

uellue:
We could actually throw a warning if users open a HDF5 file with chunking, since the chunking implementation in HDF5 seems to be a perfect tool for read amplification. In particular for compressed data you don't only get read amplification, but read amplification acerbated by decompression. It seems near impossible to write a good, efficient reader for HDF5 that works for all possible sizes, shapes and chunking schemes. Chunking in HDF5 seems to mean that exactly one access pattern, block size and settings will work well for a particular file, while all others get killed by massive read amplification.

TL;DR: If you must save data as HDF5, at least save it without compression and without chunking. Better use something else, such as plain RAW files that just work™ in most scenarios, in particular on Linux.

uellue:
Discussion with @sk1p: The cache-related settings of `h5py` seem to have no effect, even with a very large cache size.

sk1p:
> It seems near impossible to write a good, efficient reader for HDF5 that works for all possible sizes, shapes and chunking schemes. Chunking in HDF5 seems to mean that exactly one access pattern, block size and settings will work well for a particular file, while all others get killed by massive read amplification

At least for reasonable chunkings, things should be better now. I think there are two "bad" cases left:

1) "uneven" chunks like `(128, 1, 32, 32)` - creates tiny tiles and large partition sizes
2) very large `np.prod(chunks[:2] + ds.shape.sig)`

> When adjusting tile shapes we have to make sure that the adjustment of the corrections is respected, i.e. that tile boundaries are not adjacent to dead pixels. We should have a number of test case for HDF5 files with unfortunate chunking and dead pixel correction that is adjacent to chunk boundaries. This should cover a sufficient number of scenarios to make sure that this case would actually be triggered if it is at all possible since the tile shape negotiation is a bit complex in this case.

Now there is a fallback to full frame reading, if the negotiation result crosses a chunk boundary.

> The cache-related settings of h5py seem to have no effect, even with a very large cache size.

This was actually a user error, should be better now. There is still an upper limit to prevent excessive memory usage, possibly needs adjustment.

src/libertem/io/dataset/hdf5.py
```diff
@@ -253,58 +264,159 @@ def get_diagnostics(self):
...
         partition_shape = self.partition_shape(
-            target_size=self.target_size,
+            target_size=target_size,
             dtype=self.dtype,
         ) + tuple(self.shape.sig)
+
+        # if the data is chunked in the navigation axes, choose a compatible
+        # partition size (even important for non-compressed data!)
+        chunks = self._chunks
+        weird_chunks = chunks is not None and len(chunks) == 4 and chunks[0] > 1
```
uellue, line 281: Why `len(chunks) == 4`? That would only be 4D STEM, right? What about data with other dimensions? How is the chunking returned if, for example, only the navigation dimension or only the signal dimension is chunked?

src/libertem/io/dataset/hdf5.py
```diff
@@ -253,58 +264,159 @@ def get_diagnostics(self):
...
+        if weird_chunks:
+            partition_shape = (chunks[0], self.shape[1]) + tuple(self.shape.sig)
```
uellue, line 283: This makes an assumption of 2D nav, right? Can we handle n-dimensional as well?

src/libertem/io/dataset/hdf5.py
```diff
@@ -253,58 +264,159 @@ def get_diagnostics(self):
...
+        chunks = self._chunks
+        if chunks is None:
+            return True
+        # all-1 in nav dims works:
+        nav_dims = self.slice_nd.shape.nav.dims
+        chunks_nav = self._chunks[:nav_dims]
+        if all(c == 1 for c in chunks_nav):
+            return True
+        # special case which works well for us:
+        if len(chunks) == 4 and (chunks[1] == self.slice_nd.shape[1] or chunks[0] == 1):
```
uellue, line 317: Again this seems to be 4D specific? Can we make it n-D?

src/libertem/io/dataset/hdf5.py
```diff
@@ -253,58 +264,159 @@ def get_diagnostics(self):
...
         self._corrections.apply(tile_data, tile_slice)
 
-    def _get_tiles_normal(self, tiling_scheme, tileshape_nd, dest_dtype="float32"):
-        data = np.zeros(tileshape_nd, dtype=dest_dtype)
-        with self.reader.get_h5ds() as dataset:
+    def _get_read_cache_size(self) -> Tuple[int, int]:
+        chunks = self._chunks
+        if chunks is None:
+            return 1*1024*1024
+        elif len(chunks) == 4 and chunks[0] > 1:
```
uellue, line 402: 4D specific again?

src/libertem/io/dataset/hdf5.py
```diff
@@ -343,13 +461,37 @@ def _get_tiles_with_roi(self, roi, dest_dtype):
...
+        if self._chunks is not None:
+            return 512  # allow for tiled processing w/ small-ish chunks
+        # HDF5 seems to prefer larger signal slices, so we aim for 32 4k blocks:
+        return 32 * 4096 // np.dtype(self.meta.raw_dtype).itemsize
+
+    def adjust_tileshape(self, tileshape, roi):
+        chunks = self._chunks
+        if roi is not None:
+            return (1,) + self.shape.sig
+        if chunks is not None and len(chunks) == 4 and chunks[0] > 1:
```
uellue, line 482: 4D specific again?

tests/io/datasets/test_hdf5.py
```diff
@@ -423,3 +429,161 @@ def test_hdf5_with_2d_shape(lt_ctx, hdf5_2d):
...
+    neg = Negotiator()
+    tiling_scheme = neg.get_scheme(
+        udfs=udfs,
+        partition=p,
+        read_dtype=np.float32,
+        roi=None,
+        corrections=None,
+    )
+    assert len(tiling_scheme) > 1
+    next(p.get_tiles(tiling_scheme=tiling_scheme, roi=None, dest_dtype=np.float32))
```
uellue, line 589: It seems that this only tests 4D data?

In terms of testing strategy, what do you think of this approach:

* Test between 2D and, let's say, 7D, with random size of the axes.
* Constant data size. Should be large enough so that a partition doesn't fit the HDF5 tile cache, but small enough so that it fits in the FS cache and main memory.
* Test a random portfolio of chunking schemes, maybe with compression to make sure read amplification "hits"
* Use an UDF that just produces a total sum, go from 1D nav to 1D sig
* Compare the time of this UDF with doing `np.sum(hdf5file["data"])`.

If you like, I can take care of that test suite. :-)

uellue (changes requested):
Thank you for dealing with it! This looks quite tricky. HDF5 now firmly holds the title "most annoying file format to support", I guess? 😆 

To be honest, I have a hard time to follow the new code and it is not a small change anymore. Furthermore it seems that it is specialized on 4D data? Before releasing, we should probably test with a lot more parameter space coverage for dimensionality, chunking and possible intersection with dead pixel avoidance, including performance impact.

If you like, I could take care of writing a "mean" test routine that fuzzes all these parameters for stress testing the HDF5 reader.

src/libertem/io/dataset/hdf5.py
sk1p: > Why len(chunks) == 4? That would only be 4D STEM, right? What about data with other dimensions?

Yes, it's right now specialized for the 4D case. The ND case would cause more complexity, like this I can just write the solution down. I can generalize this to ND with more code, for sure.

> How is the chunking returned if, for example, only the navigation dimension or only the signal dimension is chunked?

The number of chunking dimensions always match the dataset shape.

magnunor:
I see you found one of my datasets :) Plenty more here (with even worse chunking!): https://zenodo.org/record/3466591, https://zenodo.org/record/3479124, https://zenodo.org/record/3476746

Not sure if this pull request is the right place to discuss the merits of HDF5-files for storing 4D-STEM data. But as compressed and chunked HDF5 files is a commonly used format for storing this type of data, I'm curious about the comment:

> TL;DR: If you must save data as HDF5, at least save it without compression and without chunking. Better use something else, such as plain RAW files that just work™ in most scenarios, in particular on Linux.

At least for me, the main reason for compression is simply to make the distribution and storage of the files as easy as possible. As this type of data is very compressible (compared to for example the raw mib files): typically the compressed HDF5-files are about 20% of the original mib-file.

With regards to chunking, it makes it so much easier to access sub-parts of the data, which is very common for this type of data. For example, in the dataset you're looking at here, the center beam is contained in (I think) 4 chunks. Thus, we only need to load 1.5% of the full dataset to do that specific processing. The rest of the diffraction pattern is used for some other processing.

HDF5 itself is used due to the ease of storing metadata in it, and that it has file readers for pretty much every programming language.

sk1p:
> I see you found one of my datasets :) Plenty more here (with even worse chunking!): https://zenodo.org/record/3466591, https://zenodo.org/record/3479124, https://zenodo.org/record/3476746

Thanks for publishing these! I'll see how they perform with our changes in this PR.

> Not sure if this pull request is the right place to discuss the merits of HDF5-files for storing 4D-STEM data

I think it's fine, we can move to an issue in case it gets too long :)

> At least for me, the main reason for compression is simply to make the distribution and storage of the files as easy as possible. As this type of data is very compressible (compared to for example the raw mib files): typically the compressed HDF5-files are about 20% of the original mib-file.

I agree, compression is a nice feature to have generally, and definitely useful for exchanging data. It's just the combination of an access pattern that doesn't/can't match the chunking and compression that slows things down, as read amplification is not serviced from the file system cache anymore, but also needs to decompress all the blocks, which boils down to a lot of CPU overhead (10x or more).

> With regards to chunking, it makes it so much easier to access sub-parts of the data, which is very common for this type of data. For example, in the dataset you're looking at here, the center beam is contained in (I think) 4 chunks. Thus, we only need to load 1.5% of the full dataset to do that specific processing. The rest of the diffraction pattern is used for some other processing.

Yeah - I found the dataset from the pyxem sample notebooks, where this is very well demonstrated. The comment was more ranting about HDF5 - the format, as it is realized, makes it very hard to have _generally_ efficient support on the consuming side.

A better implementation of the chunking idea would allow for smaller chunks without much overhead - instead of `(32, 32, 32, 32)`, you would just have your `(1, 1, 32, 32)` blocks or similar (maybe `(1, 1, 32, 64)` to fit in a 4k block) as a fixed unit of compression, and then let the application decide on the rest of the reading strategy. This is how the K2IS format works - and it is much more flexible to read, compared to HDF5. With HDF5, the last time I looked into this, having many smaller chunks also meant more overheads, both on-disk and when reading.

Nowadays, it's also quite efficient to read subsets from non-chunked file formats - with SSDs it becomes quite possible for example to efficiently read slices of `(256, 16, 256)` directly from a raw file, or from a MIB file.

uellue:
@magnunor The issue here is that reading chunked data is only efficient if you access it in the correct order with a chunk size that matches the chunking scheme in the file. Otherwise you get massive read amplification. In many cases, one may want to access full frames. A "naive" algorithm would iterate through all the frames in the order [0,0], [0,1], [0,2], ..., [1,0], ... If you have this access pattern on a compressed HDF5 file with a (32, 32, 32, 32) chunking, you either need a chunk cache that can hold 32 x (scan_x) x detector_size chunks, which is exceedingly large for, let's say, K2 data, or you get a 32x32 read amplification.

In effect, if you use HDF5 for a well-defined reading software, with well-defined number of dimensions and overall size of the axes, you can find a chunking scheme that works well. If you want to support many different applications that can read the data efficiently, and in an application that strives to support reading many different kinds of data from different sources, this greatly complicates the situation and will in most cases lead to very poor performance on large data. That makes chunked HDF5 plain unsuitable as an universal exchange format for data that is larger than a few 100 MB. HDF5 gives you more than enough rope to hang yourself, that is the issue here.

uellue:
As a follow-up, discussion with @sk1p:

Implementing compression for array data in a way that smaller data portions can be read efficiently will likely require some form of chunking. The advantage of the way that HDF5 solves it is that similar data will be combined in a chunk, which improves compression. If one has data in a "natural" layout, similar data portions will be entire frames apart, which may be too big for usual general-purpose compression algorithms.

In general, chunks should be as small as possible to avoid read amplification and to make sure that the chunk cache doesn't have to be very large. Memory pages are 4 kB and the L1 cache is a few tens of kB per core, which would be sensible chunk sizes from this perspective.

uellue:
Any compression and/or chunking scheme for given data should allow to read the data efficiently in the most "naive" way if it is to be used as an universal exchange format. In the case of 4D STEM that would be frame by frame, traversing the navigation space in C order. A chunk should therefore be 1 for the navigation axes with the exception of the innermost navigation axis. This makes sure that all chunks that have been read for a frame will be re-used soon when traversing the navigation space in C order.

Generally, the navigation part of a chunk should be small since the smaller it is, the less data has to be read to obtain data for a full frame. Maybe 8 or 16 in nav direction are enough for efficient compression?

The signal part of a chunk is less critical, as far as I can see. It might help if the "fast" axis, i.e. innermost axis of a chunk, is larger than the rest since this is the direction along which a "naive" algorithm will traverse frame data. In any case, a "naive" algorithm will likely request full frames from the HDF5 library, so that point is moot.

In that sense, maybe a chunking scheme (1, 8, 4, 128) or (1, 8, 32, 32) would work better in this particular case? @magnunor would you like to try how that works in your application in terms of compression ratio and access speed?

The fundamental issue in HDF5 here is that it has no notion of navigation or signal axes and allows users to create any kind of chunking scheme, including really bad ones like (128, 1, 128, 1). Choosing a good chunking scheme for data that will be exchanged between application requires three things:

1. An understanding how chunking affects compression, i.e. if and how it helps to bring similar data closer together.
2. Foresight regarding how applications would likely access data, i.e. an understanding how possible applications that open the data may work "under the hood".
3. A good mental model of the HDF5 library to understand what effect a chunking scheme will have when data is accessed that way.

This is a lot to ask from an average data scientist. Since HDF5 doesn't enforce limits and provides little guidance for choosing a chunking scheme, many HDF5 files in the wild will have less than optimal chunking.

Furthermore, in cases like OneView, K2 or K3 data where one frame is larger than the default HDF5 chunk cache of 1 MB, any chunking scheme with a size larger than 1 in navigation space will lead to read amplification when reading full frames with default settings. That means in these cases the chunk cache should be made larger, which requires awareness and understanding on the reading application side.

In summary, while some form of chunking is required for compression, HDF5 gives way too much freedom to users to choose chunking schemes, which means many of them will have severe unwanted side effects. A good exchange data format should enforce restrictions that make sure that a "naive" access scheme works.

@magnunor maybe we'd be in a position to develop something like that as an advisory, for example within ESTEEM3 and implemented in Hyperspy, pixstem, pyXem etc.? LiberTEM could (and likely will) throw warnings if it encounters HDF5 files with "bad" chunking, so that could help to spread the word. Furthermore, there should be some help in choosing good settings for the HDF5 chunk cache, since there is a significant amount of EM data from high-resolution detectors where the defaults of the HDF5 library will not work.

Anyway, that is a LOT of headache created by supporting HDF5 files. In my impression, HDF5 creates more problems than it solves. That's why I have a strong preference for simpler alternatives such as plain RAW files + JSON sidecar. Even the rather complicated K2IS raw format is easier to support than HDF5 -- mostly because it is absolutely fixed in what data is found where. Once you support one K2IS raw file, you support all of them equally well. With HDF5 files, each file can be unique on a fundamental level, with very little restrictions. This creates new headaches over and over again and again.

src/libertem/io/dataset/hdf5.py
sk1p: This is now generalized to the ND case, not yet tested/benchmarked with realistic data.

magnunor:
I agree that getting the chunking "correct" with hdf5 files can indeed be quite tricky, and it is very easy to do it very badly. However, especially with some of the automatic chunk size functions in dask, it has become a lot easier (see below). 

My experience with hdf5 has lead me to these "best practices" for chunking of 4D STEM datasets:

- Have the same chunk sizes be the same in all dimensions, for example (32, 32, 32, 32). This make sure it is easy to "transpose" the datasets, without having to load everything into memory.
- Have large enough chunk sizes, as having too many leads to extra overhead and slowness (for example (1, 1, 32, 32) is sub-optimal)
- Have small enough chunks, so that we don't have to load a large fraction of the dataset to just access a single diffraction pattern

The (32, 32, 32, 32) number is based on the testing Gary Paterson did in our [fast pixelated detectors paper](https://www.cambridge.org/core/journals/microscopy-and-microanalysis/article/fast-pixelated-detectors-in-scanning-transmission-electron-microscopy-part-i-data-acquisition-live-processing-and-storage/2753DC10CD52B1A9D7FC1D78484232DC#sec5). It found (16, 16, 16, 16), to be optimal, with (32, 32, 32, 32) being very close. I found (32, 32, 32, 32) to be more practical in my use cases.

The reason I like having "square" chunks, is that the access pattern when looking at the data is navigating to around the region you're currently looking at. Ergo, row-chunking like (128, 1, 32, 32) would lead to a lot of slowness when navigating across the y-axis in the navigation dimension.

> Any compression and/or chunking scheme for given data should allow to read the data efficiently in the most "naive" way if it is to be used as an universal exchange format.

I don't think I agree with this. Can't it just be accessed chunk-wise? The chunk size is specified within the HDF5 dataset itself, so this is easy to add to any data processing algorithm. For example with `dask.map_blocks`, or custom implementation. The (32, 32, 32, 32) chunk can be cached, while sending subsets of suitable size to the processing itself.

> Since HDF5 doesn't enforce limits and provides little guidance for choosing a chunking scheme, many HDF5 files in the wild will have less than optimal chunking.

I agree with this, but for example dask's [automatic chunking](https://docs.dask.org/en/latest/array-chunks.html#automatic-chunking) is a useful guide to good chunk sizes. For example:

```python
>>> import dask.array as da
>>> da.ones((256, 256, 256, 256), dtype='uint16', chunks='auto')
dask.array<ones, shape=(256, 256, 256, 256), dtype=uint16, chunksize=(64, 64, 64, 64), chunktype=numpy.ndarray>
```

This can easily be tuned with the `dask.array.core.auto_chunks` function.

----
Testing some of this.

### First make some files

With chunking and compression.

```python
import numpy as np
import h5py

data = np.zeros((50, 50, 200, 200), dtype=np.uint16)
data[:, :, 75:125, 75:125] = np.random.randint(0, 256, size=(50, 50, 50, 50), dtype=np.uint16)

# (1, 1, 25, 25) chunking
f = h5py.File("003_test_chunks_1_1_25_25.hdf5", mode="w")
dataset = f.create_dataset("test_data", data=data, chunks=(1, 1, 25, 25), compression="lzf")
f.close()

# (25, 25, 25, 25) chunking
f = h5py.File("003_test_chunks_25_25_25_25.hdf5", mode="w")
dataset = f.create_dataset("test_data", data=data, chunks=(25, 25, 25, 25), compression="lzf")
f.close()
```
### Summing the whole dataset

`(1, 1, 25, 25)`  vs `(25, 25, 25, 25)` chunking
```python
from time import time
import h5py
import dask.array as da

def test_sum_signal(array):
    dask_array = da.from_array(array, chunks=array.chunks)
    dask_array.sum(axis=(-1, -2)).compute()

def test_sum_navigation(array):
    dask_array = da.from_array(array, chunks=array.chunks)
    dask_array.sum(axis=(0, 1)).compute()


##############################
t0 = time()
test_sum_signal(h5py.File("003_test_chunks_1_1_25_25.hdf5","r")["test_data"])
print("Sum signal. Chunks (1, 1, 25, 25): {0}".format(time() - t0))

##############################
t0 = time()
test_sum_signal(h5py.File("003_test_chunks_25_25_25_25.hdf5","r")["test_data"])
print("Sum signal. Chunks (25, 25, 25, 25): {0}".format(time() - t0))

##############################
t0 = time()
test_sum_navigation(h5py.File("003_test_chunks_1_1_25_25.hdf5","r")["test_data"])
print("Sum navigation. Chunks (1, 1, 25, 25): {0}".format(time() - t0))

##############################
t0 = time()
test_sum_navigation(h5py.File("003_test_chunks_25_25_25_25.hdf5","r")["test_data"])
print("Sum navigation. Chunks (25, 25, 25, 25): {0}".format(time() - t0))
```
Gives:
```python
Sum signal. Chunks (1, 1, 25, 25): 131.44512271881104
Sum signal. Chunks (25, 25, 25, 25): 1.1971850395202637
Sum navigation. Chunks (1, 1, 25, 25): 139.62260055541992
Sum navigation. Chunks (25, 25, 25, 25): 1.2340946197509766
```

### With rechunking of the data

Manually setting the chunking for both datasets to `(25, 25, 25, 25)`.

```python
from time import time
import h5py
import dask.array as da

def test_sum_signal(array, dask_chunk):
    dask_array = da.from_array(array, chunks=dask_chunk)
    dask_array.sum(axis=(-1, -2)).compute()

def test_sum_navigation(array, dask_chunk):
    dask_array = da.from_array(array, chunks=dask_chunk)
    dask_array.sum(axis=(0, 1)).compute()


##############################
t0 = time()
test_sum_signal(h5py.File("003_test_chunks_1_1_25_25.hdf5","r")["test_data"], (25, 25, 25, 25))
print("Sum signal. Chunks (1, 1, 25, 25): {0}".format(time() - t0))

##############################
t0 = time()
test_sum_signal(h5py.File("003_test_chunks_25_25_25_25.hdf5","r")["test_data"], (25, 25, 25, 25))
print("Sum signal. Chunks (25, 25, 25, 25): {0}".format(time() - t0))

##############################
t0 = time()
test_sum_navigation(h5py.File("003_test_chunks_1_1_25_25.hdf5","r")["test_data"], (25, 25, 25, 25))
print("Sum navigation. Chunks (1, 1, 25, 25): {0}".format(time() - t0))

##############################
t0 = time()
test_sum_navigation(h5py.File("003_test_chunks_25_25_25_25.hdf5","r")["test_data"], (25, 25, 25, 25))
print("Sum navigation. Chunks (25, 25, 25, 25): {0}".format(time() - t0))
```
Gives:
```python
Sum signal. Chunks (1, 1, 25, 25): 3.572749376296997
Sum signal. Chunks (25, 25, 25, 25): 1.291045904159546
Sum navigation. Chunks (1, 1, 25, 25): 3.7554969787597656
Sum navigation. Chunks (25, 25, 25, 25): 1.2546427249908447
```

uellue:
@magnunor thank you for the test case! Just for fun I've compared it with memory-mapped IO on an uncompressed raw file using a Dask array, both for your example and for https://github.com/pyxem/pyxem-demos/blob/master/10%20STEM%20DPC%20Analysis%20of%20Magnetic%20Sample.ipynb.

Data can be generated this way:

```python
f = np.memmap('test_raw-50-50-50-50-uint16.raw', mode='w+', shape=(50, 50, 200, 200), dtype=np.uint16)
f[:] = data
del f
```

And this is how a memory-mapped Dask array can be created from a raw file:

```python
def mmap_load_chunk(filename, shape, dtype, sl):
    data = np.memmap(filename, mode='r', shape=shape, dtype=dtype)
    result = data[sl]
    del data
    return result
    
def mmap_dask_array(filename, shape, dtype):
    chunks = []
    blocksize=5
    for offset in range(0, shape[0], blocksize):
        chunk = da.from_delayed(
            delayed(mmap_load_chunk)(
                filename,
                shape=shape,
                dtype=dtype,
                sl=slice(offset, min(offset+blocksize, shape[0]))
            ),
            shape=(min(offset+blocksize, shape[0]) - offset, ) + shape[1:],
            dtype=dtype
        )
        chunks.append(chunk)
    return da.concatenate(chunks, axis=0)
```

This code makes sure that the data is never copied. The blocks of the Dask array are backed directly by the memory map, i.e. the file system cache. Blocks can be instantiated by Dask as needed with virtually zero overhead since creating a memory map is fast and the slicing for the chunks returns a view of the original data. The memory consumption is low and read amplification is minimized: It is limited to the memory page size (read from storage) resp. cache line size (in file system cache) when read operations are scattered across the file.

The chunking can be adapted as needed and large blocks are rather efficient. Compared to the HDF5 case, this minimizes the overheads from Dask. If the data is in the file system cache, it is about 5x faster than reading from HDF5, including transposing, extracting singe pixel virtual detectors etc. If the data is read from disk, it depends a bit how much data is accessed and how fast the storage medium is. I've tested it for the worst case, a HDD on Windows. Sometimes the HDF5 version, sometimes the raw version is faster in that scenario. On SSDs on Linux I'd expect the raw version to do a lot better.

As far as I can see, the remaining advantage of HDF5 is compression. In a quick test, "naive" blosc compression without taking any spatial relation into account achieves half the compression ratio of the HDF5 file on your experimental data.

The advantage of the raw memory map comes from combining very large Dask array chunks that reduce Dask overheads with very little read amplification from exceedingly small minimal access sizes that are managed directly by the operating system or even the CPU. That way one can efficiently extract tiny subsets from a huge Dask array chunk. An ideal compressed data format would retain that advantage, for example with compression on the file system level where the operating system would take care of the chunk cache.

uellue:
@magnunor in a quick test, memory mapping an NFTS compressed files did pretty well. Compression is not as good as the HDF5 file (24 GB -> 12 GB), but it retained much of the advantages of memory mapping. Mounting file systems from file system images is not so easy on Windows, unfortunately. In any case, letting the operating system / file system handle the "chunk cache" and compression would be the best design IMO since it can minimize read amplification and memory footprint at the same time.

uellue:
@sk1p works on my system! On Windows with a HDD, the read rate is way below the disk's capability -- sth like 25-30 MB/s instead of 100 MB/s. Since it was hard enough to fix this for readers where we have much more control in #838 compared to HDF5, it is what it is, I guess. At least there's a warning now. 👍 😄

uellue:
> The stress test seems to succeed now for practically all chunkings, even "bad" ones.

That was a statistical fluke, unfortunately. It is more like 50:50. In any case, better than before and working for the cases that are of practical relevance.

PR #988, opened 2021-03-09
Don't assume pure functions in `DaskJobExecutor`

### LINKED ISSUE #574: Assorted HDF5 bugs
- [x] Reading of nD HDF5 files, see #567
- [x] Debug performance: pickling, general UDF perf, scaling behavior
- [x] Chunk-aware access patterns, if the dataset is chunked (#984)
- [ ] Dataset initialization functions are not generally pure (for example, the one of HDF5), this may need a fix in the dask executor (#988)

sk1p:
Closes #574 

* Updated `run_tasks`, `run_function`, and `map`

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/executor/dask.py
+++ b/src/libertem/executor/dask.py
@@ -128,8 +128,11 @@ class CommonDaskMixin(object):
                 if len(locations) == 0:
                     raise ValueError("no workers found for task")
                 locations = locations.names()
-            submit_kwargs['resources'] = self._validate_resources(workers, resources)
-            submit_kwargs['workers'] = locations
+            submit_kwargs.update({
+                'resources': self._validate_resources(workers, resources),
+                'workers': locations,
+                'pure': False,
+            })
             futures.append(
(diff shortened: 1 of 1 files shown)


uellue:
Looks good, thank you! :-) I'll merge it after the data pipeline has finished.

FYI, at least on my system I couldn't trigger different behavior with or without the patch. I navigated to the file opening dialogue for a HDF5 file, cancelled opening the file, changed the compression method of the file (recreated it with the same name), and opened it again. If the result of the detection function was cached, it should show the previous compression, right? It always showed the correct compression, with or without the patch. In any case, with this change the implementation is "more correct". :-)

sk1p:
> FYI, at least on my system I couldn't trigger different behavior with or without the patch. [...]  In any case, with this change the implementation is "more correct". :-)

Indeed, this was mostly to prevent potentially hard-to-debug problems in the future.

PR #993, opened 2021-03-11
Move threading control to JobExecutor

sk1p:
The `JobExecutor` is in complete control over the threading, so it can also decide to allow fine-grained Python/C threads in each process, for example if we only start `NUM_OF_CORES/2` processes, or in single-process mode (`InlineJobExecutor`).

This is the minimal change, a complete implementation could add information about the number of threads to `UDFMeta`, as discussed before.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/docs/source/changelog/features/threading_control.rst
+++ b/docs/source/changelog/features/threading_control.rst
@@ -1,0 +1,9 @@
+[Feature] Allow some UDF-internal threading
+===========================================
+* The executor can now allow UDFs to perform some internal threading,
+  making the number of allowed threads available as :code:`UDFMeta.threads_per_worker`.
+  This is mostly interesting for ad-hoc parallelization on top of the
+  :code:`InlineJobExecutor`, but could also be used for hybrid
+  multiprocess/multithreaded workloads (:pr:`993`).
+  Threads for numba, pyfftw, OMP/MKL are still automatically controlled,
+  the :code:`UDFMeta.threads_per_worker` addition is meant for other threading mechanisms.
--- a/src/libertem/executor/base.py
+++ b/src/libertem/executor/base.py
@@ -3,7 +3,10 @@ import functools
 import asyncio
 import threading
 import queue
+from typing import Optional
+from contextlib import contextmanager
 
+from libertem.utils.threading import set_num_threads
 from libertem.utils.async_utils import adjust_event_loop_policy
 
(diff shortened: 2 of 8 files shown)


uellue:
If I understand correctly, right now the `TaskProxy` takes care of setting the environment. Alternatively, could the executor just take care of setting the threading environment and calling a Task with all necessary parameters itself? Right now it looks like each executor requires its own task proxy. Or is there a chance to re-use `TaskProxy` functionality across executors?

sk1p:
> If I understand correctly, right now the `TaskProxy` takes care of setting the environment. Alternatively, could the executor just take care of setting the threading environment and calling a Task with all necessary parameters itself?

The executor is free to implement this however it wants. The `InlineJobExexutor` part could alternatively look like this:

```python
def run_tasks(self, tasks, cancel_id):
    threads = psutil.cpu_count(logical=False)
    env = Environment(threads_per_worker=threads)
    with set_num_threads(threads):
        for task in tasks:
            result = InlineTaskProxy(task)()
            yield result, task
```


> Or is there a chance to re-use `TaskProxy` functionality across executors?

That would also be a possibility, to just pass `nthreads` to the `TaskProxy`.  I don't have a preference - either way would be fine with me. As this is strictly an internal API, we can change this at any time, if need arises.

I could also imagine that we could move the actual `set_num_threads` call to the `Environment`, and then just do

```python
with env:
    ...
```

We could then also move some of the CUDA/cupy initialization code from `run_for_partition` into `Environment`, to activate the right device etc. - but that would be a more invasive change than this PR.

uellue:
Just setting the environment in `run_tasks()` would be more straightforward to understand for me. :-)

sk1p:
> Just setting the environment in `run_tasks()` would be more straightforward to understand for me. :-)

Done.

uellue (commented):
Discussion with @sk1p: After some consideration, the `with env:` approach is probably cleanest. It separates the concern of how to set up an environment for the task from saying what that environment should be (executor) and managing tasks (`TaskProxy`). That allows to easily re-use the "environment setting" code across executors. That would also fit nicely with managing the whole CuPy/CUDA concern in the future. :-)

uellue:
Thanks a lot, looks good! I'll try that in SSB next. Merging!

PR #994, opened 2021-03-14
Support for a final postprocessing on main node

### LINKED ISSUE #628: Allow final postprocessing on the central node in an UDF

sk1p:
Add `UDF.get_results` method that is called on the master node after each merge. Fixes #628

Note that for the `StdDevUDF`, the implementation is completely backwards-compatible - this is a bit awkward, as the arrays are returned either as `ndarray` or as `BufferWrapper` . Other UDFs may have similarly decisions to make. Especially with `roi`s this may be a bit problematic - the UDF has to decide if it wants to return the derived result as `raw_result`-like array, or embedded into a `NaN`-filled array.

For other UDFs, where backwards-compatibility is not a concert, `get_results` can now return `BufferWrapper`s, too:

```python
def get_results(self):
    return {'derived': self.result(name='derived', data=np.log1p(self.buffers.original)}
```

The repetition of `name='derived'` is a bit unfortunate, but IMHO better than having to repeat the `kind/dtype/extra_shape` here, which is now looked up by buffer name.

In `get_result_buffers`, one can now use  `self.buffer(..., allocate=False)` to declare placeholder buffers, which is essentially a performance optimization.

## TODO

- [x] test case for `kind='nav'` w/ `UDF.result` and `.raw_data` etc.
- [x] details in docstring for `UDF.result`

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/udf/stddev.py
+++ b/src/libertem/udf/stddev.py
@@ -319,21 +319,22 @@ class StdDevUDF(UDF):
 
         Returns
         -------
-        pass_results : Dict[str, Union[BufferWrapper, numpy.ndarray]]
-        Result dictionary with keys :code:`'sum', 'varsum', 'num_frames'` as :code:`BufferWrapper`,
-            and :code:`'var', 'std', 'mean'` as :class:`numpy.ndarray`.
+        pass_results : Dict[str, BufferWrapper]
+            Result dictionary with keys
+            :code:`'sum', 'varsum', 'num_frames', 'var', 'std', and 'mean'`
+            as :code:`BufferWrapper`
         '''
-        num_frames = int(self.results['num_frames'].data[0])
+        num_frames = int(self.results['num_frames'].raw_data[0])
 
-        var = self.results['varsum'].data / num_frames
+        var = self.results['varsum'].raw_data / num_frames
 
         return {
             'num_frames': self.results['num_frames'],
             'varsum': self.results['varsum'],
             'sum': self.results['sum'],
-            'var': var,
-            'std': np.sqrt(var),
-            'mean': self.results['sum'].data / num_frames,
+            'var': self.result(name='var', data=var),
+            'std': self.result(name='std', data=np.sqrt(var)),
+            'mean': self.result(name='mean', data=self.results['sum'].raw_data / num_frames),
         }
 
 
@@ -343,10 +344,15 @@ def consolidate_result(udf_result):
     from raw UDF results and consolidate the per-tile frame counter
     into a single value. Convert all result arrays to `ndarray`.
 
+    Note
+    ----
+    This is mostly here for backwards-compatability - nowadays, 'var', 'std',
+    and 'mean' are already calculated in :meth:`StdDevUDF.get_results`.
+
     Parameters
     ----------
     udf_result : Dict[str, BufferWrapper]
-        UDF result with keys 'sum', 'varsum', 'num_frames'
+        UDF result with keys 'sum', 'varsum', 'num_frames', 'var', 'std', 'mean'
 
     Returns
     -------
@@ -354,14 +360,14 @@ def consolidate_result(udf_result):
         Result dictionary with keys :code:`'sum', 'varsum', 'var', 'std', 'mean'` as
         :class:`numpy.ndarray`, and :code:`'num_frames'` as :code:`int`
     '''
-    result = {}
-    result.update(udf_result)
-    result.update({
+    return {
         'num_frames': udf_result['num_frames'].data[0],
         'varsum': udf_result['varsum'].data,
         'sum': udf_result['sum'].data,
-    })
-    return result
+        'var': udf_result['var'].data,
+        'std': udf_result['std'].data,
+        'mean': udf_result['mean'].data,
+    }
 
 
 def run_stddev(ctx, dataset, roi=None, progress=False):
--- a/src/libertem/udf/stddev.py
+++ b/src/libertem/udf/stddev.py
@@ -305,10 +305,9 @@ class StdDevUDF(UDF):
 
         Returns
         -------
-        pass_results : Dict[str, Union[BufferWrapper, numpy.ndarray, int]]
-            Result dictionary with keys :code:`'sum', 'varsum' as `BufferWrapper`,
-            'var', 'std', 'mean'` as :class:`numpy.ndarray`, and
-            :code:`'num_frames'` as :code:`int`
+        pass_results : Dict[str, Union[BufferWrapper, numpy.ndarray]]
+        Result dictionary with keys :code:`'sum', 'varsum', 'num_frames'` as :code:`BufferWrapper`,
+            and :code:`'var', 'std', 'mean'` as :class:`numpy.ndarray`.
         '''
         num_frames = int(self.results['num_frames'].data[0])
 
--- a/src/libertem/udf/stddev.py
+++ b/src/libertem/udf/stddev.py
@@ -190,19 +190,26 @@ class StdDevUDF(UDF):
     ..versionchanged:: 0.5.0
         Result buffers have been renamed
 
+    ..versionchanged:: 0.7.0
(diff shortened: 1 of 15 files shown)


uellue:
Hm, good point about the `raw_data`. For `StdDevUDF` it doesn't matter at the moment since we don't have `kind='sig'` ROIs. For other UDF it could become pretty confusing, though, if we mix and match `BufferWrapper` and NumPy arrays...

IMO `get_results()` should always return a dict of `BufferWrapper` and it should do the conversion using `raw_data`. Do we have a convenient interface to create `BufferWrapper` objects from a NumPy array, since people who implement UDFs will require that? :-)

sk1p:
> Hm, good point about the `raw_data`. For `StdDevUDF` it doesn't matter at the moment since we don't have `kind='sig'` ROIs. For other UDF it could become pretty confusing, though, if we mix and match `BufferWrapper` and NumPy arrays...

Yeah.

> IMO `get_results()` should always return a dict of `BufferWrapper` and it should do the conversion using `raw_data`. Do we have a convenient interface to create `BufferWrapper` objects from a NumPy array, since people who implement UDFs will require that? :-)

That's also what I was leaning towards - I'll build an example UDF to demonstrate this - I tihnk it should be possible, otherwise I'll extend `BufferWrapper` or `UDF.buffer` with something convenient.

sk1p:
> Do we have a convenient interface to create `BufferWrapper` objects from a NumPy array, since people who implement UDFs will require that? :-)

This now exists as `UDF.result`, which creates a `BufferWrapper` around an already-allocated array (see updated PR description). On the other side, buffers can now be declared `allocate=False` in `get_result_buffers` - which allows to declare additional buffers without any serialization overhead. Adding `allocate=True` buffers did have measurable overhead in `StdDevUDF` - this should help, especially in cases of large frames.

sk1p:
@uellue let me know what you think about the API - if there's anything else to change, I'll push it together with the flake8 fix.

src/libertem/udf/stddev.py
```diff
@@ -297,12 +313,35 @@ def process_tile(self, tile):
...
+        '''
+        Calculate variance, mean and standard deviation from raw UDF results
+
+        Returns
+        -------
+        pass_results : Dict[str, Union[BufferWrapper, numpy.ndarray]]
+        Result dictionary with keys :code:`'sum', 'varsum', 'num_frames'` as :code:`BufferWrapper`,
+            and :code:`'var', 'std', 'mean'` as :class:`numpy.ndarray`.
+        '''
+        num_frames = int(self.results['num_frames'].data[0])
```
uellue, line 326: Probably one should use `.raw_data` instead of `.data`? It works correctly in this case since it is `kind='sig'`, but would set a bad example for `kind='nav'`.

The code is not using the new `self.result()` method, right? This could be a good place for that since the `consolidate_result()` method can make sure that the previous return values are kept for backwards compatibility.

uellue (changes requested):
I really like the API, looks good! Thx. :-)

Would `UDF.result()` work with `.raw_data` from a `kind='nav'` result in case a ROI is used? In that case, `BufferWrapper.data` takes care of patching with NaN. Is this tested somewhere? I couldn't find a test for `kind='nav'` + ROI. Plus, the behavior in that case should probably be documented.

Generally, one should probably clarify if `.data` (patched with NaN) or `.raw_data` is to be used with `UDF.result()`. I'd prefer using `raw_data` intuitively.

SSB is a case where the result could be a synthetic `kind='nav'` result that covers the entire navigation dimension even if a ROI is used. It could also remain `kind='single'` with `extra_shape=ds.shape.nav`, which is probably a more consistent way of handling synthetic results that cover the navigation dimension. For `kind='nav'`, I as a user would have the expectation that `len(bufferwrapper.raw_data == np.count_nonzero(roi)`

src/libertem/udf/stddev.py
sk1p: > The code is not using the new self.result() method, right? This could be a good place for that since the consolidate_result() method can make sure that the previous return values are kept for backwards compatibility.

That's what I meant with "completely backwards-compatible" - if we change this here, "direct" users of `StdDevUDF` will have to slightly adjust their code. Agree on `.raw_data` for setting an example. Maybe also something for the docs, if I can come up with a reasonable `kind='nav'` example.

sk1p:
> I really like the API, looks good!

Thx :)

> The code is not using the new self.result() method, right? This could be a good place for that since the consolidate_result() method can make sure that the previous return values are kept for backwards compatibility

Should we slightly break backwards compat for direct users of that UDF in favor of having a more consistent return type?

> Would `UDF.result()` work with `.raw_data` from a `kind='nav'` result in case a ROI is used? In that case, `BufferWrapper.data` takes care of patching with NaN. Is this tested somewhere? I couldn't find a test for `kind='nav'` + ROI. Plus, the behavior in that case should probably be documented.

That's true - I'll add tests for this.

> Generally, one should probably clarify if `.data` (patched with NaN) or `.raw_data` is to be used with `UDF.result()`. I'd prefer using `raw_data` intuitively.

I'll add some more detail to the docstring. I think it should be `raw_data`, too.

> SSB is a case where the result could be a synthetic `kind='nav'` result that covers the entire navigation dimension even if a ROI is used. It could also remain `kind='single'` with `extra_shape=ds.shape.nav`, which is probably a more consistent way of handling synthetic results that cover the navigation dimension. 

It should remain `kind='single'` IMHO, as we also have the `reconstruct_shape` property that's used everywhere - this could be changed independently from the `nav` shape, if we apply some kind of resampling, or if the scan was sparse etc.

> For `kind='nav'`, I as a user would have the expectation that `len(bufferwrapper.raw_data == np.count_nonzero(roi)`

Yes, indeed, this should be an invariant.

src/libertem/udf/stddev.py
uellue: The original `StdDevUDF` was only returning these buffers as BufferWrapper and the UDF is still returning them exactly like that, right?

```
'varsum': self.buffer(
    kind='sig', dtype=dtype
),
'num_frames': self.buffer(
    kind='single', dtype='int'
),
'sum': self.buffer(
    kind='sig', dtype=dtype
)
```

We are adding new buffers, and I don't think it would make a difference if they are `ndarray` or `BufferWrapper`. We can just make sure that `consolidate_result()`, which is the main "gateway", returns exactly the same as before. Maybe we can deprecate it, although I am unsure if that is worth the effort in this case.

uellue:
> Should we slightly break backwards compat for direct users of that UDF in favor of having a more consistent return type?

If I understand the code correctly, this only breaks backwards compatibility if someone expected that ONLY the old buffers are present, right? See https://github.com/LiberTEM/LiberTEM/pull/994#discussion_r596003294 :-)

sk1p:
> If I understand the code correctly, this only breaks backwards compatibility if someone expected that ONLY the old buffers are present, right? See [#994 (comment)](https://github.com/LiberTEM/LiberTEM/pull/994#discussion_r596003294) :-)

The `num_frames` result type is also different, `int` instead of `ndarray(dtype=int)`

uellue:
> The `num_frames` result type is also different, `int` instead of `ndarray(dtype=int)`

What stops us from letting the updated UDF still return ndarray(dtype=int)? :-)

sk1p:
> What stops us from letting the updated UDF still return `ndarray(dtype=int)`? :-)

Convenience, mostly. But I could live with it, as it should not be a frequent need to get the number of frames. Would you agree?

uellue (approved):
Looks perfect, merging! :-)

PR #995, opened 2021-03-17
Set CPU scheduling affinity

sk1p:
## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/docs/source/changelog/features/affinity.rst
+++ b/docs/source/changelog/features/affinity.rst
@@ -1,0 +1,3 @@
+[Feature] Set CPU affinity when starting dask workers
+=====================================================
+* This can improve cache locality and thus performance (by a few percent for our tests)
--- a/docs/source/changelog/features/affinity.rst
+++ b/docs/source/changelog/features/affinity.rst
@@ -1,3 +1,3 @@
 [Feature] Set CPU affinity when starting dask workers
 =====================================================
-* This can improve cache locality and thus performance (by a few percent for our tests)
+* This can improve cache locality and thus performance (by a few percent for our tests, :pr:`995`)
--- a/src/libertem/executor/dask.py
+++ b/src/libertem/executor/dask.py
@@ -2,6 +2,7 @@ from copy import deepcopy
 import functools
 import logging
 import signal
+import os
 
 from dask import distributed as dd
 
@@ -23,6 +24,8 @@ def worker_setup(resource, device):
         set_use_cuda(device)
     elif resource == "CPU":
         set_use_cpu(device)
+        if hasattr(os, 'sched_setaffinity'):
+            os.sched_setaffinity(0, [device])
     else:
         raise ValueError("Unknown resource %s, use 'CUDA' or 'CPU'", resource)
 


sk1p:
Test failure is a timeout, maybe #996 should be merged before this one.

uellue:
Could it be possible that LiberTEM sees cores in CI where it can actually not run anything? For example, if a Docker container is limited to specific cores https://thorsten-hans.com/docker-container-cpu-limits-explained#assign-containers-to-dedicated-cpus, would it still "see" the other CPUs and then try to pin workers to these, where they will never run?

I'll close and open again to re-run, to see if this was a glitch.

uellue:
Ok, that method to re-run doesn't seem to work with Azure. :-D

sk1p:
> Could it be possible that LiberTEM sees cores in CI where it can actually not run anything?

Uhh that could be possible. Maybe we can't blindly use the "CPU number" from 0 to N from the cluster spec for the affinity, it's possible we have to build up a list of "eligible" core IDs, which we index into.

We can also leave this PR open for a while until we find a solution, as it's not high priority right now.

sk1p:
> Could it be possible that LiberTEM sees cores in CI where it can actually not run anything?

At least with the docker method described in the article, `sched_setaffinity` return `EINVAL` if there is a mismatch:

```
docker run --rm -it --cpuset-cpus 1,3,5 debian:stable /bin/bash
[...]
>>> import os
>>> os.sched_setaffinity(0, [7])
  File "<stdin>", line 1, in <module>
OSError: [Errno 22] Invalid argument
>>> os.sched_setaffinity(0, [1])
>>> os.sched_setaffinity(0, [2])
  File "<stdin>", line 1, in <module>
OSError: [Errno 22] Invalid argument
>>> os.sched_setaffinity(0, [3])
>>> os.sched_setaffinity(0, [4])
  File "<stdin>", line 1, in <module>
OSError: [Errno 22] Invalid argument
```

uellue:
Discussion with @sk1p:

Use cases for pinning:
* Default: Pin one worker per physical core since that brings measurable benefits in usual cases
* TODO: Check how pinning affects simultaneous CPU and GPU use since a running CUDA operation will spin one CPU thread at 100 %. This CPU thread should be close to the GPU in question -- @sk1p if I recall your work on multi-GPU computation, this can make a real difference? Maybe allowing CPU pinning for CUDA workers and leave CPU workers that would normally be pinned to that CPU unpinned so that the OS can schedule them elsewhere when the CUDA CPU thread is spinning?
* Disable pinning. This is important for running in containers that might be restricted to specific cores, which might even change at run time.
* Handle a failed pinning attempt (`OSError`) gracefully.
* Allow customized pinning. Use case: Live processing on a large NUMA system where soft real time tasks related to processing camera data should be pinned to cores close to specific network interfaces and/or storage devices. LiberTEM workers should be pinned to keep enough cores free for such tasks.

uellue:
A MVP could be opt-out for pinning, i.e. default pinning or no pinning. Second step would be customizable pinning (proposal by @sk1p) :

```
libertem-worker --cpuset 1,2,4,5
libertem-server --no-cpu-affinity
libertem-server --cpuset 1,2,4,5,7-23
```

Third step would be pinning setup for CPUs + multiple CUDA devices. Since that is pretty complex, perhaps one could develop a routine that can make a proposal for a pinning scheme, for example by measuring transfer rates or latencies between specific cores and GPUs?

PR #1002, opened 2021-03-24
Work around a hang related to Numba in GMS

uellue:
Thx @sklam for the decisive test case!

Still unclear why it hangs when running directly in the background thread,
but this seems to work. :shrug:

## Contributor Checklist:

changes were made to the GUI)

## Reviewer Checklist:


sk1p (commented):
Looks good. Just to confirm, whatever fixes the problem with `numba` does _not_ fix the multiprocessing issues?

uellue:
> Looks good. Just to confirm, whatever fixes the problem with `numba` does _not_ fix the multiprocessing issues?

Unfortunately not: `Can't pickle local object 'main.<locals>.test'`

sk1p:
> Unfortunately not: `Can't pickle local object 'main.<locals>.test'`

Is the error similar / the same when using our `DaskJobExecutor`?

uellue:
> Is the error similar / the same when using our `DaskJobExecutor`?

In that case, it seems that Dask.Distributed fails to bring up the workers. I get an error message "The system cannot find the file specified".

sk1p:
> In that case, it seems that Dask.Distributed fails to bring up the workers. I get an error message "The system cannot find the file specified".

Okay. Should we maybe add a second example that connects to a LiberTEM cluster that is running outside of GMS, with directions on how to start the external scheduler and workers? Otherwise we will only have underwhelming single-threaded performance...

uellue:
> > In that case, it seems that Dask.Distributed fails to bring up the workers. I get an error message "The system cannot find the file specified".
> 
> Okay. Should we maybe add a second example that connects to a LiberTEM cluster that is running outside of GMS, with directions on how to start the external scheduler and workers? Otherwise we will only have underwhelming single-threaded performance...

Good point! This is confirmed to work. We might also remove the note that distributed operation is experimental and put in an instruction how to bring up a cluster, since it works pretty nicely with `libertem-worker` these days. :-)

sk1p (approved):
LGTM!

Issue #627, opened 2020-02-20
Hackathon "postprocessing"


sk1p:
- [ ] update documentation from hackmd document: #638 
- [x] update credits
- [x] add organization landing page #623 
- [x] improve consistency in merge function #631, #1003
- [ ] docs for `DataSet.diagnostics` #637 
- [x] improve usability of `Shape`/`Slice` #630

sayandip18:
Isn't this issue same as #638 ?

uellue:
> Isn't this issue same as #638 ?

#638 is a specification and spin-off from this issue since @sk1p and I realized that this would be a bigger item.

sayandip18:
>     * [ ]  update credits
 
What does this mean?

sk1p:
> What is meant by updating credits?

Sorry we didn't answer this before. This was done in https://github.com/LiberTEM/LiberTEM/pull/708 - basically it means updating our contributor and creator files and crediting people from the hackathon for their contribution.

uellue:
Closing this overview issue since the remaining issues have their own entry and can be adressed separately.

PR #1003, opened 2021-03-24
UDF: Consistency fixes and `get_results` improvements

### LINKED ISSUE #631: UDF: attribute access in merge
Instead of requiring `dict`-like access, also allow access by attribute name, i.e. `dest.pixelsum_nav_raw[:] += src.pixelsum_nav_raw`

Otherwise there is a strange disconnect between the `process_*` functions and `merge`.

### LINKED ISSUE #1000: Difference between self.results.buffername and self.results['buffername']
`UDFData` implements both `__getitem__()` and `__getattr__()`. The former returns a `BufferWrapper` object, the latter a view into the memory of the underlying data array. 

This can be a bit confusing:

* When merging results in an UDF's `process_*()`method: Use `self.results.buffername` to get the proper view.
* When inside a UDF's `merge()` function: Use sth like `dest['buffername'][:] = dest['buffername'] + src['buffername']` since the buffers are passed in as `Dict`.
* When in `get_results()`: Use `self.results['buffername']` to pass through a `BufferWrapper`, but use `self.result('buffername', some_operation(self.results.buffername))` for a final postprocessing step. Alternatively, use `self.results['buffername'].raw_data` instead of self.results.buffername

### LINKED ISSUE #1001: Boilerplate in `get_results()` with many derived buffers
Right now, a real-world `get_results()` function can look like this:

```python
...
return {
    'num_frames': self.results['num_frames'],
    'varsum': self.results['varsum'],
    'sum': self.results['sum'],
    'var': self.result(name='var', data=var),
    'std': self.result(name='std', data=np.sqrt(var)),
    'mean': self.result(name='mean', data=self.results['sum'].raw_data / num_frames),
}
```

or like this:

```python
def get_results(self):
    results = get_results(self.results)
    return {
        'fourier': self.results['fourier'],
        'complex': self.result('complex', results),
        'amplitude': self.result('amplitude', np.abs(results)),
        'phase': self.result('phase', np.angle(results)),
    }
```

The buffer name has to be repeated for each buffer. This could be improved by returning a dict of `buffername --> ndarray` and doing the conversion to a `BufferWrapper` automatically. That has two advantages: First, it saves some tedious typing, including the call to `self.result()`. Second, it enforces that results of an UDF are always `BufferWrapper`.

One could even just pass through buffers that are not in the result dict, i.e. `num_frames`, `varsum` and `sum` in the stddev example, so that one just has to worry about buffers that are to be modified or created.

sk1p:
Now, all accesses to buffers in `UDF.results` and the `src`/`dest` parameters of `UDF.merge` should be done via `__getattr__`; if the `BufferWrapper` is really needed, then `UDF.results.get_buffer` should be used.

As a follow-up to #994, `get_results` now should consistently return numpy arrays, which are then converted to `BufferWrapper`s before returning them to the user. This means we can make `get_results` a bit more DRY by not having to repeat buffer names. Also, buffers that don't have to be post-processed, can be left out in `get_results`, and will be returned unchanged. If you want to use a buffer only internally, without exposing it to the user, pass `use='private'` to `UDF.buffer`.

The `allocate` parameter was merged with the `use` parameter, and now `use='result_only'` should be used instead of `allocate=False`.

* Replace `__getitem__` with `__getattr__` in `merge` methods of built-in UDFs, in `libertem.udf` and in tests
* Move `UDFException` to new module `libertem.exceptions`
* Create a custom warning class for discouraged API usage: `UseDiscouragedWarning`
  This can be especially useful combined with `-W`, for example: `pytest -Werror:libertem.warnings.UseDiscouragedWarning tests/`

Fixes #1000, fixes #1001, fixes #631, refs #627

## Contributor Checklist:

changes were made to the GUI)

## Reviewer Checklist:

### DIFF
--- a/tests/udf/test_buffer_declarations.py
+++ b/tests/udf/test_buffer_declarations.py
@@ -347,10 +347,29 @@ class OldDictMergeAccess(UDF):
         self.results.default[:] = np.sum(frame)
 
     def merge(self, dest, src):
-        dest['default'][:] += src['default'][:]
+        dest['default'][:] = src['default'][:]
 
 
-def test_warning_for_dict_access(lt_ctx, default_raw):
+def test_warning_for_dict_access_merge(lt_ctx, default_raw):
     udf = OldDictMergeAccess()
     with pytest.warns(UseDiscouragedWarning):
-        results = lt_ctx.run_udf(dataset=default_raw, udf=udf)
+        lt_ctx.run_udf(dataset=default_raw, udf=udf)
+
+
+class OldUDFDataDictAccess(UDF):
+    def get_result_buffers(self):
+        return {
+            'default': self.buffer(kind='nav', dtype=np.float32),
+        }
+
+    def process_frame(self, frame):
+        self.results.default[:] = np.sum(frame)
+
+    def postprocess(self):
+        self.results['default'].raw_data
+
+
+def test_warning_for_dict_access_postprocess(lt_ctx, default_raw):
+    udf = OldUDFDataDictAccess()
+    with pytest.warns(UseDiscouragedWarning):
+        lt_ctx.run_udf(dataset=default_raw, udf=udf)
--- a/tests/udf/test_buffer_declarations.py
+++ b/tests/udf/test_buffer_declarations.py
@@ -1,6 +1,7 @@
 import numpy as np
 import pytest
 
+from libertem.warnings import UseDiscouragedWarning
 from libertem.exceptions import UDFException
 from libertem.udf.base import UDF
 from libertem.common.buffers import BufferWrapper
@@ -334,3 +335,22 @@ def test_get_results_nav_with_roi_full_shape(lt_ctx, default_raw):
         results['result_only'].data[roi]
     )
     assert np.allclose(results['result_only'].data[~roi], 0)
+
+
+class OldDictMergeAccess(UDF):
+    def get_result_buffers(self):
+        return {
+            'default': self.buffer(kind='nav', dtype=np.float32),
+        }
+
+    def process_frame(self, frame):
+        self.results.default[:] = np.sum(frame)
+
+    def merge(self, dest, src):
+        dest['default'][:] += src['default'][:]
+
+
+def test_warning_for_dict_access(lt_ctx, default_raw):
+    udf = OldDictMergeAccess()
+    with pytest.warns(UseDiscouragedWarning):
+        results = lt_ctx.run_udf(dataset=default_raw, udf=udf)
--- a/tests/udf/test_buffer_declarations.py
+++ b/tests/udf/test_buffer_declarations.py
@@ -67,7 +67,7 @@ class NoAllocateResultsUDF(UDF):
     def get_results(self):
         return {
             'buf1': self.results.buf1,
-            'buf2': np.array(self.results.buf1) + 1,
+            'buf2': self.results.buf1 + 1,
         }
 
 
@@ -98,7 +98,7 @@ class UnifyResultTypesUDF(UDF):
     def get_results(self):
         return {
             'buf1': self.results.buf1,
-            'buf2': np.array(self.results.buf1) + 1,
+            'buf2': self.results.buf1 + 1,
         }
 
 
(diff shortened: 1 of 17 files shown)


tests/udf/test_buffer_declarations.py
```diff
@@ -0,0 +1,336 @@
...
+2) If a buffer is only computed in `get_results`, it should be marked via
+   `use='result_only'` so it isn't allocated on workers
+3) If a buffer is only used as intermediary result, it should be marked via `use='private'`
+4) Not including a buffer in `get_results` means it will either be passed on
+   unchanged, or dropped if `use='private'`
+5) It's an error to omit an `use='result_only'` buffer in `get_results`
+6) It's an error to include a `use='private'` buffer in `get_results`
+7) All results are returned from `Context.run_udf` as `BufferWrapper` instances
+8) By default, if `get_results` is not implemented, `use='private'` buffers are dropped,
+   and others are passed through unchanged
```
uellue, line 22: This is a nice and concise summary. Maybe we can include that in the documentation instead of here as a comment?

tests/udf/test_buffer_declarations.py
```diff
@@ -0,0 +1,336 @@
...
+        assert 'buf2' not in src
+        dest.buf1[:] += src.buf1
+
+    def process_frame(self, frame):
+        pass
+
+    def get_results(self):
+        return {
+            'buf1': self.results.buf1,
+            'buf2': np.array(self.results.buf1) + 1,
```
uellue, line 101: Is the call to `np.array` necessary?

uellue (changes requested):
Looking very nice, I like it! A few minor items in the following. Strangely, I can't add review comments to code that wasn't changed. Anyway, collecting these here!

This should probably be updated since it is not a dict anymore? https://github.com/LiberTEM/LiberTEM/blob/7e9de51074edd05c1f54748b251e7f0dbb150c8c/src/libertem/udf/base.py#L771-L777

Docstring should probably be updated since we are not returning BUfferWrapper anymore? https://github.com/LiberTEM/LiberTEM/blob/7e9de51074edd05c1f54748b251e7f0dbb150c8c/src/libertem/udf/base.py#L793-L803

Do we have test cases that show that the previous access methods still work, but throw a warning? We certainly have code out there that uses the "dict" method" in `merge` since that was "the way". We might also have code that relies on `self.results['buffername'].raw_data` etc. We should make sure that `UDFData` can still emulate the previous behavior.

tests/udf/test_buffer_declarations.py
sk1p: Not necessary, I'll remove it.

tests/udf/test_buffer_declarations.py
sk1p: I copied it into the docs. I mainly used it as a guide for writing the test cases, so I'll keep it also in that file.

sk1p:
> Looking very nice, I like it! A few minor items in the following. Strangely, I can't add review comments to code that wasn't changed. Anyway, collecting these here!
> 
> This should probably be updated since it is not a dict anymore?
> 
> https://github.com/LiberTEM/LiberTEM/blob/7e9de51074edd05c1f54748b251e7f0dbb150c8c/src/libertem/udf/base.py#L771-L777
> 
> Docstring should probably be updated since we are not returning BUfferWrapper anymore?
> 
> https://github.com/LiberTEM/LiberTEM/blob/7e9de51074edd05c1f54748b251e7f0dbb150c8c/src/libertem/udf/base.py#L793-L803
> 
> Do we have test cases that show that the previous access methods still work, but throw a warning? We certainly have code out there that uses the "dict" method" in `merge` since that was "the way". We might also have code that relies on `self.results['buffername'].raw_data` etc. We should make sure that `UDFData` can still emulate the previous behavior.

Thanks for these, they should be addressed now.

uellue (approved):
Looks good, many thanks! :-)

uellue:
Thanks, merging!

uellue:
As a feedback, this is the difference this PR makes in SSB. I like it! https://github.com/Ptychography-4-0/ptychography/pull/35/commits/35a730a11cb415d4c644db7d5546b8d7f68eef22

PR #1009, opened 2021-03-29
Speed up `test_comparison{,_roi}` for K2IS and FRMS6

sk1p:
These tests are run, if the reference is there, in the default
`pytest tests/` invocation. So they considerably slow down the test runs
for me - mostly because `np.allclose` is quite slow. I see two "fixes"
for this:

1) Don't run the comparison tests by default, mark them "slow" or similar
2) Use a faster comparison function. I would argue that we can require
   equality instead of allowing a tolerance in these cases.
   On one hand, because we are reading the data as integers anyways
   (so it boils down to a `memcmp`), and on the other hand, if we were
   to convert to `float32`, there still shouldn't be a difference,
   because the numbers in range for these formats can easily be
   represented without loss of precision.

I've opted for option 2 for now, because routinely running these tests in development is IMHO valuable.

```
tests/io/datasets/test_k2is.py::test_comparison
before: 1 passed in 107.92s (0:01:47)
after: 1 passed in 14.43s

tests/io/datasets/test_frms6.py::test_comparison
before: 1 passed in 99.43s (0:01:39)
after: 1 passed in 27.13s
```

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/tests/io/datasets/test_frms6.py
+++ b/tests/io/datasets/test_frms6.py
@@ -230,7 +230,8 @@ def test_correction(default_frms6, lt_ctx, with_roi):
 @pytest.mark.skipif(stemtool is None, reason="No stemtool found")
 def test_comparison(default_frms6_uncorr, default_frms6_raw, lt_ctx_fast):
     udf = ValidationUDF(
-        reference=reshaped_view(default_frms6_raw, (-1, *tuple(default_frms6_uncorr.shape.sig)))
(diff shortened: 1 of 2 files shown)


uellue (approved):
Good point, equality should be a good assumption in this scenario since it SHOULD do exactly the same computation. If this starts to create issues in the future, we can have a closer look again.

PR #1011, opened 2021-04-06
`Context.run_udfs` improvements

### LINKED ISSUE #944: Visualization: linear viz doesn't properly show binary images
LiberTEM version: 0.5.1 or later (didn't test earlier)

Test code:

```python
from libertem.viz import visualize_simple
import matplotlib.pyplot as plt
import numpy as np

my_choice = np.random.choice(size=(128, 128), a=[0, 1], p=[0.8, 0.2])
rgb_data = visualize_simple(my_choice)
plt.figure()
plt.imshow(rgb_data)
```

Should show some sparse 10% random 1/0 noise, all-black instead.

Probably a one-off error, compare with result of this:

```python
my_choice = np.random.choice(size=(128, 128), a=[0, 1, 2], p=[0.8, 0.1, 0.1])
rgb_data = visualize_simple(my_choice)
plt.figure()
plt.imshow(rgb_data)
```

Another way to reproduce: point analysis on 1bit data set.

### LINKED ISSUE #980: Live view widgets for LiberTEM results in Jupyter notebooks etc
Right now, live updating results in a Jupyter notebook have to be coded by hand using the internal asynchronous API.

In the future, it would be nice to be able to automate that and get a live updating view of a running UDF with a few simple commands, for example for live data processing.

Possible libraries to use:

* https://github.com/bqplot/bqplot
* https://github.com/glue-viz/bqplot-image-gl

Thx @jacobjma for the pointers!

### LINKED ISSUE #1016: Implement GMS plot for #1011
This will be used for live demos with GMS integration. It can be based on the BQP plot in #1011

sk1p:
Fixes #944 
Fixes #1016
Fixes #980 

# Multiple improvements in `Context.run_udf`

* Allow running multiple UDFs "at once" on a single `DataSet`
* Allow usage from an asynchronous context
* Allow getting results iteratively as a generator (both sync and async)
* Simple plotting, including live-updates (`Context.run_udf(..., plots=True)` for very simple usage)

## Under the hood:

* Add `libertem.viz.{mpl,bgp, gms}` packages, which implement live-updated plotting
* remove code duplication in `UDFRunner.run_for_dataset{,_async}`
* call `UDFRunner.run_for_dataset_sync` from `UDFRunner.run_for_dataset_async`
* merge duplicate utility functions `sync_to_async` and `run_blocking`
* move `async_generator`, `async_generator_eager` to
  `libertem.utils.async_utils` package
* new: `libertem.utils.async_utils.async_to_sync_generator` which does exactly what the name says

# TODO

- [x] Stress test / refinement of API: Port Blobfinder plotting to this framework
- [x] Split into `run_udf` and `run_udf_iter`
- [x] Default visualization for complex numbers?
- [x] Plot: add title: name of UDF and name of channel
- [x] Add a detailed explainer notebook for all plotting
- [x] Add a detailed explainer notebook for all iter / async combinations
- [x] Documentation updates: 
  - [x] multiple UDFs
  - [x] iter method in advanced chapter
  - [x] plotting
- [x] Some refactoring
- [x] Additional tests
- [ ] Review

## Contributor Checklist:

## Reviewer Checklist:


sk1p:
Current status of this PR is: the basic features are there, and tests are passing. After a first review, the docs will need updates, and we'll probably want to add more tests for the new functionality. Also, code quality can be improved, as some of the new parts are a bit complicated.

src/libertem/api.py
```diff
@@ -510,6 +516,10 @@ def run_udf(self,
         .. versionchanged:: 0.6.0
             Added the :code:`corrections` and :code:`backends` parameter
 
+        .. versionchanged:: 0.7.0
+            Added the :code:`plots`, :code:`sync`, and :code:`iterate` parameters,
+            and the ability to run multiple UDFs on the same data at the same time.
+
```
uellue, line 521: The documentation of the `udf` parameter should include info that now iterables of UDFs are also allowed.

src/libertem/api.py
```diff
@@ -533,16 +543,33 @@ def run_udf(self,
             Restrict the back-end to a subset of the capabilities of the UDF.
             This can be useful for testing hybrid UDFs.
 
+        plots : None or True or List[List[str]] or List[LivePlot]
```
uellue, line 545: The "List[List[str]] or List[LivePlot]" could use a bit more documentation / explanation. Is there other documentation that could be linked? At least the following items:

None: Don't plot anything
True: Plot all UDF buffers, currently only 2D buffers
List[List[str]]: List of UDF buffer names to plot for each of the UDFs passed via the `udf` parameter
List[LivePlot]: LivePlot instance for each UDF, see TODO for more information on live plots.

uellue (commented):
Partial review, to be continued as soon as pending changes are merged. Hope this helps already! :-)

src/libertem/api.py
```diff
@@ -494,13 +495,18 @@ def run(
...
-                dataset: DataSet,
-                udf: UDF,
-                roi: np.ndarray = None,
-                corrections: CorrectionSet = None,
-                progress: bool = False,
-                backends=None) -> Dict[str, BufferWrapper]:
+    def run_udf(
+            self,
+            dataset: DataSet,
+            udf: UDF,
```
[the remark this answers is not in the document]
sk1p: Right, should be fixed.

src/libertem/api.py
```diff
@@ -510,13 +521,17 @@ def run_udf(self,
         .. versionchanged:: 0.6.0
             Added the :code:`corrections` and :code:`backends` parameter
 
+        .. versionchanged:: 0.7.0
+            Added the :code:`plots`, :code:`sync`, and :code:`iterate` parameters,
```
uellue, line 525: `iterate` is gone now

src/libertem/viz/mpl.py
```diff
@@ -0,0 +1,78 @@
...
+import numpy as np
+import matplotlib.pyplot as plt
+
+from .base import LivePlot
+
+
+logger = logging.getLogger(__name__)
+
+
+class MPLLivePlot(LivePlot):
```
uellue, line 13: Would it be possible to only create the plot window when `display()` is called to keep it consistent with bqp?

src/libertem/viz/base.py
```diff
@@ -0,0 +1,199 @@
...
+
+        if channel not in eligible_channels:
+            raise ValueError("channel %s not found or not plottable, have: %r" % (
+                channel, eligible_channels
+            ))
+
+        buf = UDFRunner.inspect_udf(udf, ds)[channel]
+        kind = buf.kind
+        if kind == 'sig':
+            shape = ds.shape.sig
```
uellue, line 149: What about additional `extra_shape`?

src/libertem/viz/base.py
```diff
@@ -0,0 +1,199 @@
...
+        elif kind == 'nav':
+            shape = ds.shape.nav
```
uellue, line 151: What about additional `extra_shape`?

uellue (approved):
Hi @sk1p, looks great! I am looking forward to use this from now on.

Two specific comments that we should perhaps discuss:

* Would it be possible to make the matplotlib version also only show the plot in the current cell when `display()` is called? That way we can work towards allowing the different plots as drop-in replacement, at least in the basic version without additional `*args` and `**kwargs` for the back-end. 
* How does the current implementation react to `extra_shape` to sig and nav? Would one have to pass in a postprocess method to put it in the correct shape?

As discussed, I can take care of the rest, including updating examples etc. If it is OK, I'd take over this PR once the two points above are clarified.

uellue:
Discussion with @sk1p :

1. `display()` should create the plots
2. "stress test" and possibly extension on blobfinder to see how to handle really complex buffers and derived plots
3. Percolate through LiberTEM once interface is frozen: examples, documentation etc
4. Take care of LiberTEM-live
5. Take care of SSB

uellue:
@sk1p In the latest push is a tentative solution to allow easy plotting of complex UDFs. Looking forward to your feedback! :-)

uellue:
Quick remark on performance: The damage calculation barely shows up in profiling. Main work in UDFRunner is executing tasks; main work in plotting with MPLLive2DPlot is drawing the canvas.

uellue:
TODO: damage for kind='nav' plus ROI. Since creating a boolean from NaN creates True, the parts that are not in the ROI might always be marked as damaged.

src/libertem/viz/base.py
```diff
@@ -0,0 +1,218 @@
...
+        udf : UDF
+            The UDF instance this plot is associated to. This needs to be
+            the same instance that is passed to :meth:`~libertem.api.Context.run_udf`.
+
+        roi : numpy.ndarray or None
+            Region of interest (ROI) that the UDF will be run on. This is necessary for UDFs
+            where the `extra_shape` parameter of result buffers is a function of the ROI,
+            such as :class:`~libertem.udf.raw.PickUDF`.
+
+        channel : str or function udf_result -> ndarray
```
sk1p, line 141: `channel` is now a bit more complicated - docstring needs an update. Channel can also be `None`, in which case the first channel will be plotted.

I was thinking if it's maybe possible to keep the simple "postprocessing"-style for the easy/common case where you just want to throw a function on the result before visualization. Something I just came up with:

```python
# easiest way: no channel at all, first channel is plotted
Live2DPlot(dataset=ds, udf=udf, channel=None)

# channel as str
Live2DPlot(dataset=ds, udf=udf, channel="intensity")

# channel w/ postprocessing fn
Live2DPlot(dataset=ds, udf=udf, channel=("intensity", np.log1p))

# complete control, including combinations of channels, damage etc.
Live2DPlot(dataset=ds, udf=udf, channel=my_extraction_function)
```

Thoughts?

sk1p (commented):
I'm quite happy with the changes already, some comments added below.

uellue:
Thx for the feedback @sk1p! I just overlapped with some stuff I worked on today, see latest push. Looking forward to your feedback there as well, it hopefully simplifies changing the plotting back-end! 

Regarding "Default visualization for complex numbers": For ptychography we use a side-by-side plot of amplitude and phase, which works pretty well IMO. Using a colorwheel would also be possible, but can be harder to interpret so we don't use it for ptychography and only offer it as an option for COM.

With your proposed API it would be easy to just throw in the channels and functions to show abs and angle, so I guess that fixes it. The example also shows how to arrange plots nicely if one wants more control. :-)

src/libertem/viz/base.py
uellue: Oh, this looks great! Will complicate things abit behind the scenes, but should be manageable for users and would allow passing functions through the `plots=...` parameter in the `Context` methods. I'll work on that next.

uellue:
Discussion with @sk1p: Update of general examples with new plotting methods is separated in #1020 so that we can finish releasing 0.7 and work on the dependencies in LiberTEM-live and Ptychography 4.0.

Live plotting for Analysis will likely be related to #1013.

src/libertem/analysis/com.py
```diff
@@ -200,35 +200,58 @@ def get_generic_results(self, img_sum, img_y, img_x):
...
                        key="y_imag", title="y [imag]", desc="y component of the center"),
             ])
         else:
-            f = CMAP_CIRCULAR_DEFAULT.rgb_from_vector((x_centers, y_centers, 0))
+            damage = damage & np.isfinite(x_centers) & np.isfinite(y_centers)
+            if np.count_nonzero(damage) > 0:
+                vmax = np.sqrt(np.max(x_centers[damage]**2 + y_centers[damage]**2))
+            else:
+                vmax = 1
+            f = CMAP_CIRCULAR_DEFAULT.rgb_from_vector((x_centers, y_centers, 0), vmax=vmax)
```
sk1p, line 218: Why do we need a `vmax` here now? I guess because `rgb_from_vector` doesn't like an empty array as input?

src/libertem/api.py
```diff
@@ -17,42 +18,54 @@
 from libertem.analysis.point import PointMaskAnalysis
 from libertem.analysis.masks import MasksAnalysis
 from libertem.analysis.base import AnalysisResultSet, Analysis
-from libertem.udf.base import UDFRunner, UDF
+from libertem.udf.base import UDFRunner, UDF, UDFResults
 from libertem.udf.auto import AutoUDF
+from libertem.utils.async_utils import async_generator
+from libertem.viz.mpl import MPLLive2DPlot
```
sk1p, line 24: Would it be possible to import this class lazily, to keep initial import time minimal? Maybe only if plotting is used at all? Could keep `self.plot_class` as `None` by default, and only when it is initialized in `_prepare_plots`, import `MPLLive2dPlot`. What do you think?

src/libertem/udf/masks.py
```diff
@@ -38,49 +81,83 @@ class ApplyMasksUDF(UDF):
 
     .. versionadded:: 0.4.0
     '''
+    # '''
```
sk1p, line 84: Is this commented docstring needed for some reason? If so, why?

tests/viz/test_mpl.py
```diff
@@ -0,0 +1,31 @@
...
+        matplotlib.pyplot,
+        'subplots',
+        mock.MagicMock(return_value=(mock.MagicMock(), mock.MagicMock()))
+    )
+    monkeypatch.setattr(lt_ctx, 'plot_class', MPLLive2DPlot)
+    lt_ctx.run_udf(dataset=default_raw, udf=udf, plots=True)
+    matplotlib.pyplot.subplots.assert_called()
+
+
+def test_bqp_nodisplay(monkeypatch, lt_ctx, default_raw):
```
sk1p, line 22: test function name?

sk1p (commented):
This is really good, thanks for taking over the PR :)

Some comments added inline, but looks almost ready to merge! :sparkles:

src/libertem/analysis/com.py
uellue: It _could_ happen that `damage` is empty for some reason, for example if the input data contains NaN. I wrote the code when I istarted introducing `damage` and it wasn't always set properly. IIRC an empty `damage` trips the calculation of `vmax`. If it is OK I'd prefer to leave it as-is, it is probably safer.

sk1p: Sounds good to keep as-is, maybe add a small comment?

src/libertem/udf/masks.py
uellue: No, this was a leftover from debugging a docstring issue. Thx! :-)

tests/viz/test_mpl.py
uellue: Thx, nice catch! :-)

src/libertem/analysis/com.py
uellue: Not sure how other analyses, plotting etc react to input data with NaN. It might trip for similar reasons. Would it be OK to leave that to later if/when it actually creates an issue? At that point one could introduce an all-NaN dataset to the test circuit.

src/libertem/api.py
uellue: I've turned `plot_class` into a property, leave it `None` by default and load the default class upon access. That way external users can always access it, but it is only loaded if needed. Sounds good? :-)

src/libertem/analysis/com.py
sk1p: Ok! Most analyses right now use `visualize_simple`, which should properly ignore those NaNs.

sk1p:
Thx, merging! :sparkles:

PR #1019, opened 2021-04-19
K2IS: reshaping, sync offset and time series support

### LINKED ISSUE #911: Allow specifying nav_shape, sig_shape and sync_offset for K2IS
Continuation of #793.

`nav_shape` and `sig_shape` can be easily implemented. `sync_offset` implementation was almost finished, check [this](https://github.com/LiberTEM/LiberTEM/pull/793#issuecomment-687870313) comment and [this](https://github.com/LiberTEM/LiberTEM/pull/793#discussion_r443501353) one.

AnandBaburajan:
Adds support for K2IS time series data sets.
@sk1p / @uellue, if the implementation is correct, I can continue with fixing #911 in this PR itself.

Fixes #911 

## Contributor Checklist:

changes were made to the GUI)

## Reviewer Checklist:


src/libertem/io/dataset/k2is.py
```diff
@@ -541,17 +547,27 @@ def __init__(self, path, io_backend=None):
...
-        # skip_frames is applied after synchronization.
-        self._skip_frames = -1
         self._files = None
         self._sync_offset = 0
 
     def _do_initialize(self):
         self._files = self._get_files()
+        with dm.fileDM(_get_gtg_path(self._path), on_memory=True) as dm_file:
+            nav_shape_y = dm_file.allTags.get('.SI Dimensions.Size Y')
+            nav_shape_x = dm_file.allTags.get('.SI Dimensions.Size X')
```
sk1p, line 557: Is there a reason this can't stay in its own method?

src/libertem/io/dataset/k2is.py
```diff
@@ -567,10 +583,14 @@ def _do_initialize(self):
...
-            return (int(dm_file.allTags['.SI Dimensions.Size Y']),
-                    int(dm_file.allTags['.SI Dimensions.Size X']))
+    def _get_frame_count(self):
+        with self._get_syncer().sectors[0] as sector:
+            first_block = sector.first_block()
+            last_block = sector.last_block()
+        first_frame_id = int(first_block.header['frame_id'])
+        last_frame_id = int(last_block.header['frame_id'])
+        frame_count = last_frame_id - first_frame_id
+        return (frame_count,)
```
sk1p, line 593: To make this re-usable (and to conform to its name), this could return an `int` instead of a `tuple`

src/libertem/io/dataset/k2is.py
```diff
@@ -635,7 +655,7 @@ def get_cache_key(self):
     def get_diagnostics(self):
         with self._get_syncer().sectors[0] as sector:
             est_num_frames = sector.filesize // BLOCK_SIZE // BLOCKS_PER_SECTOR_PER_FRAME
```
sk1p, line 657: This could now use the new `_get_frame_count` method, right? Or is `est_num_frames` significantly different from `_get_frame_count`?

sk1p (commented):
Thanks, this looks good! I've added a few comments and questions. After they are addressed (and a changelog entry is added), this can be merged.

src/libertem/io/dataset/k2is.py
AnandBaburajan: Ah, right. I could use a setter.

src/libertem/io/dataset/k2is.py
AnandBaburajan: Yup, thanks.

sk1p:
>  I can continue with fixing #911 in this PR itself.

Ah sorry, missed this - yes, that works for me, too!

src/libertem/io/dataset/k2is.py
AnandBaburajan: `est_num_frames` is the total number of frames before synchronization and `_get_frame_count` is the number of useful frames after sync, if I understand correctly.

sk1p: Ah right :+1:

AnandBaburajan: For the time series data I have, `est_num_frames` and `_get_frame_count` differ by 1 frame. But in regular K2 IS data, there's a large difference because of many blocks without the shutter active flag set, if I understand correctly. I haven't fully understood why there's a 1 frame difference but I'm looking into `sync_sectors`'s 'have_overlap, finding next frame' part.

AnandBaburajan:
@sk1p I've a doubt. I now understand `K2Syncer` except one thing:

"first_block_offsets # 3" (calculated by `sync_to_first_frame`) is the final `Sector.first_block_offset` (according to https://github.com/LiberTEM/LiberTEM/blob/master/src/libertem/io/dataset/k2is.py#L306) so I've been scratching my head over how "first_block_offsets # 2 and # 1" (calculated by `sync_sectors`) have any effect over `Sector.first_block_offset`.

sk1p:
> "first_block_offsets # 3" (calculated by `sync_to_first_frame`) is the final `Sector.first_block_offset` (according to https://github.com/LiberTEM/LiberTEM/blob/master/src/libertem/io/dataset/k2is.py#L306) so I've been scratching my head over how "first_block_offsets # 2 and # 1" (calculated by `sync_sectors`) have any effect over `Sector.first_block_offset`.

So, as far as I remember, `sync_sectors` sets the offset to the lower bound of valid blocks. It modifies the `first_block_offset`, which is used by subsequent calls to `get_block_by_index`, for example via this chain of calls:

`sync_to_first_frame` → `first_block_with_search` → `get_block_by_index`

So in most cases, this won't make a difference, I think, because we just go ahead and set overwrite the offset with the one having the `shutter_active` flag set.

But if, for example, the `shutter_active` flag is set for the whole data set, this takes care that we skip incomplete frames at the beginning of the data set. That's also what this `have_overlap` is actually doing: it looks at the next 32 blocks (after syncing to the right block id) and checks if they all belong to the same frame. If they don't, then the current frame is incomplete and is discarded.

AnandBaburajan:
> It modifies the `first_block_offset`, which is used by subsequent calls to `get_block_by_index`

Oh right! I should have looked into `get_block_by_index` which is being used by `first_block_with_search`.

> But if, for example, the `shutter_active` flag is set for the whole data set, this takes care that we skip incomplete frames at the beginning of the data set. That's also what this `have_overlap` is actually doing: it looks at the next 32 blocks (after syncing to the right block id) and checks if they all belong to the same frame. If they don't, then the current frame is incomplete and is discarded.

Thanks!

AnandBaburajan:
> we skip incomplete frames at the beginning of the data set

I checked the recent K2 time series data and the data used for tests, and found out that their last frame is incomplete too.
Is that a possibility too? If yes, I can work on handling this case.
Or could that be an incorrect finding? I checked the last 32 blocks of a sector after sync and found blocks with different `frame_id`s.

sk1p:
> > we skip incomplete frames at the beginning of the data set
> 
> I checked the recent K2 time series data and the data used for tests, and found out that their last frame is incomplete too.
> Is that a possibility too? If yes, I can work on handling this case.
> Or could that be an incorrect finding? I checked the last 32 blocks of a sector after sync and found blocks with different `frame_id`s.

I think that should be correct. Would make sense to skip those incomplete frames at the end, too.

uellue:
@AnandBaburajan @sk1p could you perhaps give a quick update on what is still to do here? Can we merge it soon or are there still open points?

uellue:
Tentatively assigning this to the 0.8 milestone. Or would it be relevant for ptychography or live processing?

AnandBaburajan:
> @AnandBaburajan @sk1p could you perhaps give a quick update on what is still to do here? Can we merge it soon or are there still open points?

I've finished skipping the incomplete frame at the end and currently implementing `sync_offset`. I think I'll finish this by tomorrow.

AnandBaburajan:
I've implemented reshaping, sync offset correction and support for time series data sets. But I need some help with the following error while putting a negative `sync_offset` in the GUI. The error doesn't show up while using the API on a notebook or running the tests, only when using the GUI:

```
[2021-05-13 16:20:22,788] INFO [libertem.web.dataset.prime_numba_caches:38] starting warmup
distributed.worker - WARNING -  Compute Failed
Function:  prime_numba_cache
args:      ()
kwargs:    {}
Exception: StopIteration()

[2021-05-13 16:20:22,807] ERROR [asyncio.default_exception_handler:1707] Exception in callback _chain_future.<locals>._set_state(<Future pendi...25cd4e190>()]>, <Future at 0x...StopIteration>) at /home/anand/anaconda3/envs/x/lib/python3.8/asyncio/futures.py:354
handle: <Handle _chain_future.<locals>._set_state(<Future pendi...25cd4e190>()]>, <Future at 0x...StopIteration>) at /home/anand/anaconda3/envs/x/lib/python3.8/asyncio/futures.py:354>
  File "/home/anand/anaconda3/envs/x/lib/python3.8/asyncio/events.py", line 81, in _run
    self._context.run(self._callback, *self._args)
  File "/home/anand/anaconda3/envs/x/lib/python3.8/asyncio/futures.py", line 356, in _set_state
    _copy_future_state(other, future)
  File "/home/anand/anaconda3/envs/x/lib/python3.8/asyncio/futures.py", line 332, in _copy_future_state
    dest.set_exception(_convert_future_exc(exception))
TypeError: StopIteration interacts badly with generators and cannot be raised into a Future
```

uellue:
> I've implemented reshaping, sync offset correction and support for time series data sets. But I need some help with the following error while putting a negative `sync_offset` in the GUI. The error doesn't show up while using the API on a notebook or running the tests, only when using the GUI:
> 
> ```
> [2021-05-13 16:20:22,788] INFO [libertem.web.dataset.prime_numba_caches:38] starting warmup
> distributed.worker - WARNING -  Compute Failed
> Function:  prime_numba_cache
> args:      ()
> kwargs:    {}
> Exception: StopIteration()
> 
> [2021-05-13 16:20:22,807] ERROR [asyncio.default_exception_handler:1707] Exception in callback _chain_future.<locals>._set_state(<Future pendi...25cd4e190>()]>, <Future at 0x...StopIteration>) at /home/anand/anaconda3/envs/x/lib/python3.8/asyncio/futures.py:354
> handle: <Handle _chain_future.<locals>._set_state(<Future pendi...25cd4e190>()]>, <Future at 0x...StopIteration>) at /home/anand/anaconda3/envs/x/lib/python3.8/asyncio/futures.py:354>
> Traceback (most recent call last):
>   File "/home/anand/anaconda3/envs/x/lib/python3.8/asyncio/events.py", line 81, in _run
>     self._context.run(self._callback, *self._args)
>   File "/home/anand/anaconda3/envs/x/lib/python3.8/asyncio/futures.py", line 356, in _set_state
>     _copy_future_state(other, future)
>   File "/home/anand/anaconda3/envs/x/lib/python3.8/asyncio/futures.py", line 332, in _copy_future_state
>     dest.set_exception(_convert_future_exc(exception))
> TypeError: StopIteration interacts badly with generators and cannot be raised into a Future
> ```

Hi @AnandBaburajan, I can reproduce the issue on my system. For my particular dataset `Capture52_.gtg`, it happens when I set the sync offset to 0 or a negative number. It does work fine when I set it to 1 or any larger number. As just discussed (reproduced here for others to see), the root cause could be `BasePartition.get_tiles()` returning an empty generator so that `libertem.common.numba.prime_numba_cache()` raises `StopIteration` when it tries to get a tile.

Additionally, an empty `ds.get_partitions()` generator could also be a root cause for the same reason.

For comparison I did the same test with an MIB file. Here, the same error occurs when I set the sync offset to a negative number.

uellue:
For comparison I tested the MIB file with the current master. The same issue occurs, so it is apparently not related to changes in this PR.

uellue:
@AnandBaburajan the culprit is actually `yield from io_backend.get_tiles(...)` in `BasePartition.get_tiles()` returning an empty generator if the sync offset is negative. Hope that helps! :-)

AnandBaburajan:
> For comparison I tested the MIB file with the current master. The same issue occurs, so it is apparently not related to changes in this PR.

Right, negative `sync_offsets` aren't working in the GUI. I should have tested the GUI before #793 was merged (^_^")

> the culprit is actually yield from io_backend.get_tiles(...) in BasePartition.get_tiles() returning an empty generator if the sync offset is negative. Hope that helps! :-)

Yup, thank you! I'll look into it.

AnandBaburajan:
> Right, negative `sync_offsets` aren't working in the GUI. I should have tested the GUI before #793 was merged (^_^")

I was wrong, the `sync_offset` does work in the GUI because when I removed [this](https://github.com/LiberTEM/LiberTEM/blob/master/src/libertem/web/dataset.py#L43) line and loaded a dataset with a -ve `sync_offset`, it worked fine. I think [this](https://github.com/LiberTEM/LiberTEM/blob/master/src/libertem/common/numba.py#L252) line could be the reason behind this issue because the frame at `roi[0]` would always be empty for a dataset with -ve `sync_offset`.

> > the culprit is actually yield from io_backend.get_tiles(...) in BasePartition.get_tiles() returning an empty generator if the sync offset is negative. Hope that helps! :-)
> 
> Yup, thank you! I'll look into it.

Now I'm not sure what would be the proper solution for this - should I make some changes at `prime_numba_cache` and/or should I do something when `get_tiles` returns an empty generator?

uellue:
Hi @AnandBaburajan 

> Now I'm not sure what would be the proper solution for this - should I make some changes at `prime_numba_cache` and/or should I do something when `get_tiles` returns an empty generator?

Ah I see, the ROI only contains tiles outside the actual dataset for `prime_numba_cache()`.

How are patched regions at the end of a dataset handled? Are the tiles simply skipped (empty generator correct) or are empty tiles generated (generator should never be empty for a ROI that is not empty)? The beginning and the end of a dataset should behave the same if the availabledata doesn't fill the complete nav shape.

if an empty generator is correct, one could simply get the first tile of the first partition for the entire dataset without any ROI for a quick fix. https://github.com/LiberTEM/LiberTEM/blob/efe6310bec8501bc53223962db3c7967a8b3c086/src/libertem/udf/base.py#L861-L874

Maybe one can use `libertem.udf.base.NoOpUDF` instead of `SumUDF` since that would always request default tiles, which should be always fast to supply?

@sk1p do you think there would be any performance penalties compared to using a ROI with only one frame?

AnandBaburajan:
> How are patched regions at the end of a dataset handled? Are the tiles simply skipped (empty generator correct) or are empty tiles generated (generator should never be empty for a ROI that is not empty)?

The tiles are skipped for missing data.

> If an empty generator is correct, one could simply get the first tile of the first partition for the entire dataset without any ROI for a quick fix.

[Here](https://github.com/LiberTEM/LiberTEM/blob/master/src/libertem/common/numba.py#L252), I can take into account the `sync_offset` of the dataset too, to get the first non-empty frame.

uellue:
> [Here](https://github.com/LiberTEM/LiberTEM/blob/master/src/libertem/common/numba.py#L252), I can take into account the `sync_offset` of the dataset too, to get the first non-empty frame.

That's a good idea, `get_sync_offset_info()` is implemented in the `DataSet` base class. Even if sync offset is not available for a specific dataset implementation, it should always be possible to query this information for a dataset. :-) 👍

AnandBaburajan:
> [Here](https://github.com/LiberTEM/LiberTEM/blob/master/src/libertem/common/numba.py#L252), I can take into account the `sync_offset` of the dataset too, to get the first non-empty frame.

This didn't work because in case of large -ve `sync_offset`s, the first partition might not have any non-empty tiles/frames.

> one could simply get the first tile of the first partition for the entire dataset without any ROI for a quick fix.

So I went with the above idea.

uellue:
> So I went with the above idea.

Ok, great! :-) In case we notice an impact on performance, we can still change the code to use a ROI and consume partitions until we have one that is not empty instead of using the first. :-)

src/libertem/common/numba.py
```diff
@@ -248,27 +248,31 @@ def enable_caching(self):
 def prime_numba_cache(ds):
     dtypes = (np.float32, None)
     for dtype in dtypes:
-        roi = np.zeros(ds.shape.nav, dtype=bool).reshape((-1,))
-        roi[0] = 1
+        roi = np.ones(ds.shape.nav, dtype=bool).reshape((-1,))
```
uellue, line 251: Would `roi = None` work? It should give the same result as `roi = np.ones(...)`, but skip the whole ROI-related logic of datasets.

src/libertem/common/numba.py
```diff
@@ -248,27 +248,31 @@ def enable_caching(self):
...
+                tiling_scheme = neg.get_scheme(
+                    udfs=udfs,
+                    partition=p,
+                    read_dtype=dtype,
+                    roi=roi,
+                    corrections=corrections,
+                )
+                for t in p.get_tiles(tiling_scheme=tiling_scheme):
+                    found_first_tile = True
+                    break
```
uellue, line 278: Oh, this is already iterating through partitions! In that case one should be able to set a ROI with `roi[max(-sync_offset, 0)] = True`, right? As long as the ROI is set for sth within the dataset, it should yield a tile at some point, right?

uellue (commented):
Nice, it looks like this is almost finished! :-)

👍

src/libertem/common/numba.py
AnandBaburajan: Yep, thanks. I'll do that.

uellue:
Thank you for taking care of this, looks good to me! Sync offset confirmed to work on my system. Only TODO items are a changelog entry and the rebuilt production client, and then this should be ready for merging. 🚀 🎆

AnandBaburajan:
> Thank you for taking care of this, looks good to me! Sync offset confirmed to work on my system. Only TODO items are a changelog entry and the rebuilt production client, and then this should be ready for merging.

Should we wait for Alex's review? I wonder if he would have any suggestions on the `K2Syncer` and `_set_sync_offset()`.

uellue:
> Should we wait for Alex's review? I wonder if he would have any suggestions on the `K2Syncer` and `_set_sync_offset()`.

Tests pass and the feature works. I just had another look specifically at the `K2Syncer`. The only suggestion would perhaps be to add a comment with a high-level description of what that code does and why. Other than that, it looks like your changes fit in the general architecture of the K2 IS dataset. If there is something to improve, one can always do follow-up changes.

TL; DR: Let's be bold! :-)

uellue:
Perfect, merging! 🚀

sk1p:
Congrats on getting this finished!

> Maybe one can use `libertem.udf.base.NoOpUDF` instead of `SumUDF` since that would always request default tiles, which should be always fast to supply?

We actually should do both `SumUDF` and  `NoOpUDF` independently from each other - because right now, I think the `dtype=None` case (native `DataSet` dtype) is not warmed up. You can observe this in the GUI: after warm-up, the PACBED sum starts immediately, but if you then switch to pick mode, there is a noticeable delay.

PR #1025, opened 2021-04-28
WIP: `BufferWrapper`: support for partitions that slice the signal dimensions

sk1p:
This is needed for tiled processing of (live) data from detectors that deliver their data as independent streams for each detector part. It works by only allocating for the signal slice, and then shifting the tile slices to the local coordinate system of the partition (in signal dims).

The tests are not minimal yet - I'll work on implementing support in the memory dataset for this kind - then we can test at the interfaces.

## Contributor Checklist:

## Reviewer Checklist:


sk1p:
Extracted into #1153 and #1152

PR #1038, opened 2021-06-01
Add nbval to test example notebooks

sk1p:
The plan is to run the example notebooks in the data test job, and check that the output stays the same over time. That means we will have less work when releasing, because we don't need to check the examples manually.

Piggy-backing on this PR are some changes that reduce the number of threads that are started in numba, OpenBLAS, MKL etc. - they were unused anyways, and caused problems when running many notebooks/dd clusters in parallel. Best way to monitor number of threads of the current user: `watch -n1 "ps ux -L | wc -l"`

## TODO

- [x] Make `--nbval-lax` pass
- [x] Make `--nbval` pass
- [x] Add notebooks tox environment to the data tests

## Contributor Checklist:

## Reviewer Checklist:


sk1p:
@uellue the CoM notebook fails because the `print(result.field)` is numerically different - do you have an idea why and/or how to fix this? Maybe because of slightly different versions? `numpy` patch version is different, for example; so maybe also different BLAS version etc. - the `np.allclose(...)` check at the end still succeeds. Maybe we can ignore the output of those two cells?

Also, numba doesn't like what we are doing with the `NUMBA_NUM_THREADS` env var - I'll have another look at that on Thursday probably.

Other than those failures, this should be ready for review.

src/libertem/utils/threading.py
```diff
@@ -63,6 +62,51 @@ def set_numba_threads(n):
...
+        The maximum number of threads
+    set_numba : bool, optional
+        Set the :code:`'NUMBA_NUM_THREADS'` environment variable. If None, determine if
+        Numba has been initialized and only set it if not.
+        Numba may throw a :class:`RuntimeError` if the environment is
+        altered after the number of threads has been set already.
+    """
+    if set_numba is None:
+        from numba.np.ufunc import parallel
+        set_numba = not parallel._is_initialized
```
sk1p, line 93: As we are accessing an implementation detail here, maybe we could replace this with the following?

```suggestion
        set_numba = not getattr(parallel, "_is_initialized", True)
```

Then, in case numba changes the implementation, we assume that numba is already initialized and don't trigger an error, possibly at the expense of "resource waste" by starting too many threads.

What do you think?

uellue (approved):
Thank you for taking care of this!

The change I just pushed fixes the NUMBA_NUM_THREADS issue on moellenstedt for me. It is not a 100 % solution since it just doesn't set the number of threads in the fixture. That means the workers probably use the default value for Numba.

One could set an environment variable for the entire Pytest run to limit the number of threads per worker to fix this.

For the numerical issue one could use https://numpy.org/doc/stable/reference/generated/numpy.printoptions.html to limit the digits in the notebook output.

src/libertem/utils/threading.py
uellue: Thank you, good point! :-)

sk1p:
> For the numerical issue one could use https://numpy.org/doc/stable/reference/generated/numpy.printoptions.html to limit the digits in the notebook output.

That's a good idea, I'll adjust the notebooks accordingly!

uellue:
We should probably also have some contributor documentation on how to update the notebooks properly so that tests will pass?

sk1p:
> We should probably also have some contributor documentation on how to update the notebooks properly so that tests will pass?

True, I'll add a section. A link to https://nbviewer.jupyter.org/github/computationalmodelling/nbval/blob/master/docs/source/index.ipynb covers most of it, plus mentioning the `%precision` magic.

uellue:
Awesome, merging!

uellue:
🚀

Issue #1055, opened 2021-06-28
DMDataset `files=` argument not checked before iteration, leading to uninformative error messages


matbryan52:
The DMDataset class has a `files=List[str]` argument which is distinct from most other Dataset classes which use a `path=str` argument (even if they are multi-file datasets). As a new user to LiberTEM this caught me out, as the `files=` argument is not checked at all before it is used, leading to some confusing error messages.

Firstly, we have this:
```
        if len(files) == 0:
            raise DataSetException("need at least one file as input!")
```
which is of course passes without error for either `str` or `list` arguments (and co-incidentally fails for `pathlib.Path` objects).

The next thing to be called is `DMDataset._get_filesize` which iterates over `files`. The issue occurs as a string can be iterated over just like a list, leading to a `FileNotFoundError` raised by the `os` library when it tries to _stat_ the first letter of the provided path, leading to an error like the below:
```
FileNotFoundError: [Errno 2] No such file or directory: 'h'
```
for a `files` argument of `/home/...`.

We could solve this as such:
```
assert isinstance(files, (list, tuple)), "Files argument to DMDataset must be an iterable of file paths"
```

This of course raises the additional question of why DMDataset must always work on a stack of files. The docstring references the strange memory layout of DM 4DSTEM files, which I agree with, but an EELS image, for example, has a navigation and (single) signal dimension which could be loaded using `DMDataset(path=path_to_dm3)` for consistency with other formats.

This also makes me wonder if the Dataset interfaces could be better standardised, but that's for a different Issue.

PR #1058, opened 2021-06-29
Add files= argument checking to DMDataset

matbryan52:
## Contributor Checklist:

* [ ] I have added/updated documentation for all user-facing changes **[no user-facing changes]**
* [ ] I have included the [rebuilt production build of the client](https://libertem.github.io/LiberTEM/contributing.html?#building-the-client) (only if changes were made to the GUI) **[not applicable]**

## Reviewer Checklist:

References Issue https://github.com/LiberTEM/LiberTEM/issues/1055

I can't run the tests on the IO code due to lack of access. I don't imagine the one line assert will cause a problem though!

### DIFF
--- a/src/libertem/io/dataset/dm.py
+++ b/src/libertem/io/dataset/dm.py
@@ -125,6 +125,8 @@ class DMDataSet(DataSet):
             self._nav_shape = tuple(scan_size)
         self._filesize = None
         self._files = files
+        assert isinstance(files, (list, tuple)),\
+               f"files argument must be an iterable of file paths recieved {type(files)}"
         if len(files) == 0:
             raise DataSetException("need at least one file as input!")
         self._fileset = None
--- a/src/libertem/io/dataset/dm.py
+++ b/src/libertem/io/dataset/dm.py
@@ -125,8 +125,9 @@ class DMDataSet(DataSet):
             self._nav_shape = tuple(scan_size)
         self._filesize = None
         self._files = files
-        assert isinstance(files, (list, tuple)),\
-               f"files argument must be an iterable of file paths recieved {type(files)}"
+        if not isinstance(files, (list, tuple)):
+            raise DataSetException("files argument must be an iterable\
+                                    of file paths, recieved {type(files)}")
         if len(files) == 0:
             raise DataSetException("need at least one file as input!")
         self._fileset = None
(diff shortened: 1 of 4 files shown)


src/libertem/io/dataset/dm.py
```diff
@@ -125,6 +125,8 @@ def __init__(self, files=None, scan_size=None, same_offset=False, nav_shape=None
             self._nav_shape = tuple(scan_size)
         self._filesize = None
         self._files = files
+        assert isinstance(files, (list, tuple)),\
+               f"files argument must be an iterable of file paths recieved {type(files)}"
```
uellue, line 129: An `assert` will be skipped when running Python in optimized mode. That means it can be used for internal consistency  checks. When validating user input, one should throw, for example, a `ValueError` instead.

uellue (changes requested):
Looks good, many thanks! Just two flake8 issues from pre-commit.ci. The variable `e` issue can be solved by just omitting the `... as  e` from the `with` block.

Let me know if you'd like some pointers on how to do these checks locally! One can have it activated in the IDE and/or run `pre-commit`.

@sk1p the file ending issue could have been fixed by `pre-commit` automatically. Should we activate it for all files resp. skip only the ones that we know are problematic?

(For context, we started using pre-commit.ci only last week, so we are still fine-tuning the settings and workflows)

src/libertem/io/dataset/dm.py
matbryan52: Understood, I've modified accordingly (`DataSetException` as that seems to be the style in the file).

sk1p (approved):
Looks good to me!

matbryan52:
I've turned on flake8 linting now and have installed/run pre-commit. I had forgotten the former and hadn't determined the latter was a requirement yet!

sk1p:
> the file ending issue could have been fixed by `pre-commit` automatically. Should we activate it for all files resp. skip only the ones that we know are problematic?

I'm not sure, I think I need to play a bit more with `pre-commit` locally - right now I'm mostly relying on the CI version of it - which could also do the auto fixing, though.

We need to be careful to exclude files where these whitespace changes don't make sense - like the automatically generated client files etc.

sk1p:
> > Should we activate it for all files resp. skip only the ones that we know are problematic?

What is actually needed to do this, @uellue? At least on my installation, file ending issues _are_ fixed automatically.

Issue #1045, opened 2021-06-09
Fast turn-around of CI pipeline


uellue:
We are supporting quite a number of Python versions now (good) and have a lot of tests (good). As a downside, CI takes a long time to complete. We should consider options to speed things up:

* Throw hardware at the problem and run tests in parallel where possible. See also #683 
* Use several CI services in parallel
* Compartmentalize the code and tests to only run tests that are relevant for a specific change. For example, continue splitting LiberTEM in sub-modules #261 
* Review tests: Reduce time for long-running tests; eliminate unnecessary tests. See also discussion in #914 
* Possibly only run the full test matrix for release candidates

sk1p:
Using pre-commit.ci for flake8, mypy etc. #1036 should also help, as we don't "waste" azure pipeline jobs for these things.

sk1p:
Identifying slow tests, take from the currently slowest run (example: py36 on Windows):

`libertem` main test suite:

```
============================ slowest 10 durations =============================
27.72s call     tests/corrections/test_corrset.py::test_patch_corr_odd
17.04s call     tests/test_analysis_clust.py::test_cluster_analysis
13.93s call     tests/server/test_copy.py::test_copy_notebook
13.72s call     tests/corrections/test_detector.py::test_detector_patch_large
12.01s call     tests/template/test_clust_template.py::test_clust_default
11.98s call     tests/io/test_cached.py::test_with_roi
10.42s call     tests/test_analysis_raw.py::test_pick_analysis
9.75s call     tests/udf/test_stddev.py::test_stability
9.60s call     tests/common/test_numba.py::test_numba_prime
9.43s call     tests/template/test_sd_template.py::test_sd_roi
= 710 passed, 231 skipped, 21 deselected, 4 xfailed, 26 warnings in 619.45s (0:10:19) =
```

`nbval`:

```
============================= slowest 10 durations =============================
73.37s call     examples/radialfourier.ipynb::Cell 8
34.02s call     examples/pcmclustering.ipynb::Cell 19
10.80s call     examples/Introduction to UDFs.ipynb::Cell 7
8.09s call     examples/Introduction to UDFs.ipynb::Cell 25
6.96s call     examples/simple libertem API.ipynb::Cell 5
6.24s call     examples/pcmclustering.ipynb::Cell 11
5.96s call     examples/pcmclustering.ipynb::Cell 8
5.00s call     examples/pcmclustering.ipynb::Cell 2
[...]

Issue #1041, opened 2021-06-09
First analysis on a fresh installation


uellue:
Running the first analysis on a fresh installation still takes a long time (over 30 s, vs. under 1 s after warmed up).

Refs #886

## How to reproduce

Tested on Win10, Python 3.9.5, wheel from Continuous (current master)

* Install LiberTEM in a fresh environment
* Open a file, for example the BLO example
* Run a simple analysis

## What happens

* LiberTEM takes a long time for the first analysis

## What should happen

* LiberTEM makes a good first impression with fast results

Ideas for fixing, with @sk1p:

* Port Numba code to Cython where possible
* Ship compiled Numba in wheel
* Warm up Numba cache during setup

PR #1062, opened 2021-07-07
Cache tox environments on Windows and Mac OS X

sk1p:
Adds parameters that allows recreating the tox environment, and/or run slow tests:

image

This is meant for more comprehensive testing before cutting a release. Some additional tests that took over 10 seconds on any of the platforms are now also marked as slow.

This also disables some of the Azure specific code coverage reporting, as that is sometimes really slow. In general, it appears that some of the jobs are just scheduled on systems that are I/O overloaded - even checking out the source code, unpacking the pip cache etc. takes a long time sometimes:

image

image

image

Refs #1045

## TODO

- [ ] Check coverage when manually running slow tests - maybe merge w/ coverage of previous `tox` run

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,7 +21,3 @@ repos:
   rev: "3.2"
   hooks:
   - id: pyroma
-- repo: https://github.com/miki725/pre-commit-twine-check
-  rev: '47f7fff370d1a0c83e4291aa323a6380b69d7b75'
-  hooks:
-  - id: twine-check
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -10,6 +10,17 @@ variables:
     value: $(Pipeline.Workspace)/.pip
   - name: npm_config_cache
     value: $(Pipeline.Workspace)/.npm
+  - name: tox_dir
+    value: $(Build.Repository.LocalPath)/.tox
+
+parameters:
+  # On windows and mac os x, creating tox environments is costly,
+  # so we disable it by default, caching the tox environments for the
+  # next run. They are rebuilt when setup.py or test_requirements.txt change
+  - name: recreate_tox
+    displayName: Recreate tox environments? (Win/Mac)
+    type: boolean
+    default: false
 
 stages:
   - stage: test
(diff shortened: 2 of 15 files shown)


sk1p:
@uellue @twentyse7en some of your tests have been disabled for CI in this PR. Could you please review these, and see if there may be another way to test these parts of the code?

In total, we are losing about 1% coverage with this PR (the currently displayed number may not be accurate, as I just triggered a manual run including the slow tests).

sk1p:
This is a run including the slow tests: https://dev.azure.com/LiberTEM/LiberTEM/_build/results?buildId=843&view=results

The coverage looks off a bit, as it should now be back to "normal". It's possible that the coverage for `slow`-marked tests needs to be merged manually, I'll have a look into this later.

uellue:
@sk1p Looks good! Did we see improvements already? I'll have a look at the tests that you marked as slow. Should I push them into this PR?

sk1p:
> Did we see improvements already?

Yes - the slow tests take about 2 to 3 minutes, if you check the link above. With slow tests disabled, we are in total maybe 5 minutes faster, if things go right, so we still have quite some room for improvement, maybe in the long tail...

uellue:
When evaluating the test times we should keep in mind that some tests use session-scoped fixtures that are quite heavy to create and may also require significant warm-up when they are used for the first time -- namely the local Dask executor. The same goes for tests that use Numba-compiled code. Some test will always draw the short straw and be a lot slower than the others. The clustering template happens to be run first from all the template tests, so that contributes to it appearing very slow.

sk1p:
> Some test will always draw the short straw and be a lot slower than the others. 

Yes, indeed - that's probably also the reason `test_com_with_zero_frames` is so slow. This goes into the direction of #1041 - if the numba compilation load is reduced, first-use times should improve a lot.

> The clustering template happens to be run first from all the template tests, so that contributes to it appearing very slow.

I'm not convinced that's the whole story, as the "normal" clustering analysis test was also really slow. The `template/test_pick_analysis` test is much, much faster than, let's say the SD or clustering notebook tests...

uellue:
> I'm not convinced that's the whole story

Yes, it contributes only a part of the extra time. Discussion with @sk1p: It seems to be the import of sklearn. A candidate to move to a separate module, at some point.

sk1p:
Coverage merging via `--cov-append` seems to work (according to `--cov-report=term`), we are just hitting a bug in codecov.io it seems, similar to [this one](https://community.codecov.com/t/codecov-uploads-from-azure-pipelines-are-failing-with-build-numbers-do-not-match/1100). I'm not too worried about that, I think it's a glitch in codecov.

PR #1067, opened 2021-07-14
Escape globs to allow special characters in file names

### LINKED ISSUE #1066: Non-standard characters in path trip MIB reader
## How to reproduce

Try to open an MIB dataset via the HDR file where the path contains, for example, square brackets.

## What happens

`get_filenames()`

https://github.com/LiberTEM/LiberTEM/blob/6b6ca850fd71ba3f00ed420267b7d97311343394/src/libertem/io/dataset/mib.py#L94-L107

fails to find the MIB files and returns an empty list:

```python

from libertem.io.dataset.mib import get_filenames

get_filenames(r'E:\LargeData\LargeData\ER-C-1\groups\data_science\data\reference\MIB[FAIL]\20200518 165148\default.hdr')

[]
```

It works if the folder `MIB[FAIL]` is renamed to `MIB`.

## What should happen

All characters that can "legally" be part of a file system path work.

We should also check other multi-file datasets such as K2IS and FRMS6 if they suffer from a similar issue. Probably the globs or regexp should be fixed. We should also test that in CI.

uellue:
Closes #1066

* A bit of refactoring to make the code more uniform and easier to test
* Skip only tests that need test data since the glob tests don't need it

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -361,6 +361,23 @@ def raw_data_8x8x8x8_path(tmpdir_factory):
 
 
 @pytest.fixture
+def naughty_filename():
+    '''
+    Return a string with many special charaters that tests the limits
+    of what the file system allows on that platform. This allows to stress-test
+    globs or regular expressions applied to paths.
+    '''
+    system = platform.system()
+    # See https://en.wikipedia.org/wiki/Filename#Comparison_of_filename_limitations
+    if system == 'Windows':  # NTFS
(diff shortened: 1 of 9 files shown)


sk1p:
Re-running data tests since the locale settings have changed.

sk1p (approved):
LGTM!

Issue #805, opened 2020-06-08
SEQ: Support for loading excluded pixels from XML file


sk1p:
Extracted from #778: implement loading excluded pixels from XML file. Should be quite easy, only needs to convert the ranges specified in the XML to a `sparse.COO` array.

PR #1077, opened 2021-08-10
Implemented support for loading excluded pixels from XML, issue #805

DYVZ8X:
## Contributor Checklist:

## Reviewer Checklist:


sk1p:
I'll do a more detailed review later today, but for now: congrats on getting it working and the PR submitted! As you can see in the "checks" section on this page, there are still some issues that need to be solved, primarily formatting the code to make flake8 happy. You can run `pre-commit` locally, too, as [described in the documentation](https://libertem.github.io/LiberTEM/contributing.html#code-style).

Other than that, you can also [already add a changelog entry](https://libertem.github.io/LiberTEM/contributing.html#code-contributions) for this change.

sk1p:
The `check_example_notebooks` failure is not your fault - this is caused by runtime variations because of more than usual load on the CI system.

src/libertem/io/dataset/seq.py
```diff
@@ -269,32 +441,53 @@ def _maybe_load_mrc(self, path):
...
+    def _load_xml_from_file(self, path):
+        if not os.path.exists(path + ".Config.Metadata.xml"):
+            return None
+        if not os.path.exists(path + ".metadata"):
+            return None
+        else:
+            tree = ET.parse(path + ".Config.Metadata.xml")
+            root = tree.getroot()
+            file = open(path + ".metadata", mode="rb")
+            met = file.read()
```
sk1p, line 453: When reading from files, the context manager form should be preferred to properly close the file in all cases:

```suggestion
            with open(path + ".metadata", mode="rb") as file:
                met = file.read()
```

(this is especially important for some Python implementations that don't immediately GC objects once they go out of scope, like PyPy)

src/libertem/io/dataset/seq.py
```diff
@@ -269,32 +441,53 @@ def _maybe_load_mrc(self, path):
...
 
     def get_correction_data(self):
         return CorrectionSet(
             dark=self._dark,
             gain=self._gain,
+            excluded_pixels=self._excluded_pixels,
         )
 
+    def get_sig_shape(self):
+        return self._sig_shape
```
sk1p, line 473: Is this used anywhere? I think this can be removed (also, `ds.shape.sig` should be easy enough)

src/libertem/io/dataset/seq.py
```diff
@@ -130,6 +130,181 @@ def _get_image_offset(header):
...
+    """
+    sx, sy = a.shape
+    sxc = sx // binning * binning
+    syc = sy // binning * binning
+    # crop:
+    ac = a[:sxc, :syc]
+    return ac.reshape(ac.shape[0] // binning, binning, ac.shape[1] // binning, binning).sum(3).sum(1)
+
+
+def cropping(arr, start_size, req_size, offsets):
```
sk1p, line 230: Again, function name could be clearer

sk1p (changes requested):
Thank you for taking care of this. The code is now much easier to follow and understand than the previous versions I looked at, good work!

I've added quite a few comments inline, please let me know if anything is still unclear.

There is a page on corrections in our documentation, here:

https://libertem.github.io/LiberTEM/reference/corrections.html

It would be good to add the new feature to this page, too, in addition to a changelog entry.

Thanks!

tests/io/datasets/test_seq.py
```diff
@@ -120,12 +120,86 @@ def test_positive_sync_offset(default_seq, lt_ctx):
...
+        "UnbinnedFrameSizeY": 2048,
+        "OffsetX": 1024,
+        "OffsetY": 1024,
+        "HardwareBinning": 2
+    }
+    test_arr = np.zeros((1024, 1024), dtype=bool)
+    test_arr[644] = True
+    test_arr[643] = True
+    test_arr[:, 143:145] = True
+    test_arr[:, 138] = True
```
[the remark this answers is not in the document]
DYVZ8X: Since it uses the 4096*4096 map we will we first crop it to the size of 2048 in the middle part(because of the offset) according to the metadata that means we substract 1024 from the column coordinates of the bad pixel map which will be 286 287 288 276 if we bin these values we will get 143 143 144 138

DYVZ8X:
@bdalevin Hi, I have noticed that in the test XML files there are two badPixelMaps one is binned and the other is not and I was wondering what is the purpose of the (smaller) binned one since its defect values could be calculated from the unbinned, bigger bad pixel map directly.

uellue:
Hi @DYVZ8X, right now the entire SEQ test module is skipped if the test data is not available. Since your tests for excluded pixels don't depend on the data, you can modify the SEQ module to only skip tests that actually require the data.

Here's a diff with an example where I did this for FRMS6: https://github.com/LiberTEM/LiberTEM/pull/1067/commits/0543acb6737ccf4c909c78a083a10e959983a17e#diff-da2368c6194672d7394faed3a77a97da61edf62dcb57b81541805967174bf4cc

The "skipif" decorator is not assigned to the special `pytestmark` variable which applies it to the entire module, but any other variable, in this case `needsdata`. Now `@needsdata` can be added to any test function that requires data.

sk1p:
@DYVZ8X FYI the "cognitive complexity" measurement is not a hard guideline - if we feel that one of the functions is too complex, we will clear this up in the code review. If this is currently a blocker for you, feel free to ignore these codeclimate issues for now.

uellue:
@DYVZ8X could you maybe try to bundle related changes in commits so that the number of commits stays lower? In particular, it would be a lot better to first run all checks and tests, and commit only after everything passes resp. after a significant milestone is reached. That way we don't end up with individual commits for each flake8 problem, like your last two commits. Just to be clear, separate commits for milestones, separating unrelated changes and commits for intermediate review are perfectly OK! It is just about finding the right balance here.

tests/io/datasets/test_seq.py
```diff
@@ -5,44 +5,49 @@
...
-        io_backend=MMapBackend(),
-    )
-
-    ds.set_num_cores(4)
-    assert tuple(ds.shape) == (8, 8, 128, 128)
-    return ds
+        ds.set_num_cores(4)
+        assert tuple(ds.shape) == (8, 8, 128, 128)
+        return ds
+    return None
```
sk1p, line 46: Test fixtures can't be marked with `skipif` like this - adding `@needsdata` to the individual tests should be enough, the other changes to this fixture should be undone.

tests/io/datasets/test_seq.py
```diff
@@ -5,44 +5,49 @@
...
+from libertem.io.dataset.seq import xml_map_sizes, xml_unbinned_map_maker, xml_map_index_selector
+from libertem.io.dataset.seq import xml_defect_coord_extractor, array_cropping
 from libertem.common import Shape
 from libertem.common.buffers import reshaped_view
 from libertem.udf.sumsigudf import SumSigUDF
 from libertem.udf.raw import PickUDF
 from libertem.io.dataset.base import TilingScheme, BufferedBackend, MMapBackend
 
 from utils import get_testdata_path, ValidationUDF
+import defusedxml.ElementTree as ET
```
sk1p, line 18: `defusedxml` is a third-party package, so it needs to be added to `setup.py` as a dependency if we want to start using it. Right now, it is probably installed because one of our other dependencies also depends on it, but that is just sheer luck, and can change in the future.

tests/io/datasets/test_seq.py
```diff
@@ -5,44 +5,49 @@
 import pytest
 
 from libertem.executor.inline import InlineJobExecutor
-from libertem.io.dataset.seq import SEQDataSet
+from libertem.io.dataset.seq import SEQDataSet, _load_xml_from_string, xml_defect_data_extractor
+from libertem.io.dataset.seq import xml_map_sizes, xml_unbinned_map_maker, xml_map_index_selector
+from libertem.io.dataset.seq import xml_defect_coord_extractor, array_cropping
```
sk1p, line 10: You can combine these into a single import statement, something like this:


```python
from libertem.io.dataset.seq import (
    SEQDataSet, _load_xml_from_string,
    xml_defect_data_extractor, # and so on
)
```

tests/io/datasets/test_seq.py
```diff
@@ -120,12 +129,217 @@ def test_positive_sync_offset(default_seq, lt_ctx):
...
+    xy_map_sizes, map_sizes = xml_map_sizes(bad_pixel_maps)
+    unbinned_x, unbinned_y = xml_unbinned_map_maker(xy_map_sizes)
+    map_index = xml_map_index_selector(unbinned_x, unbinned_y, map_sizes)
+    defects = xml_defect_coord_extractor(bad_pixel_maps[map_index], map_index, map_sizes)
+    expected_defects = {"rows": [['2311', '2312'], ['1300']],
+                        "cols": [['2300', '2301'], ['600']],
+                        "pixels": [['100', '230']],
+                        "size": (4096, 4096)
+                        }
+    assert defects.__eq__(expected_defects) is True
```
sk1p, line 339: Why not `defects == expected_defects`?

tests/io/datasets/test_seq.py
```diff
@@ -120,12 +129,217 @@ def test_positive_sync_offset(default_seq, lt_ctx):
...
+    assert xy_map_sizes == xy_map_sizes_expected
+
+
+def test_unbinned_map_maker():
+    xml = '<?xml version="1.0" encoding="UTF-8" standalone="yes"?>' \
+          '<BadPixels>' \
+          '<BadPixelMap Rows="4096" Columns="4096"></BadPixelMap>' \
+          '<BadPixelMap Rows="2048" Columns="2048"></BadPixelMap>' \
+          '<BadPixelMap Rows="2048" Columns="1024"></BadPixelMap>' \
+          '<BadPixelMap Binning="2" Rows="4080" Columns="4096"></BadPixelMap>' \
```
sk1p, line 268: Hm, why do the `BadPixelMap`s have different aspect ratios? Don't they always conform to the sensor aspect ratio?

sk1p: Same question for the other test cases below

tests/io/datasets/test_seq.py
```diff
@@ -120,12 +129,217 @@ def test_positive_sync_offset(default_seq, lt_ctx):
...
+
+def test_xml_excluded_pixels_unbinned():
+    xml_string = '<?xml version="1.0" encoding="UTF-8" standalone="yes"?>' \
+                 '<Configuration><PixelSize></PixelSize><DiffPixelSize>' \
+                 '</DiffPixelSize><BadPixels><BadPixelMap Rows="4096" ' \
+                 'Columns="4096"><Defect Rows="2311-2312"/><Defect Rows="3413-3414"/><Defect ' \
+                 'Column="2311"/></BadPixelMap><BadPixelMap Binning="2" ' \
+                 'Rows="2048" Columns="2048"><Defect Rows="1155-1156"/><Defect ' \
+                 'Rows="1706-1707"/></BadPixelMap></BadPixels>' \
+                 '</Configuration>'
```
sk1p, line 155: For readability, it would be really nice to have the XML properly formatted. For example like this:

```suggestion
    xml_string = '''
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<Configuration>
    <PixelSize></PixelSize><DiffPixelSize></DiffPixelSize>
    <BadPixels>
    <BadPixelMap Rows="4096" Columns="4096">
        <Defect Rows="2311-2312"/>
        <Defect Rows="3413-3414"/>
        <Defect Column="2311"/>
    </BadPixelMap>
    <BadPixelMap Binning="2" Rows="2048" Columns="2048">
        <Defect Rows="1155-1156"/>
        <Defect Rows="1706-1707"/>
    </BadPixelMap>
    </BadPixels>
</Configuration>
    '''
```

This may be a bit more "verbose", but also much better readable than before

sk1p (changes requested):
I've added some comments - I didn't have a detailed look at the test cases again, as it doesn't really make sense until we know why there is a separate `BadPixelMap` that is binned. Let me know if my comments are clear and if you have any questions!

tests/io/datasets/test_seq.py
DYVZ8X: I was not sure if a case where hey have different aspect ratios possible so I implemented it in the xml_map_index_selector's else branch

sk1p:
Thanks, merged!

uellue:
Is the XML format documented somewhere? I am consolidating the changelog and couldn't find a public resource that describes the file format.

DYVZ8X:
> Is the XML format documented somewhere? I am consolidating the changelog and couldn't find a public resource that describes the file format.

I didn't write a documentation for the xml file, but If I should then please let me know where.

uellue:
@DYVZ8X thank you for the feedback! For me it is still a bit unclear where these XML files come from. Are they written by the acquisition software or does a user write them "by hand"? Who else consumes these files? The point is not so much to define the format, but to specify what kind of workflow we support here. Since we don't define the format ourselves and also don't write these files, I'd prefer to link to an external resource such as official Direct Electron documentation. However, Google didn't find anything on this.

Did you obtain sample XML files from somewhere or did you write them following a specification from Direct Electron? Where did you obtain the former or the latter?

DYVZ8X:
> @DYVZ8X thank you for the feedback! For me it is still a bit unclear where these XML files come from. Are they written by the acquisition software or does a user write them "by hand"? Who else consumes these files? The point is not so much to define the format, but to specify what kind of workflow we support here. Since we don't define the format ourselves and also don't write these files, I'd prefer to link to an external resource such as official Direct Electron documentation. However, Google didn't find anything on this.
> 
> Did you obtain sample XML files from somewhere or did you write them following a specification from Direct Electron? Where did you obtain the former or the latter?

In the test_seq.py I wrote them by hand, but to test if really everything is working fine I got one from you at the beginning of my internship, also I think these files are written by the software as it would be really hard to write a binary metadata file for the xml file.

uellue:
Ok, I see. I've included finding and linking proper documentation of the XML format from the vendor as a FIXME. CC @bdalevin :-)

PR #1100, opened 2021-09-03
Unset environment variables properly after setting thread count

### LINKED ISSUE #1053: Starting a Context might set the Numba threads to 1 for the main process
## How to reproduce

* Start a normal Context
* Start a Context with InlineJobExecutor
* Run a UDF on the normal context
* Run a UDF on the inline Context

Tested on Win 10, Python 3.7.10 with 0.7 release.

## What happens

The inline Context warns that the Numba threads are fixed to 1 and can't be set to the number of cores

## What should happen

The workers of the normal Context use one Numba thread, while the main process can still use multithreading.

uellue:
Closes #1053

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/docs/source/changelog/bugfix/numbathreads.rst
+++ b/docs/source/changelog/bugfix/numbathreads.rst
@@ -1,0 +1,6 @@
+[Bugfix] Reset environment properly when setting thread count
+=============================================================
+
+* Make sure the environment is cleaned up properly in the main process
+  after adjusting it temporarily to launch Dask workers with thread
+  count settings (:issue:`1053`, :pr:`1100`)
--- a/src/libertem/utils/threading.py
+++ b/src/libertem/utils/threading.py
@@ -120,3 +120,6 @@ def set_num_threads_env(n=1, set_numba=None):
         yield
     finally:
         os.environ.update(old_env)
+        for key in env_keys:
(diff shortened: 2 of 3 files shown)


uellue:
The env setting in pytest.ini is confirmed to not influence the Numba thread count with the inline executor. It is always the number of physical cores, apparently, even if that is more than 4.

uellue:
> The env setting in pytest.ini is confirmed to not influence the Numba thread count with the inline executor. It is always the number of physical cores, apparently, even if that is more than 4.

Ah I see:

```
..\..\..\.conda\envs\libertem\lib\site-packages\_pytest\config\__init__.py:1233
  c:\users\dwebe\.conda\envs\libertem\lib\site-packages\_pytest\config\__init__.py:1233: PytestConfigWarning: Unknown config option: env
```

uellue:
> The env setting in pytest.ini is confirmed to not influence the Numba thread count with the inline executor. It is always the number of physical cores, apparently, even if that is more than 4.

Actually, the Numba threads are set properly, just that the UDF meta information contains the CPU count independent of threading settings for Numba etc. The error message came from missing installation of `pytest_env`. Working with this env setting still requires a small fix in the test case.

Issue #1026, opened 2021-05-05
Code audit: Assignment to out-of-range views


uellue:
LiberTEM relies heavily on views to direct array operations to the correct part of a result buffer. During work on https://github.com/LiberTEM/LiberTEM-live, a potential class of bugs was discovered that we should defend against.

NumPy executes the following code without triggering an error:

```python

a = np.array((1, 2, 3))

a[13:17] = 23
```

1. Specifying an out-of-range view creates an empty view.
2. Assigning to an empty view works without error and discards the value

Assigning to an empty view and silently discarding the value hides potential bugs in the size and/or index calculation for the buffer or view, which in turn can lead to wrong results.

The current LiberTEM code base probably doesn't suffer from this bug since the results are checked against references that are calculated with other methods. However, it makes finding the root cause of such bugs in new or changed code much harder.

As a first fix, `BufferWrapper.set_view_for_...` triggers an error if the resulting view is empty in a development version by @sk1p.

To close this Issue, we should do the following:

- [ ] Merge the fix by @sk1p 
- [ ] Assert that the created view is not only not empty, but exactly of the expected shape.
- [ ] Include tests that confirm that an error is thrown
- [ ] Audit the LiberTEM code base for other potential occurrences of empty views, in particular assignments to empty views.

uellue:
I've played around a bit with that and made some progress for buffers. However, we assign to slices also in other places, most notably I/O code.

All this code is pretty performance-critical, so one should probably avoid bloating these code paths with too many checks. For that reason I've posted the question on Stack Overflow if there is an elegant solution in NumPy already: https://stackoverflow.com/questions/69047282/how-to-prevent-accidental-assignment-into-empty-numpy-views

uellue:
It seems that manually checking is our best option for now. I've raised https://github.com/numpy/numpy/issues/19831 since checking within NumPy would probably be much easier and faster to implement. The validation code for slices will probably start to mirror code within NumPy that derives the actual shape and slice from the provided array indices.

PR #1102, opened 2021-09-06
Defend against out of bounds slices

uellue:
Confirm that the views have the expected shape.

Refs #1026

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/common/buffers.py
+++ b/src/libertem/common/buffers.py
@@ -380,7 +380,8 @@ class BufferWrapper:
     def _get_slice(self, slice: Slice):
         real_slice = slice.get()
         result = self._data[real_slice]
-        # Defend against #1026
+        # Defend against #1026 (internal bugs), allow deactivating in
+        # optimized builds for performance
         assert result.shape == tuple(slice.shape) + self.extra_shape
         return result
 
@@ -447,7 +448,8 @@ class BufferWrapper:
                 offset = partition_slice.origin[0]
             result_start = tile_slice.origin[0] - offset
             result_stop = result_start + tile_slice.shape[0]
-            # Defend against #1026
+            # Defend against #1026 (internal bugs), allow deactivating in
+            # optimized builds for performance
             assert result_start < len(self._data)
             assert result_stop <= len(self._data)
             # shape: (1,) + self._extra_shape
--- a/src/libertem/common/buffers.py
+++ b/src/libertem/common/buffers.py
@@ -377,6 +377,13 @@ class BufferWrapper:
             raise RuntimeError("Cache is not empty, has to be flushed")
         return self._data
 
+    def _get_slice(self, slice: Slice):
+        real_slice = slice.get()
+        result = self._data[real_slice]
+        # Defend against #1026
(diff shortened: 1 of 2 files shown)


src/libertem/common/buffers.py
```diff
@@ -377,17 +377,23 @@ def get_view_for_dataset(self, dataset):
             raise RuntimeError("Cache is not empty, has to be flushed")
         return self._data
 
+    def _get_slice(self, slice: Slice):
+        real_slice = slice.get()
+        result = self._data[real_slice]
+        # Defend against #1026
+        assert result.shape == tuple(slice.shape) + self.extra_shape
```
sk1p, line 384: I think this should not be an `assert`, but an explicit exception `raise`. Maybe a `ValueError`?

src/libertem/common/buffers.py
```diff
@@ -439,6 +447,9 @@ def get_view_for_tile(self, partition, tile):
                 offset = partition_slice.origin[0]
             result_start = tile_slice.origin[0] - offset
             result_stop = result_start + tile_slice.shape[0]
+            # Defend against #1026
+            assert result_start < len(self._data)
+            assert result_stop <= len(self._data)
```
sk1p, line 452: Again, maybe explicit `raise`s instead, so they keep working even in "optimized" settings?

sk1p (commented):
:+1: on the general implementation, some comments added. Any way to trigger the issue with a test case? Maybe with "invalid" input data, like a partition with a slice that lies outside of the dataset?

Just as a note for cross-referencing, the changes overlap with #1025, which will need to be adjusted to fit (and should possibly be merged soon, too).

src/libertem/common/buffers.py
uellue: Since the buffer slicing is used internally and is used in a performance-critcal code path, my thinking was that the `assert` catches issues with upstream use within LiberTEM and then allows to run an optimized build that skips the checks. Or do we have scenarios where an end-user provides the slices?

uellue:
> Maybe with "invalid" input data, like a partition with a slice that lies outside of the dataset?

Yes, just triggering it outright should be easy. I was just wondering if we can trigger it somehow in more high-level code to make sure it helps prevent such errors. Do you remember how we triggered this in LiberTEM-live? 🙈

src/libertem/common/buffers.py
sk1p: That's true. Maybe you could add a comment for future readers?

sk1p:
> > Maybe with "invalid" input data, like a partition with a slice that lies outside of the dataset?
> 
> Yes, just triggering it outright should be easy. I was just wondering if we can trigger it somehow in more high-level code to make sure it helps prevent such errors. Do you remember how we triggered this in LiberTEM-live?

I don't remember the details - I think it was just a somehow-broken `Partition` subclass?

uellue:
> I think it was just a somehow-broken `Partition` subclass?

That works to trigger it! :-)

sk1p (approved):
LGTM!

PR #1107, opened 2021-09-08
Dynamic task scheduling; coordinate calculation improvement

### LINKED ISSUE #1108: Performance of `self.meta.coordinates` with ROI and `process_frame`
Accessing coordinates when processing with an ROI is very expensive, in particular with `process_frame` where each coordinate is calculated individually and the cache is not effective. The root cause seem to be in
https://github.com/LiberTEM/LiberTEM/blob/master/src/libertem/io/dataset/base/coordinates.py and https://github.com/LiberTEM/LiberTEM/blob/master/src/libertem/io/dataset/base/roi.py

A few ideas for improvement:

* If any coordinate is requested for a partition, calculate all coordinates within the ROI immediately and cache them since most likely they will be needed.
* Make sure the  `_roi_to_nd_indices` function is also accelerated. Possibly use  `_roi_to_indices` and use [numba_unravel_index_multi](https://github.com/LiberTEM/LiberTEM/blob/master/src/libertem/common/numba/__init__.py#L54)?

uellue:
Tasks are assigned to workers at the time they are submitted. That means
a fast worker such as a GPU will get the same number of tasks assigned as a CPU
if all tasks are submitted at once.

Here, the ability of dask.distributed to add tasks dynamically to as_completed
is used to schedule more tasks only after previous tasks are completed. That way
they will be assigned to the worker that just became free so that faster workers
receive more work.

That yields a 20 % speed bonus in synthetic tests on an 18-core machine
with one GPU where a cuda worker completes a task 10x faster than a CPU worker.

Drive-by change: Speed up coordinate calculation for `process_frame()` with ROI and correct partition shape in `UDFMeta`.

Closes #1108 

## Contributor Checklist:

## Reviewer Checklist:


uellue:
Accidentally pushed an unrelated change, maybe it is OK to review it together? :-)

PR #1111, opened 2021-09-20
Helper function to guess CoM paramters

uellue:
To be considered experimental!

The new functions help to guess CoM parameters by minimizing RMS curl and choosing divergence distributed towards lower values. To be tested more with real data!

## Contributor Checklist:

* [n/a] I have included the [rebuilt production build of the client](https://libertem.github.io/LiberTEM/contributing.html?#building-the-client) (only if changes were made to the GUI)

## Reviewer Checklist:


src/libertem/analysis/com.py
```diff
@@ -154,6 +172,84 @@ def magnitude(y_centers, x_centers):
...
+    # end.
+    div = divergence(corrected_y, corrected_x)
+    all_range = np.maximum(-np.min(div), np.max(div))
+    hist, bins = np.histogram(div, range=(-all_range, all_range), bins=5)
+    polarity_off = np.sum(hist[:1]) < np.sum(hist[-1:])
+    if polarity_off:
+        angle += 180
+    if angle > 180:
+        angle -= 360
+    return (angle, flip_y, np.mean(y_centers), np.mean(x_centers))
```
sk1p, line 250: Maybe we could use a `namedtuple` here instead? That would minimize errors from messing up the ordering (y/x vs x/y) when unpacking (and still allow unpacking like a normal tuple, if you really want that)

sk1p (commented):
Generally looks good to me, added two comments inline.

src/libertem/analysis/com.py
```diff
@@ -154,6 +174,130 @@ def magnitude(y_centers, x_centers):
...
+    -------
+    (straight, flipped)
+        Root mean square of the curl as a function of :code:`scan_rotation` from
+        0 to 359 degrees in steps of one, with :code:`flip_y=False` (straight)
+        and code:`flip_y=True` (flipped).
+    '''
+    straight = np.zeros(360)
+    flipped = np.zeros(360)
+    if roi is None:
+        # The last row and column contain artefacts
```
sk1p, line 208:
```suggestion
        # The last row and column contain artifacts
```

sk1p:
LGTM, thanks!

PR #1118, opened 2021-09-30
Caching threadpoolctl modules

### LINKED ISSUE #1117: `threadpoolctl.threadpool_limits(n)` is surprisingly expensive
Results from a profiling session of https://github.com/joblib/threadpoolctl/blob/ee7fad84615fbdf846a69a915530542b3f7efee9/threadpoolctl.py#L334:

```python
   334                                               def __init__(self, user_api=None, prefixes=None,  modules=None):
   335       250        122.0      0.5      0.0          if modules is None:
   336       250        162.0      0.6      0.0              self.prefixes = [] if prefixes is None else prefixes
   337       250         92.0      0.4      0.0              self.user_api = [] if user_api is None else user_api
   338                                           
   339       250         94.0      0.4      0.0              self.modules = []
   340       250     936453.0   3745.8     98.7              self._load_modules()
   341       250      11421.0     45.7      1.2              self._warn_if_incompatible_openmp()
   342                                                   else:
   343                                                       self.modules = modules
```

We should look into caching the modules. However, we also have to consider that the modules are possibly only loaded after an UDF task is executed, meaning we have to be careful when to cache.

- [x] Track the state of modules that `threadpoolctl` autoloads throughout script execution
- [x] Implement caching to re-use the module list

uellue:
Closes #1117

Resetting the module cache on a distributed executor causes
so much overhead that it negates any benefits. For that reason
this change accepts that the threading settings might
depend on UDF execution order under some circumstances.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/benchmarks/udf/test_simple_udf.py
+++ b/benchmarks/udf/test_simple_udf.py
@@ -56,9 +56,12 @@ def test_udf_noncontiguous_tiles(lt_ctx, backend, benchmark):
     group="udf overheads"
 )
 def test_overhead(shared_dist_ctx, benchmark):
-    ds = shared_dist_ctx.load('memory', data=np.zeros((1024, 1024, 2)), sig_dims=1)
-    # many partitions
-    ds.set_num_cores(32*1024)
+    ds = shared_dist_ctx.load(
+        'memory',
+        data=np.zeros((1024, 1024, 2)),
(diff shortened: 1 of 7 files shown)


sk1p (approved):
:sparkles:

PR #1128, opened 2021-10-07
Large UDF parameters on large files overrun worker memory

uellue:
The included notebook triggers excessive memory use on Dask workers.
The size of UDF parameters seems to be the reason.

## Contributor Checklist:

## Reviewer Checklist:


sk1p:
Is this now "fixed" by #1143? It should now work properly given that memory in the order of `n_workers*sizeof(A)` is available, right?

uellue:
Fix confirmed in current master! :rocket:

uellue:
The dist case is now actually faster than the inline one, impressive!

uellue:
...and this use case is also tested in the benchmarks for #1143, so no need for this reproducer anymore. :-)

PR #1136, opened 2021-10-15
Prototype that uses futures and dask arrays

uellue:
It emulates the basic steps of LiberTEM UDF: Cut the data in partitions,
calculate results dynamically per partition, assemble on main node.

Different from the current executor, partition and parameters are passed in as
function arguments so the scheduler has more insight and control over that.

CC @sk1p @matbryan52  :-)

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/prototypes/advancedexecutor/minimaldask.ipynb
+++ b/prototypes/advancedexecutor/minimaldask.ipynb
@@ -1,0 +1,405 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "suspected-strip",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "import distributed\n",
+    "import dask\n",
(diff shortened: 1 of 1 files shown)


sk1p:
> This would allow a user to supply their data either from a source external to LiberTEM, or use a util function like the existing make_dask_array on a LiberTEM Dataset to get an appropriate Dask array. I think it's already been discussed that there are obvious issues with re-chunking and having no control over how the array is loaded from disk etc. At its worse the root node could 'load everything' in one operation! [At least if we were in charge of creating the array via make_dask_array, we can set up a good load strategy.]

Sounds good :+1:

> I will have a try mocking something like this up, unless you're already working on something similar?

As Dieter will only be back next Monday, I'd say go ahead.

uellue:
Should we merge this as a testing ground, or should we close it?

sk1p:
I've merged it - I think it makes sense to keep it around for experimenting.

PR #1137, opened 2021-10-22
Prototype implementation DaskDataSet wrapper class

matbryan52:
This follows on from https://github.com/LiberTEM/LiberTEM/pull/1136 as a wrapper for Dask arrays which makes them compatible with the normal UDF interface. There are no modifications to current LiberTEM master, only a new dataset implementation.

The dataset builds LiberTEM partitions out of the existing Dask chunks, applying a merge-only strategy to avoid read amplification while making the chunk structure compatible with the existing LiberTEM Partition interface (and everything which goes with that). The major assumption for efficient performance of this class is that the chunks in the Dask array can all be computed individually without incurring upstream read amplification.

A major design factor was the fact that Partitions currently have to be contiguous in the flattened nav dimension, so if the Dask array is blocked in more than one nav dimension we are obliged to do a lot of merging (except when all higher nav dimensions are chunked like (1, 1, 1, 1, ...)). If we ever lift this requirement then this class could be made a lot more efficient.

See the accompanying notebook and the docstrings for more detail.

## Contributor Checklist:

## Reviewer Checklist:


src/libertem/io/dataset/dask.py
```diff
@@ -0,0 +1,483 @@
...
+        if new_chunking != array.chunks:
+            original_n_chunks = [len(c) for c in array.chunks]
+            chunksizes = get_chunksizes(array)
+            orig_min, orig_max = chunksizes.min(), chunksizes.max()
+            array = array.rechunk(new_chunking)
+            log.info('Applying re-chunking to increase minimum partition size. '
+                     f'n_blocks: {original_n_chunks} => {[len(c) for c in array.chunks]}. '
+                     f'Min chunk size {orig_min / 1e6:.1f} => {min_size / 1e6:.1f} MiB , '
+                     f'Max chunk size {orig_max / 1e6:.1f} => {max_size / 1e6:.1f} MiB.')
+        return array
```
uellue, line 234: We have similar code used in https://github.com/LiberTEM/LiberTEM/blob/d294d5d7b06dcb3ea6954c804b6eef53dc281fa5/src/libertem/io/dataset/hdf5.py#L390-L409

Maybe some of these functions can be re-used here? Possibly one can move https://github.com/LiberTEM/LiberTEM/blob/d294d5d7b06dcb3ea6954c804b6eef53dc281fa5/src/libertem/io/dataset/hdf5.py#L106-L133 to `libertem.io.utils`?

tests/io/datasets/test_dask_ds.py
```diff
@@ -0,0 +1,217 @@
...
+import dask.array as da
+
+from libertem.udf.sumsigudf import SumSigUDF
+from libertem.udf.sum import SumUDF
+from libertem.io.dataset.base import DataSetException
+from libertem.io.dataset.dask import DaskDataSet
+
+from utils import dataset_correction_verification
+from utils_dask import _mk_dask_from_delayed
+
```
uellue, line 12: Currently this is all tested with an `InlineJobExecutor`. Maybe at least one test can be on a distributed executor, for example like in https://github.com/LiberTEM/LiberTEM/blob/d294d5d7b06dcb3ea6954c804b6eef53dc281fa5/tests/io/datasets/test_frms6.py#L318-L336

The benchmark is already on a distributed context, so I guess it will work just fine. Having one test here makes sure it is run in CI.

matbryan52: Ah, I hadn't appreciated that detail of the fixturing. I'll make that change.

uellue: No worries, I'm already writing one!

src/libertem/io/dataset/dask.py
matbryan52: I'm looking into this. I believe the biggest difference I've seen so far is that in the current code I handle arrays with arbitrary chunking along an axis (not just uniform chunks with a remainder), but I need to see where we can make the code common.

tests/io/datasets/test_dask_ds.py
matbryan52: Thanks!

src/libertem/io/dataset/dask.py
uellue: Good point about the variable chunk size! Also, chunks in HDF5 can be quite small compared to Dask array chunks so it would not be advisable to iterate over all of them.

uellue:
By the way, with `math.prod()` vs `np.prod()` I decided it is time to fix this for good: https://github.com/LiberTEM/LiberTEM/pull/1145

uellue:
@matbryan52 I'll take care of a HyperSpy integration example now, if that suits you!

matbryan52:
> @matbryan52 I'll take care of a HyperSpy integration example now, if that suits you!

Of course, I'm looking into the HDF5 question for now.

src/libertem/io/dataset/dask.py
matbryan52: It seems like there is no support for non-uniform chunksizes along an axis in HDF5 except for the remainder chunk at the end.

When exporting using `da.to_hdf5` Dask or h5py seems to re-chunk each axis so that the normal chunksize is the same as the first chunksize, then lumps everything else into a remainder:

```python
dar = da.ones((59, 32, 32, 32), chunks=((12, 45, 2), -1, -1, -1))
da.to_hdf5('file.hdf5', '/data', dar)
print(h5py.File('file.hdf5')['data'].chunks)
>>> (12, 32, 32, 32)
```

uellue:
@matbryan52 the example triggers a warning we should look into: 

```python

dask_ds = ctx.load("dask", lazy_4d.data, sig_dims=2)
pick_a = ctx.create_pick_analysis(dataset=dask_ds, y=45, x=60)
pick_res = ctx.run(pick_a)
```

```
c:\users\dwebe\.conda\envs\libertem\lib\site-packages\distributed\worker.py:3877: UserWarning: Large object of size 215.36 MiB detected in task graph: 
  (<UDFTask [<class 'libertem.udf.raw.PickUDF'>]>, ' ... 528e80d7ef', 0)
Consider scattering large objects ahead of time
with client.scatter to reduce scheduler burden and 
keep data on workers

    future = client.submit(func, big_data)    # bad

    big_future = client.scatter(big_data)     # good
    future = client.submit(func, big_future)  # good
  warnings.warn(
```

200 MB is about the size of the dataset. Perhaps it gets baked into the Task somehow?

matbryan52 (commented):
Nice! Soon it will be possible to keep wrapping the same data in an infinite loop?

I also notice something strange with PickAnalysis the other day. Do you want me to try to debug it or are you already on it?

uellue:
> Soon it will be possible to keep wrapping the same data in an infinite loop?

Would be a good test for overheads! :-)

> Do you want me to try to debug it or are you already on it?

Maybe you can have a look first? I have to leave now. It feels like the ROI may have something to do with it. Otherwise we can look together next week. :-)

matbryan52:
OK, I'll see if I can reproduce it in a more minimal example.

matbryan52:
> OK, I'll see if I can reproduce it in a more minimal example.

https://gist.github.com/matbryan52/f2a092d6706e9dd28e52d1d8e0de0b0d

After looking through the code, the gist there pretty much reproduces what the current combination of Hyperspy+LiberTEM does, generating the same error.

Dask has a doc note about it: https://docs.dask.org/en/latest/array-creation.html#memory-mapping

I'm not sure there is a good way to get around this. There are hacky solutions like trying to re-create the memmap on each worker using the np.memmap.filename attribute, but actually if the object has already been wrapped as a dask array I think the information is lost anyway! I'm kind of surprised there is no Dask-provided memmap-based array creation routine which is backed by dask.delayed. I will look more into why this is.

uellue:
> Dask has a doc note about it:

Nice that you found it! https://github.com/dask/dask/pull/7380 https://github.com/dask/dask/issues/7355 😄 

> I'm not sure there is a good way to get around this.

Since we can't control what people do in their code and it "works for them"™️ with the default threaded Dask executor, we can consider adding a threaded executor for single-host setups, for example based on `concurrent.futures`. We can even consider making it the default if it works well. That way we'd be more tolerant to Dask arrays that are inefficient on a multiprocess cluster.

uellue:
> the gist there pretty much reproduces what the current combination of Hyperspy+LiberTEM does, generating the same error

Yes, it really seems so. Here's how it happens as an overview:

## BLO reader returns a read-only memory map

https://github.com/hyperspy/hyperspy/blob/b5815c01d303d6598484d97096622708b38e41e0/hyperspy/io_plugins/blockfile.py#L180-L297

## Hyperspy reading seems to be indirected here to a specific `Signal` sub-class

https://github.com/hyperspy/hyperspy/blob/b5815c01d303d6598484d97096622708b38e41e0/hyperspy/io.py#L663-L738

## `LazySignal` seems to use the same arguments in the constructor and attributes as `BaseSignal`:

https://github.com/hyperspy/hyperspy/blob/707e1f96d39bef369a4d5ed6b65e226c4d69e025/hyperspy/_signals/lazy.py#L87-L92

https://github.com/hyperspy/hyperspy/blob/707e1f96d39bef369a4d5ed6b65e226c4d69e025/hyperspy/signal.py#L2133-L2169

## `to_array()`, finally, seems to turn things into a Dask array with `dask.array.from_array()`:

https://github.com/hyperspy/hyperspy/blob/707e1f96d39bef369a4d5ed6b65e226c4d69e025/hyperspy/_signals/lazy.py#L49-L84

uellue:
The BLO reader would probably be rather easy to fix: With `lazy=True`, just return a Dask array with memory-mapped delayed chunks like in https://docs.dask.org/en/latest/array-creation.html#memory-mapping

In any case, LiberTEM should also work with `dask.array_from_array(numpy.memmap(...))`, and that means using a threaded executor.

sk1p (commented):
@matbryan52 I had a look at the I/O backend part of this PR and it looks good to me.

In `_adapt_chunking`, maybe the log level should be "warn" in case we hit a pathological perf case?

Thanks!

Btw. the restriction of only allowing whole signal frames per partition will probably lifted soon (see for instance #1025), which may help with some weird chunking configurations.

src/libertem/io/dataset/dask.py
sk1p: Ok, sounds like the best option is to keep separate functions for HDF5 and Dask

sk1p:
Let me know if I can help with anything for this PR - now that #1149 has landed, a few details might need to be changed. I could take care of rebasing onto master, if that makes sense. Let me know!

matbryan52:
> @matbryan52 I had a look at the I/O backend part of this PR and it looks good to me.

Thanks for taking a look! at least we're using the standard memmap backend now. 

One of the commits I've pushed [57c22c5](https://github.com/matbryan52/LiberTEM/commit/57c22c5580e633b2a049cf047bae683370450fce) is actually a note about a potential problem with the backend if the Dask array happens to be Fortran-ordered. Apparently `np.frombuffer` cannot read from a f-ordered array and just throws a ValueError. I originally added an `np.ascontiguousarray` to the `open` method but this can cause memory problems on the worker as we're forced to make a copy. For now I've just left it because I think it's an edge case. I was able to reproduce it by doing the following, though:

```python
import numpy as np
import dask.array as da
import libertem.api as lt

if __name__ == '__main__':
    with lt.Context() as ctx:
        dar = da.ones((20, 100, 512, 512), chunks=(2, -1, -1, -1), dtype=np.float32)
        ds = ctx.load('dask', dar, sig_dims=2)
        pick_a = ctx.create_pick_analysis(dataset=ds, y=4, x=27)
        pick_res = ctx.run(pick_a)
```

> In `_adapt_chunking`, maybe the log level should be "warn" in case we hit a pathological perf case?

I have pushed this change.

> Btw. the restriction of only allowing whole signal frames per partition will probably lifted soon (see for instance #1025), which may help with some weird chunking configurations.

That will be great to remove, the more chunking flexibility the better, really.

>Let me know if I can help with anything for this PR - now that #1149 has landed, a few details might need to be changed. I could take care of rebasing onto master, if that makes sense. Let me know!

Could you explain what you have in mind about what might need to be changed? The partition layout is constrained here by the array backing the DS, so we have a bit less flexibility than usual to re-partition optimally. We could of course adapt the partition merging (if any) so that we get a good tiling scheme, based on the ROI, of course, but that might be quite complex.

sk1p:
> (https://github.com/matbryan52/LiberTEM/commit/57c22c5580e633b2a049cf047bae683370450fce) is actually a note about a potential problem with the backend if the Dask array happens to be Fortran-ordered. Apparently `np.frombuffer` cannot read from a f-ordered array and just throws a ValueError. I originally added an `np.ascontiguousarray` to the `open` method but this can cause memory problems on the worker as we're forced to make a copy. For now I've just left it because I think it's an edge case. I was able to reproduce it by doing the following, though:
> 
> ```python
> import numpy as np
> import dask.array as da
> import libertem.api as lt
> 
> if __name__ == '__main__':
>     with lt.Context() as ctx:
>         dar = da.ones((20, 100, 512, 512), chunks=(2, -1, -1, -1), dtype=np.float32)
>         ds = ctx.load('dask', dar, sig_dims=2)
>         pick_a = ctx.create_pick_analysis(dataset=ds, y=4, x=27)
>         pick_res = ctx.run(pick_a)
> ```

I think we can leave this as an open issue for now - we will notice if this happens more often than we think. It could be fixed by calling a method on the `MMapFile` instead of using `np.frombuffer` - this method could be overridden by your `FakeDaskMMapFile`. Although I haven't tested what happens in the I/O code when it is presented with an 'F'-order array...

> > In `_adapt_chunking`, maybe the log level should be "warn" in case we hit a pathological perf case?
> 
> I have pushed this change.

:+1: thx!

> Could you explain what you have in mind about what might need to be changed? 

Because the `TilingScheme` is now calculated once per `DataSet` (with an approx. partition size as input), some methods have moved from `Partition` to `DataSet`:

- `Partition._get_decoder` is moved to `DataSet.get_decoder`
- `Partition.need_decode` is moved to `DataSet.need_decode` - that uses `DataSet.get_io_backend`, so you may need to move your `get_io_backend` impl to `DaskDataSet`
- The `Decoder` needs to be passed to the `Partition` in `DataSet.get_partitions` (to have a single point, in the `DataSet`, where the `Decoder` is constructed

That should be all changes required. The reasoning for moving these methods is that we need to know if we are decoding/copying for choosing the right tile size (it can be much larger if we are working directly on mmapped data vs. if we need to decode into an intermediate buffer) - and for that we need to instantiate the IO backend, together with the `Decoder`.

> The partition layout is constrained here by the array backing the DS, so we have a bit less flexibility than usual to re-partition optimally. We could of course adapt the partition merging (if any) so that we get a good tiling scheme, based on the ROI, of course, but that might be quite complex.

I think this is mostly important for the next step, when we do adaptive partition sizing (this is work in progress). If we can get this PR merged before that, I can take care that this is properly handled.

sk1p:
The test "failure" (hang) looks unrelated - I'll push an increase of the timeout into your branch (and mark the thread as daemon thread, so it doesn't keep the main thread from exiting if it hangs for some reason)

sk1p:
This should now be ready to be merged, correct? Would you be fine with me squashing the commits into one? We usually try to keep the commit history linear (by rebasing instead of merging the upstream master branch) - that's maybe something for the contribution docs, too...

matbryan52:
> This should now be ready to be merged, correct? Would you be fine with me squashing the commits into one? We usually try to keep the commit history linear (by rebasing instead of merging the upstream master branch) - that's maybe something for the contribution docs, too...

Of course, whatever you normally do. I tend to make a lot of small commits when I work but it makes for a very long PR I know!

sk1p:
Thanks, merged!

uellue:
🎉 🎉 🎉 Congratulations, a major milestone for LiberTEM! :-)

PR #1144, opened 2021-11-02
Docker packaging

### LINKED ISSUE #484: Make a docker image
https://github.com/LiberTEM/LiberTEM/packages?package_type=Docker

The GitHub "packages" feature is now generally available, which means we can easily publish docker images of LiberTEM directly into the GitHub docker registry. A possible starting point would be the AppImage build script, which could easily be converted to a Dockerfile. Good thing is we can skip the miniconda setup and directly start with the miniconda:alpine image, for example.

sk1p:
Fixes #484 

- Rename Dockerfile for dist tests to Dockerfile.dev
- WIP: scripts/release: add docker_tags command to find suitable tags
- update .dockerignore
- WIP: pipeline job that builds a docker image
- WIP: production docker image

## TODO

- [x] Proper tagging
- [x] Documentation

## Contributor Checklist:

## Reviewer Checklist:


sk1p:
Note: need to take care when merging, as the pipeline definitions most likely have already changed since I created this branch.

IIRC the missing part of this PR is mostly fixing the listen address for this specific case - we need to listen on `0.0.0.0`, but keep token auth disabled. Maybe add an `--insecure` command line flag to allow for this configuration, or something similar?

Also missing is proper image tagging (see the newly added `docker_tags` command in the release script - that needs to be used to tag and push the resulting image)

Possibly we could also push to the [github container registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry) in addition to docker hub, so we are not 100% dependent on docker hub continuing to work.

/cc @shabihsherjeel

uellue:
The bash script kind of works locally, but it is hard to test since `scripts/release` draws a lot of its info from the Azure env. Also, I don't have the token so I can't really test if pushing works. Let's first see if the pipeline runs through for the PR first.

After that, I'm a bit unsure. My only idea would be to just merge it and see what happens, also when baking RCs. 🙈

sk1p:
> Also, I don't have the token so I can't really test if pushing works

If you have a docker hub login, I can add you to the libertem organization there

azure-pipelines.yml
```diff
@@ -351,7 +351,7 @@ stages:
           docker pull libertem/libertem-dev || true
           docker build --cache-from=libertem/libertem-dev -t libertem/libertem-dev --build-arg BUILDKIT_INLINE_CACHE=1 ../../ -f Dockerfile.dev
           # make this succeed even if the access token is not available:
-          (docker login -u sk1p -p $(DOCKER_ACCESS_TOKEN) && docker push libertem/libertem-dev) || true
```
sk1p, line 354: FYI this worked fine just yesterday, as evident from the [last `libertem/libertem-dev` push](https://hub.docker.com/r/libertem/libertem-dev)

uellue:
If I understood the flow of events correctly, we are building and pushing the production image only in the second stage now, right? That means it is untested and completely independent of the dev version, if I understood correctly. A bit of a chicken-egg problem since we'd have to tag and push something before the dist tests, but don't know yet if it is sth we want to push, right?

I can think of three ways to solve this:

* Use a special tag, for example "triage"
* Use libertem-dev repository with a "testing" tag
* Use libertem-testing repository

Or is there a way to first build the image locally, then derive the test image from it, start up the workers, and only push it after the dist tests passed?

In the previous push it seemed that it tried to run the docke release also for a PR. It also does that for the dev image. Isn't that a security issue? A bad actor could replace our Dockerfile with `FROM:my_crypto_miner_pwn` and change `scripts/release` to push the image to `latest`, and it would be live in production for our users, right?

Generally, what about the tokens $(DOCKER_ACCESS_TOKEN), $(LT_RELEASE_UPLOAD_PYPI_PASSWORD), ...? Can't someone just make a PR and include "email me $TOKEN" in the azure-pipelines.yml?

uellue:
> change `scripts/release` to push the image to `latest`

I mean, change the bash script in azure-pipelines.yml

azure-pipelines.yml
uellue: Hm strange, I got this: https://dev.azure.com/LiberTEM/LiberTEM/_build/results?buildId=1394&view=logs&j=ced89aaa-f671-5884-bb56-a0d56c033fe1&t=2127ab87-9cad-521d-7c58-c5a20da19f7f&l=1634

sk1p:
> In the previous push it seemed that it tried to run the docke release also for a PR. It also does that for the dev image. Isn't that a security issue? A bad actor could replace our Dockerfile with `FROM:my_crypto_miner_pwn` and change `scripts/release` to push the image to `latest`, and it would be live in production for our users, right?
> Generally, what about the tokens $(DOCKER_ACCESS_TOKEN), $(LT_RELEASE_UPLOAD_PYPI_PASSWORD), ...? Can't someone just make a PR and include "email me $TOKEN" in the azure-pipelines.yml?

Ah, sorry, I forgot to mention, and this makes testing a bit harder, of course: only builds on `master` have access to the tokens, so a maintainer has to merge such a change to leak a token. That's also the reason we get errors like this:

```
/home/vsts/work/_temp/232ec229-6ffe-41cf-a7d3-5b6dbc7f24ba.sh: line 6: DOCKER_ACCESS_TOKEN: command not found
```

Previously this error was just ignored, because our process for dist builds is tolerant to such errors. Maybe here we should check if the token is set and only then try to push, and also err out the pipeline if the push fails in this case?

azure-pipelines.yml
sk1p: Ah, see comment below about only `master` having access to tokens

uellue: Ah OK! I'll change it back and test properly if we want to run etc

uellue:
> > In the previous push it seemed that it tried to run the docke release also for a PR. It also does that for the dev image. Isn't that a security issue? A bad actor could replace our Dockerfile with `FROM:my_crypto_miner_pwn` and change `scripts/release` to push the image to `latest`, and it would be live in production for our users, right?
> > Generally, what about the tokens $(DOCKER_ACCESS_TOKEN), $(LT_RELEASE_UPLOAD_PYPI_PASSWORD), ...? Can't someone just make a PR and include "email me $TOKEN" in the azure-pipelines.yml?
> 
> Ah, sorry, I forgot to mention, and this makes testing a bit harder, of course: only builds on `master` have access to the tokens, so a maintainer has to merge such a change to leak a token. That's also the reason we get errors like this:
> 
> ```
> /home/vsts/work/_temp/232ec229-6ffe-41cf-a7d3-5b6dbc7f24ba.sh: line 6: DOCKER_ACCESS_TOKEN: command not found
> ```
> 
> Previously this error was just ignored, because our process for dist builds is tolerant to such errors. Maybe here we should check if the token is set and only then try to push, and also err out the pipeline if the push fails in this case?

Ah, that makes sense, I was wondering already! I'll look into it. :-)

sk1p:
> If I understood the flow of events correctly, we are building and pushing the production image only in the second stage now, right? That means it is untested and completely independent of the dev version, if I understood correctly. A bit of a chicken-egg problem since we'd have to tag and push something before the dist tests, but don't know yet if it is sth we want to push, right?

Yeah - we have a similar issue with the rest of the tests already - we don't really have the typical build->test->deploy cycle, because we don't run our test cases against the wheel we build, but we rebuild the wheel instead in the "deploy" stage.

> * Use a special tag, for example "triage" [...]

The "right way"™ would be to tag with the git sha, because we can also have two different builds running in parallel. The current setup for dist tests (tagging as libertem/libertem-dev) is just set up this way for caching - we don't care where the cache base comes from, as long as there is something we can build on.

> Or is there a way to first build the image locally, then derive the test image from it, start up the workers, and only push it after the dist tests passed?

Yeah, this isn't such an easy issue - there are some issues if we want to test against something that is as close to the "production" docker image as possible.

For running the dist tests, we do need an image with many non-trivial changes compared to the production image, like an installation of ipyparallel and all test requirements (so if we base the test docker image, the test requirements could overwrite a package with a newer version than installed in the prod image).

We could make this a bit more faithful by making the dist test setup a bit more complicated: instead of using a "mono-image" for all cases (ipyparallel-{worker,scheduler}, dask-{worker,scheduler}, test environment), we could build separate images on top of the prod image for ipyparallel-* and test environment, and use the prod image unchanged for dask-{scheduler,worker}.

Another idea was to run `tests/server/*` against a `libertem-server` started from the prod image, to make these "true" integration tests - that involves quite a bit of work.

uellue:
Ok, thx for the explanation!

> The "right way"™ would be to tag with the git sha

Hmm, these would accumulate over time, right? I'll try sth two-stage: Build the production image with the dist tests, use it for dask-{scheduler,worker}, and base the libertem-dev image on the local libertem:continuous image.

Only if on master and after passing the tests I'd then push the production image to continuous, hoping that the previous successful test in the PR and the successful dist tests mean that the image is "good enough" for continuous. In reality the CI pipeline practically never fails after merging if it passed for the PR, so this is a reasonable assumption IMO.

In the deploy phase I'd then only re-build, tag and push for other tags than "continuous", using "continuous" as a cache. These would have passed the full CI so that "latest" and named tags should all be good. What do you think?

uellue:
@sk1p 

```python

branch = env['branch']
is_master = branch == "master"
is_stable_branch = branch and STABLE_VERSION_PAT.match(branch) is not None

# ...

if is_stable_branch:
    tags.append("stable")

if is_master:
    tags.append("latest")
```

That means "latest" is also true for "continuous" and RCs? Intuitively I'd expect that "latest" is the latest stable release. How is it usually done on Docker?

A service release for an older release would overwrite "stable"? I mean, currently we only do service releases for the last stable release, it just looks like sth. that may throw us off course later.

Maybe "continuous", "latest" for the newest stable release, and named versions would be enough? Maybe one can list all tags an only update "stable" (resp. "latest", see above) if the tag is actually the newest stable tag?

uellue:
I've just given it a shot. The build works locally on moellenstedt. There's a local test failure in the cache dataset which I don't fully understand. :-)

packaging/docker/docker-compose.yml
```diff
@@ -1,21 +1,26 @@
 version: "3.4"
 
 services:
+  baseimage:
```
sk1p, line 4: What's the reason for having this included here? As it is a service, it should probably have a different name, if we want to keep it here.

packaging/docker/Dockerfile.base
```diff
@@ -0,0 +1,29 @@
+# syntax=docker/dockerfile:1.0.0-experimental
```
sk1p, line 1: As we now have three different `Dockerfile`s, it would be nice to have a small comment at the top of this file, documenting 1) what the build will be tagged as, and 2) a short description of what it does.

Then the `Dockerfile`s should be understandable without having to also read the pipeline definition.

sk1p (commented):
Thanks for taking over the PR! I've added a few comments inline.

When testing this, my first try was to just run `docker-compose run tests` in `packaging/docker/`, which failed spectacularly:

```
base: Pulling from libertem/libertem-dev
[...]
Successfully built 89eaf5766d74
Successfully tagged libertem/libertem:continuous
WARNING: Image for service scheduler was built because it did not already exist. To rebuild this image you must use `docker-compose build` or `docker-compose up --build`.
Starting docker_scheduler_1 ... 
Starting docker_scheduler_1      ... error
WARNING: Host is already in use by another container

ERROR: for docker_scheduler_1  Cannot start service scheduler: driver failed programming external connectivity on endpoint docker_scheduler_1 (96eb7c13e621961ebc1ce4b632092adfdc5Starting docker_ipy-controller_1 ... done

ERROR: for scheduler  Cannot start service scheduler: driver failed programming external connectivity on endpoint docker_scheduler_1 (96eb7c13e621961ebc1ce4b632092adfdc55dce029155022a9a159123e0ecc9b): Error starting userland proxy: listen tcp4 0.0.0.0:8787: bind: address already in use
ERROR: Encountered errors while bringing up the project.
```

I'm currently retrying with `build` first, then `run`, and will report back when that is finished...

packaging/docker/Dockerfile.base
uellue: Only one file now. :-)

scripts/release
```diff
@@ -554,6 +570,34 @@ def do_build_appimage(ctx):
     print("success, files=", appimage_files)
 
 
+@cli.command()
+def docker_tags():
+    """
+    Prints the tags that should be used for any docker images that are built,
+    separated by whitespace.
```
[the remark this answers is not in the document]
uellue: Ok, I've explained it more now. :-)

scripts/release
```diff
@@ -154,6 +163,13 @@ def get_latest_tag():
         return ""
 
 
+def get_latest_stable_tag():
+    tags = get_all_tags()
+    stable_tags = [tag for tag in tags if STABLE_VERSION_PAT.match(tag)]
+    sorted_tags = natsort.natsorted(stable_tags)
```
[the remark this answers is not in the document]
uellue: Ah, thx! That should be more robust if we want to also consider non-stable tags.

uellue:
This should now be in a state where it can be reviewed again. :-)

uellue:
@sk1p can you perhaps check the access token in CI? Somehow it isn't working. I'm not sure if it was working at all before because of the ` || true` to skip silently in case of failure.

uellue:
It seems that "$DOCKER_ACCESS_TOKEN" is actually not "", otherwise it wouldn't try to run. It SHOULD be wrong and it SHOULD not be leaked if I push in this PR, so I'll just debug it's content and force-push it away after so that we don't have the dangerous state in Master.

uellue:
We'd still have to see if the cache works also in CI. I had run the build locally and pushed libertem/libertem:triage, but it rebuilt the  whole thing. Locally, the cache worked.

sk1p:
> can you perhaps check the access token in CI? Somehow it isn't working. I'm not sure if it was working at all before because of the ` || true` to skip silently in case of failure.

It definitely worked before, see for example [the job logs for the latest `master` build here](https://dev.azure.com/LiberTEM/LiberTEM/_build/results?buildId=1416&view=logs&j=084abd41-e52e-5029-96ce-5a53825f2290&t=d1564280-6641-52ce-789b-553df8c76b2f). It was also evident that the image was being [pushed to docker hub](https://hub.docker.com/r/libertem/libertem-dev/tags). The CI job just [won't have access to the token in the PR](https://github.com/LiberTEM/LiberTEM/pull/1144#issuecomment-983897674), so we will have to "test in master" for some of these changes...

sk1p:
> We'd still have to see if the cache works also in CI. I had run the build locally and pushed libertem/libertem:triage, but it rebuilt the whole thing. Locally, the cache worked.

To reproduce this locally, it's a good idea to prune away the local images. For this to work, you first need to `docker container prune` to remove stopped containers, and then something like `docker image prune -a` to remove the images themselves.

uellue:
It looks like I could reproduce and fix the issue locally. Apparently it was using an older version of `python:3.9-slim`, which made the cache ineffective in CI where it likely used a more recent version of `python:3.9-slim`.

sk1p:
> It looks like I could reproduce and fix the issue locally. Apparently it was using an older version of `python:3.9-slim`, which made the cache ineffective in CI where it likely used a more recent version of `python:3.9-slim`.

That's a very good point, and good catch!

packaging/docker/dist-build.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/env bash
```
sk1p, line 1:
```suggestion
#!/usr/bin/env bash
```

packaging/docker/dist-push.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/env bash
```
sk1p, line 1:
```suggestion
#!/usr/bin/env bash
```

packaging/docker/release.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/env bash
```
sk1p, line 1:
```suggestion
#!/usr/bin/env bash
```

sk1p:
> It looks like I could reproduce and fix the issue locally. Apparently it was using an older version of `python:3.9-slim`, which made the cache ineffective in CI where it likely used a more recent version of `python:3.9-slim`.

I just confirmed that this fixed the issue for me for locally building - after pruning the cache was successfully used :tada:

uellue:
In CI it cached `RUN apt-get update && apt-get install`, but not `RUN python3.9 -m venv /venv/`. Probably we should see how it behaves in subsequent PRs when the first version from CI is pushed on Docker hub? Maybe there are subtle differences? 🤷

sk1p:
> In CI it cached `RUN apt-get update && apt-get install`, but not `RUN python3.9 -m venv /venv/`. Probably we should see how it behaves in subsequent PRs when the first version from CI is pushed on Docker hub? Maybe there are subtle differences?

That is weird indeed, as you suggest, it could be differences in different docker versions (maybe?) - as I understand, a `RUN` layer should only run the command exactly once, and it should only be invalidated by previous layers: in this case, it should only be run if the `builder0` layer is different from the previous run...

uellue:
Should be ready for final review now! :-)

scripts/release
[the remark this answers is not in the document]
sk1p: Hm, I don't see it - `docker-tags` now includes `continuous` again, was this the intention? So in `release.sh` it will do a `docker tag libertem/libertem:continuous libertem/libertem:continuous`, correct? I'm fine with this, if it works, I'm just wondering...

src/libertem/io/utils.py
```diff
@@ -6,7 +6,11 @@
     import pwd
 
     def get_owner_name(full_path, stat):
-        return pwd.getpwuid(stat.st_uid).pw_name
+        try:
+            name = pwd.getpwuid(stat.st_uid).pw_name
+        except KeyError:
+            name = stat.st_uid
```
sk1p, line 12: So this is hit if the file owner UID has no name, correct? Surprised we didn't hit this before :)

sk1p (commented):
Thank you for taking over the PR! Looks good overall, I've added a few comments to be addressed, mostly on docs and shell scripting stuff.

I'm now trying to run the dist tests locally, and it hangs at `tests/test_ipython.py` - is there anything special that now needs to be done for running dist tests locally? I just used `dist-build.sh` and ran `docker-compose run tests` afterwards. Maybe docs for this need an update, too? Or am I not supposed to use the `dist-build.sh` locally, and instead use `docker-compose build`, as documented?

src/libertem/io/utils.py
uellue: Yes, precisely! It tends to happen within Docker containers...

scripts/release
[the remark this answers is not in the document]
uellue: Yes, precisely. I wasn't 100 % happy with pushing an image before it went through QA. It broke the logical structure of the pipeline for me. The "push" step is not executed often anyway. So now the "continuous" image is official together with the other deploy. It also keeps that code path in continuous use. :-)

uellue:
> I'm now trying to run the dist tests locally, and it hangs at `tests/test_ipython.py` - is there anything special that now needs to be done for running dist tests locally? I just used `dist-build.sh` and ran `docker-compose run tests` afterwards. Maybe docs for this need an update, too? Or am I not supposed to use the `dist-build.sh` locally, and instead use `docker-compose build`, as documented?

Hmm, it works for me™️ on Moellenstedt. We have to tell users to also "docker -compose down", by the way. Maybe that also helps in your case?

sk1p:
> Hmm, it works for metm on Moellenstedt. We have to tell users to also "docker -compose down", by the way. Maybe that also helps in your case?

Yeah, there were still some old worker containers hanging around, now it makes progress!

packaging/docker/dist-build.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+cd packaging/docker/ || exit
```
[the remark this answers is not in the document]
uellue: Oh, that's neat magic! Changing dirs was annoying.

PR #1145, opened 2021-11-05
Replace NumPy `prod()` with infinite width integer `prod()`

### LINKED ISSUE #846: Make sure size calculations don't overflow
`np.prod()` is currently used all over the LiberTEM code base to calculate the number of entries based on the shape of array-like objects, like datasets, partitions, tiles, slices etc. [`np.prod()` upcasts integers to `np.int`](https://numpy.org/doc/stable/reference/generated/numpy.prod.html), [which is 32 bit on Windows even on 64 bit](https://github.com/numpy/numpy/issues/9464). This is often too small to hold the number of pixels in datasets.  This already caused two overflow bugs (#493 #495 #844) in LiberTEM.

We should make sure that this can't happen so easily for a third time. Aspects could include:

* Add a method or property to relevant objects to calculate the size safely, similar to https://numpy.org/doc/stable/reference/generated/numpy.ndarray.size.html that uses Python arbitrary size integers.
  * Edit: `Shape` already has a `size` property that uses `reduce` and `operator.mul`, which should be safe. However, looking at the code base it is not always clear if a shape is passed as a `Shape` object or as a tuple.
* Make sure we don't use `np.prod()` unsafely in the code base.
* Introduce a dedicated `size_prod()` function that upcasts to int64.
* Monkey patch `np.prod()` to always upcast to `np.int64` instead of `np.int`.
  * How to deal with Numba? Overloading `np.prod()`?

uellue:
Avoids integer overflows in size calculations

Closes #846, see there for more detail

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/common/math.py
+++ b/src/libertem/common/math.py
@@ -1,8 +1,12 @@
 from typing import Iterable, Union
-from typing_extensions import get_args
 
 import numpy as np
 
+
+_prod_accepted = (
+    int, bool,
+    np.bool_, np.signedinteger, np.unsignedinteger
+)
 
 ProdAccepted = Union[
     int, bool,
@@ -21,7 +25,7 @@ def prod(iterable: Iterable[ProdAccepted]):
     result = 1
 
     for item in iterable:
-        if isinstance(item, get_args(ProdAccepted)):
+        if isinstance(item, _prod_accepted):
             result *= int(item)
         else:
             raise ValueError()
--- a/src/libertem/common/math.py
+++ b/src/libertem/common/math.py
@@ -1,0 +1,15 @@
+try:
+    from math import prod
+except ImportError:
+    def prod(iterable):
+        '''
+        Safe product for large size calculations.
+
+        :meth:`numpy.prod` uses 32 bit for default :code:`int` on Windows 64 bit. This
+        function is a replacement for :meth:`math.prod` for versions < Python 3.8
+        that uses infinite width integers.
+        '''
+        result = 1
+        for item in iterable:
+            result *= item
+        return result
--- a/src/libertem/common/math.py
+++ b/src/libertem/common/math.py
@@ -1,15 +1,27 @@
-try:
-    from math import prod
-except ImportError:
-    def prod(iterable):
-        '''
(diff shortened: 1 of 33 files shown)


uellue:
Not sure if or how to document this. It should not have any user-facing effects, but it is relevant for contributors.

src/libertem/common/math.py
```diff
@@ -0,0 +1,15 @@
...
+        '''
+        Safe product for large size calculations.
+
+        :meth:`numpy.prod` uses 32 bit for default :code:`int` on Windows 64 bit. This
+        function is a replacement for :meth:`math.prod` for versions < Python 3.8
+        that uses infinite width integers.
+        '''
+        result = 1
+        for item in iterable:
+            result *= item
```
sk1p, line 14: Should the `item` be cast explicitly to `int` here? Because `iterable` could also contain `numpy` types, which means `np.ndarray.__mul__` will be used, correct? How does `math.prod` work in that case? It would be nice to have a type annotation here, too (right now, that would be `Iterable[int]`, if I'm not mistaken).

Maybe we do need some basic unit tests for this?

sk1p (commented):
In general :+1:. In the lines where you continued to use `np.prod`, should there be a comment describing why it is _not_ using `libertem.common.math.prod`? To make it easier in the future to audit for this.

sk1p:
FYI: `get_args` needs to be imported from `typing_extensions` for compatibility with Py<3.8

sk1p:
Ok, so `get_args` needs `PEP 560` support, which was added in Python 3.7, so we are out of luck there... hm.

uellue:
> Ok, so `get_args` needs `PEP 560` support, which was added in Python 3.7, so we are out of luck there... hm.

Ok, so no detailed typing. I already tried to construct a Union from a tuple, but that doesn't seem to work.

sk1p:
> Ok, so no detailed typing.

How about copy & paste for now, and we fix it to `get_args` once we drop Py 3.6?

sk1p (approved):
LGTM!

Issue #382, opened 2019-06-27
Adaptive partition sizes


sk1p:
For more computationally intensive workloads, it would make sense to show results for smaller units. For this, an adaptive approach could work: start with small partition size and scale it up until we arrive at a nice interactive rate. Could also adapt to load changes on the computation nodes.

For this to work, we need to separate partitions as unit of work and partitions as unit of storage for data locality.

sk1p:
Design: could have an iterator class that is returned by `DataSet.get_partitions` which has as state the current partition size "wish". The tasks need to be submitted in batches, and based on the runtime of a batch, the partition size should be adjusted. This affects the size of partitions that are generated from the iterator afterwards.

Should first implement the interface and general functionality and slowly migrate datasets to actually adhere to these wishes.

sk1p:
I've edited the issue to include a first stab at a spec for these changes - if this sounds sensible, I'll implement a prototype.

/cc @uellue  @matbryan52 anything to be added from your perspective? Any pain points regarding our `dask` integrations?

uellue:
The summary sounds good!

> The performance trend needs to keep in mind differences between GPU and CPU.

That's an interesting one since we can even have different performance per worker on the same device class. Probably it will be enough to track average performance per class, though, since we don't have to be 100 % precise with the update rate, just in the right ballpark.

> How does this interact with live processing?

In my gut feeling it will play together quite nicely. I just don't see it 100 % yet how it should be done. With a distributed data source we'd have to make sure that all source nodes agree on the size of a specific partition. I'd leave this as a next step since adaptive partition size will probably be useful for offline data as well and doesn't necessarily have to be used for every source, so we can figure this out later.

I was also thinking about how to determine the starting partition size. If we start too large, we would "stare at an empty result in the GUI", and if we start too small we reduce max performance through overheads and congestion at the main node. What do you think about starting with a range of partition sizes from a min to a max size? That would have two benefits:

1. We'd make sure that we get at least a few updates quickly and can adapt the size for the following steps
2. The results don't all arrive at the same time, meaning the main node can merge and send out new tasks without causing workers to wait.

For small datasets it would probably be smart to reduce partition size in powers of two so that we can always stack them without overlap, assuming that twice the partition size takes roughly twice the time. Otherwise we might have stragglers.

If that is not practical, for example if we only have a single eligible worker or are doing live processing, we should probably start relatively small, right?

[...]

PR #1149, opened 2021-11-10
Tiling scheme reorganization

sk1p:
Negotiate the tiling scheme once, to reduce overheads for both iteration and live use cases

# TODO
- [x] fix failing test case

## Contributor Checklist:

## Reviewer Checklist:


sk1p:
The failing test case: `tests/io/test_tiling_negotiation.py::test_get_scheme_upper_size_roi`

Because we aren't doing the negotiation per-partition anymore, we can't take the "local" `roi` of the partition into account when building the scheme. So instead of a tile shape of `(3, 144, 144)`, we get something like `(65, 28, 144)` - meaning we have about 5 times as many tiles in this case. 

@uellue do you remember if there is a common case where this happens, or if this was just an edge case / worst case which we handled because we could?

I think the proper fix for this is to just grow the partition in case of a sparse `roi`, in order to always have deep enough tiles.

uellue:
@sk1p The ROI can be really anything. Rebalancing the partitioning so that we have an equal number of frames per partition and the optimal number of partitions would be perfect. :-) For the resistive switching example that should bring significant benefits on large machines because the file is not very large, meaning one partition per core, and the number of frames per partition is probably all over the place with the ROI shape. That means many cores will just be idle and the total time is the time for the largest partitions.

That feels pretty close to #382, right? Would that be within reach?

sk1p:
> That feels pretty close to #382, right? Would that be within reach?

I'll have a look into it - I think it makes sense to tackle this now.

uellue:
> That feels pretty close to #382, right? Would that be within reach?

In more detail, it would mean we calculate the partitioning for each run and have non-uniform partition size, which means we are pretty close to just decide on the fly how large they are. We should calculate the optimal tiling scheme for the largest partition size, since that reduces overheads for fast algorithms, while slower algorithms are probably less affected if the tiles are smaller than expected.

sk1p:
> we are pretty close to just decide on the fly how large they are.

That's what I would aim for, yes.

> We should calculate the optimal tiling scheme for the largest partition size, since that reduces overheads for fast algorithms, while slower algorithms are probably less affected if the tiles are smaller than expected.

Agreed.

So regarding this PR, I would `xfail` the offending test, then it should be ready to be merged.

uellue:
I did some benchmarks and the result is a bit inconclusive for the UDF overheads because of high standard deviation. Since the main point is code re-organization anyway and it is at least not getting slower, so it is OK! :-)

Generally LGTM! :-)

PR #1151, opened 2021-11-12
Implement preloading for server and worker

uellue:
Essential to load HDF5 with custom filters. See https://libertem.github.io/LiberTEM/reference/dataset.html#hdf5

## Contributor Checklist:

## Reviewer Checklist:


uellue:
Adding priority since the feature was in our documentation, but not implemented.

uellue:
@sk1p I'm not so sure if I implemented passing the parameter to the web server correctly. Could you have a look please? :-)

sk1p (changes requested):
>  I'm not so sure if I implemented passing the parameter to the web server correctly. Could you have a look please? :-)

It should work like this! An alternative would be to put `preload` into `SharedState`, like `local_directory`, then we don't need to pass it down individually (this is especially important when we start to add more "stuff" that needs to be passed down...)

uellue:
Thx @sk1p! :-) Should be ready now.

sk1p (approved):
LGTM!

PR #1147, opened 2021-11-08
WIP: Executor based on concurrent.futures

uellue:
Based on #1137 to help with testing the integration, to be rebased
once that is merged!

Closes #1146

TODO:

- [x] Proper benchmarking, verify on `da.from_array(np.memmap(...))` with a very large file
- [ ] General testing strategy for multiple executors, refs also #983
- [ ] Easy selection of executor in `Context()

## Contributor Checklist:

* [ ] I have added or updated my entry in [the creators.json file](https://github.com/LiberTEM/LiberTEM/blob/master/packaging/creators.json)
* [ ] I have added [a changelog entry](https://github.com/LiberTEM/LiberTEM/tree/master/docs/source/changelog) for my contribution
* [ ] I have added/updated documentation for all user-facing changes
* [ ] I have added/updated test cases
* [ ] I have included the [rebuilt production build of the client](https://libertem.github.io/LiberTEM/contributing.html?#building-the-client) (only if changes were made to the GUI)

## Reviewer Checklist:

* [ ] `/azp run libertem.libertem-data` passed

Issue #473, opened 2019-11-08
Meta issue: Advanced executor


uellue:
This Issue collects other Issues that depend on extended capabilities in the executor, i.e. where just running Dask futures in the cluster in the current way may not be enough.

<strike>- #364 GPU processing could require dedicated worker processes for CPU and GPU processing, and balancing between the two.</strike>
- #335 Shared memory requires one function to set it up once, and then others to attach to it while a job runs, and then a reliable mechanism to clean it up. Avoid race conditions, dead locks and orphans. See #1006 #1005 for prototyping.
- #186 Parallel execution of CPU, mass storage and network bound tasks, plus small service tasks like file browsing. That means the OS on the processing node should drive the decision on which task is executed next without oversubscription or idling of the CPU. Frequent context switching is quite expensive since we lose the CPU cache! See also https://unix.stackexchange.com/questions/466722/how-to-change-the-length-of-time-slices-used-by-the-linux-cpu-scheduler
- #242 Per-node intermediate reduction.
- Live data processing, similar to #186. See k2is branch in https://github.com/LiberTEM/LiberTEM-live/ for a prototype.
- Support for better decoupling of I/O and computation, for example via liburing #268 
- #199 API in different languages and Python versions
- #626 Improve interface and implementation for AUX data. Live AUX data for live processing could be neat. Specific example: Adjust SSB reconstruction parameters following the actual microscope settings.

uellue:
Support for Dask-based execution should probably remain as an option even after we have an advanced executor, since Dask allows to run on ipyparallel clusters and enables integration with Hyperspy et al, see #922.

uellue:
@sk1p @matbryan52 This could be the right issue to coordinate the executor and UDFRunner redesign.

matbryan52:
Wondering if either there has been progress/planning on this or, if not, should we make a plan?

I've been examining the architecture and think I have a reasonable idea of how it all fits together - I've been adding to sort of (simplified!) diagram here: https://gist.github.com/matbryan52/21d32163af26fca2ffa521fdfa6c836f

uellue:
@matbryan52 Thank you for the diagram! It looks a bit intimidating. :-D It will be a great tool to understand the flow, we can work on refining that and then restructure it for the new requriements.

[...]

Issue #427, opened 2019-09-13
Document API stability


sk1p:
Currently, our reference documentation includes some private APIs, for example, many methods of `DataSet` may need to change in the future. On the other hand, `Slice` and `Shape` are in the "Internal API" section, and are used as part of `UDFMeta`, so are quasi-public. We need to make this more clear!

sk1p:
See also: https://github.com/Carreau/frappuccino to enforce API stability, used for example by scikit-image

PR #1158, opened 2021-11-23
Delayed and concurrent Job Executor, take 2

### LINKED ISSUE #224: Include preloading in documentation
Include info from #218 and #223 in documentation as soon as we describe using an external dask cluster.

uellue:
This PR supersedes #983 since the recent changes in the LiberTEM internals made a complete re-implementation easier than a rebase.

Closes #224 

- [x] Also include #1147 to use the same testing strategy.

* Port to recent changes in UDFRunner and Executor
* Same code to run UDFs directly or in a delayed fashion, only difference
  is the executor.
* Skip calculating UDF results if they are not required
* Testing strategy for multiple executors

TODO and TBD items:
* No happy path (yet), the result is a plain dask.delayed value
* In particular, no chunking of the result
* Documentation, examples etc

#983 allowed to calculate results for UDFs that use default merge as a chunked Dask array by bypassing the merge function and concatenating the result buffers instead. However, that became more difficult with the recent changes in the UDF interface such as running multiple UDFs in one pass and allowing `get_results()` for UDFs.

If one wanted to get results as Dask arrays, one should think more how that could be possible within the current boundaries of the UDF interface, or consider an extension. LiberTEM is inherently built around assigning into buffers procedurally, while Dask uses a more functional style. Also, the recently added much improved `map()` implementation in Hyperspy means that we can focus on other use cases.

## Contributor Checklist:

## Reviewer Checklist:


uellue:
Marked as WIP since I had an idea how to make proper Dask arrays for result buffers. @matbryan52 perhaps we can discuss once I have the alternative approach prototyped and see if there are still better options?

src/libertem/executor/dask.py
```diff
@@ -429,30 +429,49 @@ def close(self):
...
     @classmethod
     def make_local(cls, spec=None, cluster_kwargs=None, client_kwargs=None,
             preload: Optional[Tuple[str]] = None):
         """
         Spin up a local dask cluster
 
-        interesting cluster_kwargs:
-            threads_per_worker
-            n_workers
+        See http://distributed.dask.org/en/stable/api.html#distributed.SpecCluster
```
sk1p, line 470: Hmm, so I know that this comment about `n_workers` and `threads_per_worker` was outdated - but maybe we can re-add something that actually works for easily creating an executor with a custom number of CPU workers.

sk1p (commented):
I know this is still WIP, but I hat a quick look at the changes, mostly in `Context`

src/libertem/executor/integration.py
```diff
@@ -0,0 +1,38 @@
+import types
+
+import dask
+import dask.delayed
```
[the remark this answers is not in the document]
uellue: Good point, good to check! It doesn't start anything since it will just use whatever Dask scheduler is configured when `compute()` is called. The overhead of importing it is zero:

image

It appears that Dask loads it somewhen, somehow anyway.

src/libertem/api.py
```diff
@@ -73,14 +102,8 @@ class Context:
...
 
-    def __init__(self, executor: JobExecutor = None, plot_class=None):
-        if executor is None:
-            executor = self._create_local_executor()
-        if not isinstance(executor, JobExecutor):
-            raise ValueError(
-                f'Argument `executor` is not an instance of {JobExecutor}, '
-                f'got type "{type(executor)}" instead.'
-            )
+    def __init__(self, executor: Optional[Union[JobExecutor, str]] = None, plot_class=None):
```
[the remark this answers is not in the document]
uellue: Ok, will do! Good point, it keeps getting more and more convoluted...

src/libertem/executor/dask.py
uellue: Ok, I'll add documentation! It is actually very easy using `cluster_spec()` and `detect()`.

matbryan52:
> Marked as WIP since I had an idea how to make proper Dask arrays for result buffers. @matbryan52 perhaps we can discuss once I have the alternative approach prototyped and see if there are still better options?

I put together a prototype last week which created Dask result buffers, and which the user can assign into in the usual way (`self.results.intensity[:] += val`), but it's quite brittle! I had it working for a simple nav and sig buffer cases but didn't think about things like ROIs and corrections (though it might 'just work'). We also could only have result buffers chunked in the flat nav dimension as that is a constraint of the Partition interface.  If the nav dimension is then reshaped to 2D it would (effectively) require computing the whole array unless it happens to be that the partitions span nav_shape[1].

docs/source/reference/executors.rst
```diff
@@ -0,0 +1,41 @@
...
+LiberTEM executors
+==================
+
+See :ref:`executors` for a general overview.
+
+.. versionadded:: 0.9.0
+    Previously, the executor API was internal. Since influence on the executor
+    is important for integration with Dask and other frameworks,
+    the API is now documented to help with that. Nevertheless, this is still
+    experimental and may change between releases without notice.
```
sk1p, line 12: Can we move this documentation to the development section, and clearly mark it as an internal API? I think it would be a mistake to make the `JobExecutor` API public (or even quasi-public), especially in its current state. It may be important for integration with dask etc. - but what is important for the _user_ is not the internal executor API.

"still experimental" means to me that it is already usable but subject to change, and on the track to be stabilized. If I read that in the API docs of a library, I might start writing code (outside of that library!) against that API, and expect that to keep the same in principle, just with some minor changes. I don't think this can be expected from the `JobExecutor` API currently.

Making changes like #1143 in an efficient manner very much depends on keeping the API private, so we can change all implementations and users in lock-step. Generally, if we make an API available to users as experimental which is not on track to stabilize, it will create problems both on our side, and on the user side. We can't really make sweeping changes, because once there is user code out there, we rightfully have some commitment to keep it working or to make a reasonable migration possible, and on the user side, it will be very inconvenient if every new release means changing code, even if a migration is possible.

The user-facing parts are another story: the executors themselves should be documented, and which should be chosen in which situation, how they can be constructed (`DaskJobExecutor.{connect,make_local}` etc.).

If we want to think about making the executor API public (even experimental), we should wait until we are progressed further with #473, in my opinion - I don't think much of the current API will survive in its current form.

I hope this does not read as too argumentative - and to take a step back: 

- What is the user case for calling any of the methods on the `JobExecutor` interface, except for the constructor(s)?
- What is the use case for subclassing either `JobExectuor` or one of out implementations?
- Generally, what is the actual use case for making this API public?

TL;DR: #427

uellue: Good points! I've pushed some changes. Does that sound better? Argumentative is good to get the best result together. :-)

uellue:
> I'll try to get the proto finished then we can talk about it?

Sounds good! 👍

uellue:
Superseded by #1170 and #1168

PR #1159, opened 2021-11-25
WIP: Adaptive and dynamic partition sizing

### LINKED ISSUE #382: Adaptive partition sizes

### LINKED ISSUE #933: make_dask_array: roi parameter broken
Example:

```python
sparse_roi = np.random.choice([True, False], size=ds.shape.nav, p=[0.1, 0.9])
a2, _ = make_dask_array(ds, roi=sparse_roi)
ctx.executor.client.compute(a2).result()
```

Results in:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-66-a0875084e88a> in <module>
----> 1 ctx.executor.client.compute(a2).result()

~/.virtualenvs/libertem/lib/python3.8/site-packages/distributed/client.py in result(self, timeout)
    222         if self.status == "error":
    223             typ, exc, tb = result
--> 224             raise exc.with_traceback(tb)
    225         elif self.status == "cancelled":
    226             raise result

~/.virtualenvs/libertem/lib/python3.8/site-packages/dask/utils.py in __call__()
    892 
    893     def __call__(self, obj, *args, **kwargs):
--> 894         return getattr(obj, self.method)(*args, **kwargs)
    895 
    896     def __reduce__(self):

~/source/LiberTEM/src/libertem/io/dataset/base/tiling.py in reshape()
    390         # NOTE: "shedding" our DataTile class on reshape, as we can't properly update
    391         # the slice to keep it aligned with the reshape process.
--> 392         return np.asarray(self).view(np.ndarray).reshape(*args, **kwargs)
    393 
    394     @property

ValueError: cannot reshape array of size 12451840 into shape (16,128,256,256)

> /home/alex/source/LiberTEM/src/libertem/io/dataset/base/tiling.py(392)reshape()
    390         # NOTE: "shedding" our DataTile class on reshape, as we can't properly update
    391         # the slice to keep it aligned with the reshape process.
--> 392         return np.asarray(self).view(np.ndarray).reshape(*args, **kwargs)
    393 
    394     @property
```

This is partially caused by the reshape here:

https://github.com/LiberTEM/LiberTEM/blob/317c17ae516e6720051524884217578d69be94ba/src/libertem/contrib/daskadapter.py#L24

If we only reshape if we don't have a `roi`, this should be consistent with the behavior of numpy.

I think something else that changes the results is the handling of partitions that don't yield any tiles in `get_macrotile`. The simplest scenario is just picking a single frame via `roi`. Only one partition will yield a tile (which should only contain the one frame), for all others, `get_macrotile` will substitute an all-zero block of partition size:

https://github.com/LiberTEM/LiberTEM/blob/317c17ae516e6720051524884217578d69be94ba/src/libertem/io/dataset/base/partition.py#L178-L193

sk1p:
Fixes #382, fixes #933 

The general structure is done, but may need updates after #1137 lands, as it assumes uniform chunking currently.

## Still TODO

- [ ] The actual calculation of new partition sizes (we already get feedback on task timings, but that is not yet used)
- [ ] Take data locality into account. In the first iteration, this could mean that partitions provide slice -> location mappings, which means we can generate partitions that don't cross borders of these slices. This also means 
- [ ] In `make_dask_array`, get the partition sizing right (by having some API that gives us a default size maybe?)
  - [ ] Use same capability in tiling scheme negotiation as a default partition size
- [ ] Constraints for non-uniform chunking
- [ ] Update documentation on I/O for new API

## Contributor Checklist:

* [] I have included the [rebuilt production build of the client](https://libertem.github.io/LiberTEM/contributing.html?#building-the-client) (only if changes were made to the GUI)

## Reviewer Checklist:


sk1p:
Discussion: some parts (dictating slices for partitions from the outside) would be very useful for #1264, so I'll extract these into a separate PR

sk1p:
Closing for now, as it's become stale.

PR #1165, opened 2021-12-07
ApplyMasksUDF: re-use Slice objects from TilingScheme

sk1p:
Micro-optimization in the `ApplyMasksUDF`: directly re-use `Slice` objects from the `TilingScheme` instead of re-creating them using `discard_nav`. I noticed this when profiling something for #1159.

When using our usual test data, it is a bit hard to measure; even with the inline executor and multiple rounds, the numbers are not stable; I could cherry-pick a run that is faster by 500ms, or one that is exactly as fast as before - we still have some work to do here...

When running a test script on a large (~200GiB) test file, it can be measured more easily:

Before (sum/mean/min/max):

```
total: [75.06046856753528, 15.012093713507056, 14.906086668372154, 15.101721707731485]
```

After (sum/mean/min/max):

```
total: [73.61619642749429, 14.723239285498858, 14.58340313937515, 14.796591619960964]
```

And the small benchmark script I used:

<details>

```python
import time
import logging

import numpy as np

from libertem.udf.masks import ApplyMasksUDF


def main():
    logging.basicConfig(level=logging.INFO)
    ctx = Context()
    # ds = ctx.load("mib", path="/home/alex/source/upstream-libertem/benchmarks/io/data/some_prefix/MIB/20200518 165148/default.hdr")
    ds = ctx.load("k2is", path="/cachedata/alex/03_25V_.gtg")

    sig_shape = ds.shape.sig

    def mask():
        return np.ones(sig_shape, dtype=bool)

    udf = ApplyMasksUDF(mask_factories=[mask], backends=('numpy',))

    # warmup
    ctx.run_udf(udf=udf, dataset=ds)

    deltas = []

    for i in range(5):
        t0 = time.perf_counter()
        ctx.run_udf(udf=udf, dataset=ds)
        t1 = time.perf_counter()
        delta = t1 - t0
        print(f"run {i}: {delta}")
        deltas.append(delta)

    stats = [
        np.sum(deltas),
        np.mean(deltas),
        np.min(deltas),
        np.max(deltas),
    ]

    print(f"total: {stats}")


if __name__ == "__main__":
    main()
```
</details>

This also includes fixes for the memory dataset, which didn't properly communicate its tile shape, when one was forced by the user. Now, the `TilingScheme` should properly represent the slices of the tiles returned by `get_tiles`.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/benchmarks/io/test_k2is.py
+++ b/benchmarks/io/test_k2is.py
@@ -92,6 +92,37 @@ class TestUseSharedExecutor:
             iterations=1,
         )
 
+    @pytest.mark.parametrize(
+        "prefix", PREFIXES
+    )
+    @pytest.mark.benchmark(
+        group="io"
+    )
+    def test_mask_repeated(self, benchmark, prefix, lt_ctx):
+        hdr = os.path.join(prefix, K2IS_FILE)
+        flist = filelist(hdr)
+
+        ctx = lt_ctx
+        ds = ctx.load(filetype="k2is", path=hdr)
+
+        sig_shape = ds.shape.sig
+
+        def mask():
+            return np.ones(sig_shape, dtype=bool)
+
+        udf = ApplyMasksUDF(mask_factories=[mask], backends=('numpy', ))
+
+        # warmup:
+        ctx.run_udf(udf=udf, dataset=ds)
+        warmup_cache(flist)
+
+        benchmark.pedantic(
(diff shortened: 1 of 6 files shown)


uellue (approved):
Looks good to me! :-)

sk1p:
Data tests failure was probably because of high load on the "CI machine", re-running failed jobs now...

PR #1167, opened 2021-12-09
Add `UDFMeta.sig_slice` and `UDFMeta.tiling_scheme_idx`

### LINKED ISSUE #1166: cupy v10 support?
Of special note in the [upgrade guide](https://docs.cupy.dev/en/latest/upgrade.html) is a [change in `cupy.cuda.Device` behavior](https://docs.cupy.dev/en/latest/upgrade.html#change-in-cupy-cuda-device-behavior) - does this affect our cupy use?

sk1p:
They are meant as a direct replacement of accessing the `DataTile` attributes directly, and serve as an official interface for this functionality, which also works with cupy.

In addition, removes a now-redundant `UDFMeta` recreation in the `UDFRunner`, which was needed before to set the tiling scheme.

Fixes #1166

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/udf/base.py
+++ b/src/libertem/udf/base.py
@@ -73,6 +73,9 @@ class UDFMeta:
 
     .. versionchanged:: 0.6.0
         Information on compute backend, corrections, coordinates and tiling scheme added
+
+    .. versionchanged:: 0.9.0
+        :code:`tiling_scheme_idx` and `sig_slice` added
     """
 
     def __init__(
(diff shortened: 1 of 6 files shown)


sk1p:
Still working on a few fixes/tests for `process_frame`/`process_partition`, so not entirely done yet

src/libertem/udf/base.py
```diff
@@ -145,6 +148,19 @@ def tiling_scheme(self) -> Optional[TilingScheme]:
...
+        return self._tiling_index
+
+    @tiling_scheme_idx.setter
+    def tiling_scheme_idx(self, new_idx: int) -> None:
+        self._tiling_index = new_idx
+
+    @property
+    def sig_slice(self) -> Slice:
+        assert self._tiling_scheme is not None
+        return self._tiling_scheme[self._tiling_index]
```
uellue, line 162: Ok, so the point is that the sig slice is reusable for tiles with the same tiling index, while the tile slice changes with each tile, right?

uellue (approved):
Looks good, fix confirmed! 🚀

src/libertem/udf/base.py
sk1p: Indeed - churning through additional objects with `tile_slice.discard_nav()` or similar does create a measurable overhead for throughput-limited workloads like `ApplyMasksUDF`

PR #1168, opened 2021-12-10
Release GIL in more Numba methods

matbryan52:
This commit was extracted from #1158 as it is independent but touches a number of files. The original commit message is:

It turns out that tiling slows down the threaded case. This
change releases the GIL for more parts of the LiberTEM internals.  The
float32 case uses a very large single tile for the partition and shows
the difference.

Rebased from @uellue commit d0037924

## Contributor Checklist:


sk1p:
Thanks for extracting the commit, really makes sense as separate PR. LGTM!

PR #1172, opened 2021-12-17
Integrate CoM parameter guess into web GUI

### LINKED ISSUE #1112: Integrate CoM parameter guess in GUI
Follow-up to #1111: Integrate guess to GUI, for example with an "auto" button.

It turns out that finding the correct alignment between scan and detector is critical for correct results, in particular with ptychography, and at the same time error-prone.

With this function one can quickly check if the parameters distributed with a dataset make sense and/or find a good guess if the required metadata is not available.

sk1p:
Fixes #1112 - adds a "Guess parameters" button to the CoM analysis.

Also includes a bunch of new type annotations. @uellue: most important part to review is the RPC itself, in `libertem.analysis.com`, and maybe trying this on a few data sets.

## TODO

- [x] Sometimes, the guess still diverges - need to find out if this is in the new code, or in the guess itself
- [x] Add a few integration tests for the new API endpoint
- [x] Run analysis if we don't have a result yet
- [x] Add feedback: spinner while the API is still guessing

## Contributor Checklist:

## Reviewer Checklist:


uellue (approved):
This is really nice, I'll use that a lot! In particular automatic centering is NICE, that was super annoying before. Fixed the issue with diverging by applying the shift in the correct reference frame. :-)

uellue:
Agreed with the TODOs, these were indeed a bit confusing when using it.

sk1p:
> Fixed the issue with diverging by applying the shift in the correct reference frame. :-)

Thanks for that - I guess I didn't properly read the docs for the parameter guess functions :)

This should be ready to be merged once CI passes.

uellue:
Thank you, merging! :-)

PR #1179, opened 2022-01-12
Experimental: support for the TVIPS binary format

sk1p:
## TODO

- [x] Add format support to the README (and other documentation?)
- [x] Add GUI support
- [x] Support for guessing the `nav_shape` from the image headers? It's a bit unclear whether this can be done reliably (how do we know a specific header is present for an image? What is the value if it is not present?)

## Contributor Checklist:

## Reviewer Checklist:


jhgee:
@sk1p Sure, I will test it!

sk1p:
I have now also added GUI support (for testing, please run `npm install && npm run start` in the client folder while having a LiberTEM server running on port 9000). What I was not able to test locally is reading file sets of multiple `.tvips` files, but I have implemented support for this according to the documentation.

jhgee:
It works. 🥳 
I tested it with a 129 GB large data set (64 x 64 x 4096 x 4096 = 2^36 px!) and it works super nicely! the dataset consists of several 32GB large .tvips files.
I could easily make use of the sync_offset feature from #793 which is super convenient and works like a charm.
Thanks again!
image

uellue:
Thx for the feedback! I'd merge this after #1170 to not risk another conflict. Rebasing this will likely be easier than the 190 commit behemoth...

sk1p:
> It works. I tested it with a 129 GB large data set (64 x 64 x 4096 x 4096 = 2^36 px!) and it works super nicely! the dataset consists of several 32GB large .tvips files. I could easily make use of the sync_offset feature from #793 which is super convenient and works like a charm. Thanks again!

Thank you for giving it a go! For properly guessing the `sync_offset` parameter, I'm still in contact with Tietz to clearly understand the metadata in the image headers. If you do have a smaller 4D STEM data set available, which also has a non-zero `sync_offset` and that you could share with us, that would also be helpful. It would also be possible to merge this PR as-is, and improve TVIPS support afterwards, probably in a v0.9.1 release or similar.

matbryan52:
Just wanted to add that I've been able to run tests on my machine with the standard test file. I've been through the DS code and nothing jumps out at me as unusual, though I don't know much at all about the format definition! If there's something specific you'd like tested, let me know.

sk1p:
> I just uploaded another data set 2GB in size and shared via email with @uellue and @sk1p ;-)

Thank you! cryptshare is also requesting a password - I guess you need to mail that separately from the sharing itself?

sk1p:
I have now added support for detecting both `sync_offset` and `nav_shape` automatically from the image headers. I'm not sure how reliable this is, but for the test data from @jhgee it successfully detects both. If it can't detect those automatically, the old default values are used instead (`sync_offset=0` and `nav_shape=(number_of_frames,)`), so it should not regress in that case. If we get reports (and sample data) of this not working properly we can fine-tune the guessing. There is also a hard-coded upper bound until which we search for the beginning of the scan, which might need to be changed in case of a gross mis-synchronization. It will also fail if the whole first file in the set doesn't contain the first line of the scan.

> Just wanted to add that I've been able to run tests on my machine with the standard test file. I've been through the DS code and nothing jumps out at me as unusual, though I don't know much at all about the format definition! If there's something specific you'd like tested, let me know.

Thanks for giving this a go and having a look at the code! Nothing specific to look at, I think.

src/libertem/io/dataset/tvips.py
```diff
@@ -75,6 +79,37 @@ def convert_to_python(self, raw_data):
 ]
 
 
+image_header_v2_dtype = [
+    ('ICounter', 'u4'),                 # image counter, continues through all files
+    ('ITime', 'u4'),                    # unix time stamp
+    ('IMS', 'u4'),                      # timestamp milliseconds
+    ('LUT_Index', 'u4'),                # LUT index (?)
+    ('Faraday', 'float32'),             # faraday cup value (unit?)
+    ('TEM_Mag', 'u4'),                  # magnification (unit?)
```
jhgee, line 88: the pixel size of acquired image can most probably be calculated from real pixel size of the camera: `px_size = px_size_real * binning / TEM_Mag`
`px_size_real` is 15.5 um for [XF416(R)](https://www.tvips.com/camera-systems/temcam-xf-series/)

jhgee:
I just checked the new version with a few acquisitions and it works!

sk1p:
> I just checked the new version with a few acquisitions and the parameter detection works!

Thanks for the feedback! If nothing else comes up, I'll rebuild the client js stuff, then the PR should be ready to be merged.

PR #1192, opened 2022-01-26
MIB: fix and document `disable_glob` parameter

sk1p:
## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/io/dataset/mib.py
+++ b/src/libertem/io/dataset/mib.py
@@ -125,8 +125,11 @@ def _get_sequence(f: "MIBHeaderReader"):
     return f.fields['sequence_first_image']
 
 
-def get_image_count_and_sig_shape(path: str) -> Tuple[int, Tuple[int, int]]:
-    fns = get_filenames(path)
+def get_image_count_and_sig_shape(
+    path: str,
+    disable_glob: bool = False,
+) -> Tuple[int, Tuple[int, int]]:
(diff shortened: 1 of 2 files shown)


sk1p:
Thinking about this a bit, we could also implement a heuristic that automatically disables globbing (based on the sequence numbers in the .mib files), but that could be a bit surprising behavior. On the other hand, it would also work from the GUI, where it currently fails with an obscure error...

uellue:
Probably better to add a GUI element that allows to enable or disable globbing? The heuristic could be used to set the starting value. As you write, the behavior could be really surprising depending on how people name and organize their files. With the GUI element it would be transparent that there are two modes available, and people can influence the behavior.

sk1p:
> Probably better to add a GUI element that allows to enable or disable globbing? The heuristic could be used to set the starting value. As you write, the behavior could be really surprising depending on how people name and organize their files. With the GUI element it would be transparent that there are two modes available, and people can influence the behavior.

I think this would go hand-in-hand with general support for specifying stacks of files in the GUI, for cases where no default naming scheme exists (MIB, TVIPS, FRMS6, K2IS, maybe others all have a usable default naming scheme, but if you have a stack of raw files, or a stack of DM3, things will be different...) - basically #498

uellue:
Discussion with @sk1p: We defer a "proper" solution for #498 to later and merge this now as-is. :-) Thx for taking care of it!

PR #1193, opened 2022-01-26
Documentation pass release 0.9.0

matbryan52:
Pass through the documentation for release 0.9.0, with changes mainly in docs for the executors, and the new Dask-based features.

Also adds a `docs` badge to the Readme.

There are a few small code changes, to be discussed:

- added a check and test that `merge_all` returns a dictionary with all the correct keys
- added option `Context.make_with('delayed')`

### TODO:
- [x] Clean up the Hyperspy integration notebook a bit

## Contributor Checklist:


matbryan52:
The initial test failure was because (in my mind) I expected that we require `merge_all` to return results for _all_ declared result buffers. I've modified the check so that it only requires a _subset_ of the declared buffers, but perhaps this should be discussed ?

matbryan52:
HS Notebook is now updated, though I have to run NBVAL tests (I think?). I'll have to figure that out.

docs/source/api.rst
```diff
@@ -64,14 +85,19 @@ See also :ref:`threading` for more information on multithreading in UDFs.
...
+main thread and workers efficiently, but is severely slowed down by the
+Python `global interpreter lock <https://wiki.python.org/moin/GlobalInterpreterLock>`_
+under many circumstances. Furthermore, it can create thread safety issues such as
 https://github.com/LiberTEM/LiberTEM-blobfinder/issues/35.
 
+It is also in principle possible to use a :class:`python:concurrent.futures.ProcessPoolExecutor`
+as backing for the :class:`~libertem.executor.concurrent.ConcurrentJobExecutor`,
+though this is untested and is likely to lead to worse performance than the
+LiberTEM default :class:`~libertem.executor.dask.DaskJobExecutor`.
+
```
uellue, line 100: I tried that, actually, and it didn't work. I don't remember how exactly it failed, though. Maybe one could add an xfail test case just for completeness sake, since we both had the same idea independently? :-)

uellue (commented):
Nice, many thanks for the updates! A big improvement. :+1: 

If you have time and energy, perhaps you could also take a look at the rest of the documentation, not only the new parts? Alex and I have just read it too often already to see what can still be improved. Doesn't have to be super detailed, can also just be opening Issues for later if the fix would take longer!

uellue:
> HS Notebook is now updated, though I have to run NBVAL tests (I think?). I'll have to figure that out.

It is tested in the Data CI pipeline. You should have rights to trigger it. Just write `/azp run libertem.libertem-data` as a comment to this PR.

docs/source/api.rst
matbryan52: It did run for me, but only when I wasn't pickling a local function / closure, specifically.

uellue: Ah right, that was it!

matbryan52:
> If you have time and energy, perhaps you could also take a look at the rest of the documentation, not only the new parts? Alex and I have just read it too often already to see what can still be improved. Doesn't have to be super detailed, can also just be opening Issues for later if the fix would take longer!

I can take a look quick, but really I think there is scope for some more substantial re-thinking of parts of the docs, particularly a 'beginner's guide'. Right now all the information is there but it's not always simple enough or easy to find.

sk1p:
> I can take a look quick, but really I think there is scope for some more substantial re-thinking of parts of the docs, particularly a 'beginner's guide'. Right now all the information is there but it's not always simple enough or easy to find.

Agreed. In the past, we have often done a "doc sprint" after the release, maybe it's time for another iteration after v0.9.0?

docs/source/udf/advanced.rst
```diff
@@ -579,20 +579,20 @@ Threading
 ---------
 
 By default, LiberTEM uses multiprocessing with one process per CPU core for
-offline processing. In that scenario, UDFs should only use a single thread to
-avoid oversubscription.
+offline processing :class:`~libertem.executor.dask.DaskJobExecutor`. In that
+scenario, UDFs should only use a single thread to avoid oversubscription.
```
sk1p, line 583: I think the sentence is missing something, like "[...] for offline processing _using_ `DaskJobExecutor`" or similar?

sk1p (commented):
Generally looks good to me, comments/suggestions on some details inside.

docs/source/udf/advanced.rst
matbryan52: I've made that change, thanks!

src/libertem/io/dataset/dask.py
```diff
@@ -67,7 +67,11 @@ class DaskBackendImpl(MMapBackendImpl):
 
 
 class DaskDataSet(DataSet):
+    _default_min_size = 128 * (2**20)  # MB
+
```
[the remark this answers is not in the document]
matbryan52: I've move the value to a private @property on the class with its own docstring. The reason for making the change was to let a user change the default once-and-for-all rather than embedding it in __init__.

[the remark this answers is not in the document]
matbryan52: Thanks for finding this, though, I didn't know about the interaction with Sphinx/Docstring finding!

uellue:
Hi @matbryan52, is this ready for merging? Just asking since it is still a draft PR. :-)

matbryan52:
> Hi @matbryan52, is this ready for merging? Just asking since it is still a draft PR. :-)

For me yes, assuming the tests pass appropriately ! I didn't have a chance to do a full doc pass but I agree with @sk1p that dedicating a few days to docs later would be a good idea.

Issue #1196, opened 2022-01-28
Default nav shape for our reference TVIPS dataset; opening with Nav shape (0, x) or (y, 0) in the GUI.


uellue:
## How to reproduce

* Open `/LargeData/ER-C-1/groups/data_science/data/reference/TVIPS/rec_20200623_080237_000.tvips` for the first time or reset parameters.
* Click "Load dataset" with the detected parameters.

I've tested it with Singularity on Moellenstedt, just FYI.

## What happens

LiberTEM detects the nav shape as `(10, 0)`. Clicking "Load dataset" with this nav shape parameter doesn't trigger any action. The dialogue just remains as-is. After changing to (10, 1), the file can be loaded.

I've tested a nav shape of (0, x) or (y, 0) with MIB as well and get the same behavior.

Not getting any reaction confused me as a user.

## What should happen

First, the autodetected shape should not contain a zero. If the nav shape can't be detected properly for 4D STEM, it should default to (y, 1).

Second, attempting to load with an invalid shape should trigger an error or give some other feedback to the user, instead of just not doing anything.

Since that actually confused me even as an experienced user, users will likely encounter this at least for TVIPS data, and it should be straightforward to fix, I'd propose to include it in 0.9.

uellue:
After a bit of digging, it seems that the TVIPS dataset detects the correct shape `(10, )`, but the web GUI doesn't handle the missing parameter well. @sk1p could you perhaps have a look? :-)

AnandBaburajan:
@sk1p I can take this up in case you haven't started working on this already.

sk1p:
> @sk1p I can take this up in case you haven't started working on this already.

Thanks! I haven't started on it yet, so feel free to take over. Let me know if you need assistance 😊

sk1p:
@AnandBaburajan FYI our CI is having some issues right now, but feel free to create a PR anyways, I'll manually build and test if needed. Thanks!

PR #1202, opened 2022-02-02
Handle shapes properly in GUI

AnandBaburajan:
Fixes the `the autodetected shape should not contain a zero. If the nav shape can't be detected properly for 4D STEM, it should default to (y, 1)` part of #1196. I looked into the `attempting to load with an invalid shape should trigger an error or give some other feedback to the user, instead of just not doing anything` part but haven't figured out yet why the Formik error messages aren't showing up now. Maybe that part can be fixed in a separate PR?

I also refactored some code, renamed some functions and fixed [this](https://github.com/AnandBaburajan/LiberTEM/blob/master/client/src/dataset/components/TupleInput.tsx#L41) bug.

: )

Edit: also updates formik to 2.x

## Contributor Checklist:

## Reviewer Checklist:


sk1p:
Hi Anand,

First of all, thank you for taking care of this!

> Fixes the `the autodetected shape should not contain a zero. If the nav shape can't be detected properly for 4D STEM, it should default to (y, 1)` part of #1196.

Great! In my testing, the fix worked well.

> I looked into the `attempting to load with an invalid shape should trigger an error or give some other feedback to the user, instead of just not doing anything` part but haven't figured out yet why the Formik error messages aren't showing up now. Maybe that part can be fixed in a separate PR?

It's possible that upgrading our JSON schema validation library (ajv) has caused this issue, or at least parts of it. Validation errors from the schema aren't properly "connected" to the right fields, and the merging of custom errors into the others is actually also broken. But even then, the errors aren't showing up for some reason. I'll have a deeper look into this.

> I also refactored some code, renamed some functions and fixed [this](https://github.com/AnandBaburajan/LiberTEM/blob/master/client/src/dataset/components/TupleInput.tsx#L41) bug.

:+1:

sk1p:
> Validation errors from the schema aren't properly "connected" to the right fields, and the merging of custom errors into the others is actually also broken. But even then, the errors aren't showing up for some reason. I'll have a deeper look into this.

Small update: I tracked this down to `formik.touched.{nav,sig}_shape` not being set for some reason. I'll continue debugging later. Updating formik to the latest version didn't seem to have an effect, either.

AnandBaburajan:
> First of all, thank you for taking care of this!

Hello Alex! Thank you and Dieter for everything you've done for me!

> Small update: I tracked this down to `formik.touched.{nav,sig}_shape` not being set for some reason. I'll continue debugging later. Updating formik to the latest version didn't seem to have an effect, either.

Ah. Here's what I tried doing so far. I passed `touched` in [here](https://github.com/LiberTEM/LiberTEM/blob/master/client/src/dataset/components/MIBParamsForm.tsx#L34), logged it and saw that all the fields were getting touched. Then I passed `errors` in there too, logged it and it was empty everytime for all the fields. Then I logged the `converted` [here](https://github.com/LiberTEM/LiberTEM/blob/master/client/src/dataset/validate.ts#L20) before throwing it, and it shows the expected errors in the logs. So I think the Formik errors aren't being set properly.

sk1p:
> So I think the Formik errors aren't being set properly.

Yeah - it actually was a two-part problem. One, the errors from JSON schema validation were not passed to the right fields of the form, and two, the `TupleInput` was missing a `onBlur` handler - with the usual form fields, the validation errors are only shown once you have "touched" a form field. In Formik, touching means the field was once focused and focus was moved somewhere else. This also makes sense, as you may not want to show your user a form full of errors at the beginning, but only once the user has started to fill fields, validation errors are shown for those fields.

I've pushed a fix for this issue into your branch - feel free to test further and let me know if I missed something!

uellue:
Error message and default value works! Thx for taking care of it. :-)

When it press "reset", it now leaves the field for the dataset name empty. For other file formats it defaults to the filename.

uellue:
> When I press "reset", it now leaves the field for the dataset name empty. For other file formats it defaults to the filename.

Actually, in an anonymous browser window it works. Just a fluke, please ignore!

uellue:
All in all, the form issues seem fixed now, Looks good to me, just needs CI and the rebuilt client, right?

sk1p (approved):

sk1p:
> All in all, the form issues seem fixed now, Looks good to me, just needs CI and the rebuilt client, right?

:+1:

AnandBaburajan:
> Yeah - it actually was a two-part problem. One, the errors from JSON schema validation were not passed to the right fields of the form, and two, the `TupleInput` was missing a `onBlur` handler - with the usual form fields, the validation errors are only shown once you have "touched" a form field. In Formik, touching means the field was once focused and focus was moved somewhere else. This also makes sense, as you may not want to show your user a form full of errors at the beginning, but only once the user has started to fill fields, validation errors are shown for those fields.

Ah, thanks for looking into it.

> I've pushed a fix for this issue into your branch - feel free to test further and let me know if I missed something!

Looks good to me too, and I've pushed the rebuilt client.

sk1p:
Merging, thx again!

PR #1207, opened 2022-02-12
Sparse processing with LiberTEM

### LINKED ISSUE #1205: Sparse version of `MemoryDataSet`
For testing the general pipeline of sparse input data, independently from file format support.

uellue:
Closes #1205

No new dataset required for testing, apparently! MemoryDS does the job.

Still TODO:

- [x] Flag for UDFs that support sparse data -- how to handle different formats?
- [x] Fallback: Convert to dense before passing a tile to a UDF that doesn't support
      sparse arrays
- [x] Check out the new sparse arrays in SciPy 1.8: https://docs.scipy.org/doc/scipy/reference/sparse.html
- [x] GPU support: introduce "flat sig" array format so that 2D arrays work. See https://github.com/LiberTEM/LiberTEM/pull/1317#issuecomment-1225507179 
- [x] Sparse support for LiberTEM UDFs
- [x] In-depth test of all corners of LiberTEM
- [x] Fix integration with LiberTEM-live (currently fails because of the changed interface of `get_tiles`)
- [x] Review and merge https://github.com/LiberTEM/sparseconverter/pull/1

## Contributor Checklist:

## Reviewer Checklist:


uellue:
The new arrays in `scipy.sparse` just change the interface in release 1.8. they seem to use the same back-end implementation and have the same limitation to two dimensions as before.

sparse pydata.org GCXS dot product is almost as fast as the SciPy CSR/CSC implementations. I'll check if they support the same broadcasting etc. as NumPy arrays. That is particularly relevant for binned SSB.

uellue:
Assigning to 0.11 to support event-based detectors.

sk1p:
Could be interesting/useful: [sparse support in pytorch](https://pytorch.org/docs/stable/sparse.html#tensor-methods-and-sparse)

sk1p:
Same in tensorflow: https://www.tensorflow.org/api_docs/python/tf/sparse/SparseTensor

uellue:
FYI: https://github.com/cupy/cupy/issues/7035

src/libertem/utils/devices.py
```diff
@@ -39,20 +40,30 @@ def detect():
...
         cores = 2
-
+    cuda_info = {}
     try:
         cudas = [device.id for device in numba.cuda.gpus]
+        for cuda in cudas:
+            cuda_info[cuda] = {}
+            cuda_ctx = numba.cuda.current_context(cuda)
+            try:
+                cuda_info[cuda]['mem_info'] = cuda_ctx.get_memory_info()
```
sk1p, line 50: IMHO we should unpack this into our own data structure, so the interface isn't defined as "whatever numba does currently". We should own the types used in our interfaces. I wouldn't object to copying the current `numba` behavior like this:

```python
class MemoryInfo(typing.NamedTuple):
    free: int
    total: int
```

sk1p:
I've rebased to the current `master` branch, and there are a few points of contention:

- in #1336 and previous PRs, cluster creation was changed to allow integer arguments, and repeated device IDs etc. - this needs to be double-checked 
    - as commented inline, it would be nice to have the memory info for cuda devices optionally, so people don't always have to run `detect` when creating a cluster, but can simply say "give me 4 workers on cuda device 0" or "use two cuda devices" and have things work out
    - there are functions `assign_cudas` and `make_cpu_plan` - probably they need to be merged into one
- in the stddev UDF, the way we calculate the sum of variances changed - this had a conflict with the fixes for complex numbers. @uellue could you double-check if my port of the fix is correct?

sk1p:
Full CI run with slow tests here:

https://dev.azure.com/LiberTEM/LiberTEM/_build/results?buildId=2052&view=results

sk1p:
Doctest failure:

```
2022-11-08T15:10:13.2496988Z __________________ [doctest] libertem.udf.masks.ApplyMasksUDF __________________
2022-11-08T15:10:13.2497550Z 055 
2022-11-08T15:10:13.2497802Z 056     Examples
2022-11-08T15:10:13.2498637Z 057     --------
2022-11-08T15:10:13.2498917Z 058 
2022-11-08T15:10:13.2499211Z 059     >>> dataset.shape
2022-11-08T15:10:13.2499506Z 060     (16, 16, 32, 32)
2022-11-08T15:10:13.2499791Z 061     >>> def my_masks():
2022-11-08T15:10:13.2500173Z 062     ...     return [np.ones((32, 32)), np.zeros((32, 32))]
2022-11-08T15:10:13.2500603Z 063     >>> udf = ApplyMasksUDF(mask_factories=my_masks)
2022-11-08T15:10:13.2501280Z 064     >>> res = ctx.run_udf(dataset=dataset, udf=udf)['intensity']
2022-11-08T15:10:13.2501897Z UNEXPECTED EXCEPTION: KeyError((None, 'numpy'))
2022-11-08T15:10:13.2502291Z Traceback (most recent call last):
2022-11-08T15:10:13.2502683Z   File "/usr/lib/python3.7/doctest.py", line 1337, in __run
2022-11-08T15:10:13.2503076Z     compileflags, 1), test.globs)
2022-11-08T15:10:13.2503481Z   File "<doctest libertem.udf.masks.ApplyMasksUDF[3]>", line 1, in <module>
2022-11-08T15:10:13.2504262Z   File "/azp/agent/_work/1/s/.tox/py37/lib/python3.7/site-packages/libertem/api.py", line 775, in run_udf
2022-11-08T15:10:13.2504725Z     iterate=False,
2022-11-08T15:10:13.2505413Z   File "/azp/agent/_work/1/s/.tox/py37/lib/python3.7/site-packages/libertem/api.py", line 1038, in _run_sync
2022-11-08T15:10:13.2505931Z     udf_results = run_gen_get_last(_run_sync_wrap())
2022-11-08T15:10:13.2506716Z   File "/azp/agent/_work/1/s/.tox/py37/lib/python3.7/site-packages/libertem/common/async_utils.py", line 95, in run_gen_get_last
2022-11-08T15:10:13.2507239Z     result = gen.__next__()
2022-11-08T15:10:13.2507957Z   File "/azp/agent/_work/1/s/.tox/py37/lib/python3.7/site-packages/libertem/api.py", line 1024, in _run_sync_wrap
2022-11-08T15:10:13.2508404Z     for udf_results in result_iter:
2022-11-08T15:10:13.2509095Z   File "/azp/agent/_work/1/s/.tox/py37/lib/python3.7/site-packages/libertem/udf/base.py", line 2362, in run_for_dataset_sync
2022-11-08T15:10:13.2509600Z     for part_results, task in result_iter:
2022-11-08T15:10:13.2510315Z   File "/azp/agent/_work/1/s/.tox/py37/lib/python3.7/site-packages/libertem/udf/base.py", line 2305, in results_for_dataset_sync
2022-11-08T15:10:13.2511083Z     dataset, executor, roi, corrections, backends, dry
2022-11-08T15:10:13.2512016Z   File "/azp/agent/_work/1/s/.tox/py37/lib/python3.7/site-packages/libertem/udf/base.py", line 2238, in _prepare_run_for_dataset
2022-11-08T15:10:13.2512738Z     udf.init_result_buffers(executor=executor)
2022-11-08T15:10:13.2513465Z   File "/azp/agent/_work/1/s/.tox/py37/lib/python3.7/site-packages/libertem/udf/base.py", line 989, in init_result_buffers
2022-11-08T15:10:13.2514021Z     self.results = UDFData(self.get_result_buffers())
2022-11-08T15:10:13.2514804Z   File "/azp/agent/_work/1/s/.tox/py37/lib/python3.7/site-packages/libertem/udf/masks.py", line 214, in get_result_buffers
2022-11-08T15:10:13.2515382Z     dtype = np.result_type(self.meta.input_dtype, self.get_mask_dtype())
2022-11-08T15:10:13.2516128Z   File "/azp/agent/_work/1/s/.tox/py37/lib/python3.7/site-packages/libertem/udf/masks.py", line 115, in get_mask_dtype
2022-11-08T15:10:13.2516590Z     return self.masks.dtype
2022-11-08T15:10:13.2517273Z   File "/azp/agent/_work/1/s/.tox/py37/lib/python3.7/site-packages/libertem/common/container.py", line 159, in dtype
2022-11-08T15:10:13.2517783Z     return self.computed_masks.dtype
2022-11-08T15:10:13.2518535Z   File "/azp/agent/_work/1/s/.tox/py37/lib/python3.7/site-packages/libertem/common/container.py", line 245, in computed_masks
2022-11-08T15:10:13.2519071Z     self._computed_masks = self._compute_masks()
2022-11-08T15:10:13.2519796Z   File "/azp/agent/_work/1/s/.tox/py37/lib/python3.7/site-packages/libertem/common/container.py", line 219, in _compute_masks
2022-11-08T15:10:13.2520300Z     [to_dense(m) for m in mask_slices]
2022-11-08T15:10:13.2521001Z   File "/azp/agent/_work/1/s/.tox/py37/lib/python3.7/site-packages/libertem/common/container.py", line 219, in <listcomp>
2022-11-08T15:10:13.2521497Z     [to_dense(m) for m in mask_slices]
2022-11-08T15:10:13.2522179Z   File "/azp/agent/_work/1/s/.tox/py37/lib/python3.7/site-packages/libertem/common/sparse.py", line 7, in to_dense
2022-11-08T15:10:13.2522653Z     res = for_backend(a, NUMPY)
2022-11-08T15:10:13.2523385Z   File "/azp/agent/_work/1/s/.tox/py37/lib/python3.7/site-packages/libertem/common/array_backends.py", line 753, in for_backend
2022-11-08T15:10:13.2523971Z     converter = get_converter(source_backend, backend, strict)
2022-11-08T15:10:13.2524784Z   File "/azp/agent/_work/1/s/.tox/py37/lib/python3.7/site-packages/libertem/common/array_backends.py", line 734, in get_converter
2022-11-08T15:10:13.2525296Z     return converters[identifier]
2022-11-08T15:10:13.2526043Z   File "/azp/agent/_work/1/s/.tox/py37/lib/python3.7/site-packages/libertem/common/array_backends.py", line 514, in __getitem__
2022-11-08T15:10:13.2526566Z     return self._converters[item]
2022-11-08T15:10:13.2527063Z KeyError: (None, 'numpy')
2022-11-08T15:10:13.2527739Z /azp/agent/_work/1/s/.tox/py37/lib/python3.7/site-packages/libertem/udf/masks.py:64: UnexpectedException
```

Is this example not something that is also covered by other tests? I wonder why this is so special and fails... 

First debugging: 

- The mask factory returns a list of arrays
- in `MaskContainer._compute_masks`, as the factory is callable, it's called and wrapped in another list, so `mask_slices` is of type `List[List[np.ndarray]]`
- then, for each element of `mask_slices`, we call `to_dense`
- `get_backend(List)` returns `None`
- there's no converter from `None` to `numpy`, ↯

sk1p:
I've now extracted the `make_gpu_plan` changes into #1353 - maybe we can continue the specific discussion there.

Janghyun-jo:
(This is @uellue)

We should convert to tuple before concatenating the shapes :-)

'''
File ~\Documents\src\LiberTEM\src\libertem\io\dataset\raw_csr.py:134, in RawCSRDataSet.initialize(self, executor)
    131     if prod(self._sig_shape) != prod(sig_shape):
    132         raise ValueError(f"Sig size mismatch between {self._sig_shape} and {sig_shape}.")
--> 134 shape = Shape(self._nav_shape + self._sig_shape, sig_dims=len(self._sig_shape))
    135 self._descriptor = descriptor = executor.run_function(get_descriptor, self._path)
    136 executor.run_function(
    137     check,
    138     descriptor=descriptor,
    139     nav_shape=nav_shape,
    140     sig_shape=sig_shape
    141 )

TypeError: can only concatenate list (not "tuple") to list
'''

sk1p:
Pipeline run with tox env recreation: https://dev.azure.com/LiberTEM/LiberTEM/_build/results?buildId=2109&view=logs&j=4d6f0f97-9d62-53bc-79d2-96c85fc226be&t=b5ec37f2-4755-55f4-9304-bcf3d588aca8

uellue:
The issue is reproducible on my laptop and doesn't occur with the master branch. I'll try to find out what is going on here...

sk1p:
Latest recreate-run: https://dev.azure.com/LiberTEM/LiberTEM/_build/results?buildId=2111&view=logs&j=4d6f0f97-9d62-53bc-79d2-96c85fc226be 

Mitigation seems to work, just need to update notebooks; probably makes sense to do this in a separate PR.

uellue:
FYI I've kept the memory dataset fix and stddev fix in this PR since they are interdependent. One could address them separately in the right order, if that is preferred, but it would take some extra work.

uellue:
Possibly one should also distinguish IO sizes (ruled by density) and UDF preferred sizes, where UDFs can maybe even consider the density and array backend they'll receive.

matbryan52 (commented):
Just had a first-pass look through the documentation, particularly (and a quick look through the code). First impression is that everything is described in a technical sense though what is lacking are sufficient examples of a sparse-supporting UDF running on a sparse-generating DataSet, especially a UDF which can support different sparse APIs (or CPU + GPU). This might be a good case for another example notebook.

As a way to learn a bit more about how the new interfaces work I will try creating an example of this type.

src/libertem/udf/base.py
```diff
@@ -1169,6 +1394,26 @@ def get_backends(self) -> BackendSpec:
         '''
         return ('numpy', )
 
+    def forbuf(self, arr, target):
```
matbryan52, line 1397: This is a comment on the previous method `get_backends`, namely the docstring needs updating to reflect the fact the possible return values have changed, and to reflect the support for sparse + gpu.

src/libertem/io/dataset/memory.py
```diff
@@ -221,33 +229,51 @@ def __init__(self, tileshape=None, num_partitions=None, data=None, sig_dims=2,
...
-            self._nav_shape = tuple(nav_shape)
+            nav_shape = tuple(nav_shape)
         if sig_shape is None:
-            self._sig_shape = self.shape.sig
+            sig_shape = data.shape[-sig_dims:]
         else:
-            self._sig_shape = tuple(sig_shape)
-            self.sig_dims = len(self._sig_shape)
+            sig_shape = tuple(sig_shape)
+        reshaped_data = self.data.reshape((-1, *sig_shape))
```
matbryan52, line 255: This reshape will fail if data is `scipy.sparse` or other 2d-only format, yet we supply a 2D sig_shape. The reshaped view is only used to calculate the image_count, so it is possible support this at least for creation of the dataset. As for generating tiles from this combination of 2D scipy.sparse => 3/4D datacube I don't yet know if that's supported?

matbryan52:
> 



> Just had a first-pass look through the documentation, particularly (and a quick look through the code). First impression is that everything is described in a technical sense though what is lacking are sufficient examples of a sparse-supporting UDF running on a sparse-generating DataSet, especially a UDF which can support different sparse APIs (or CPU + GPU). This might be a good case for another example notebook.
> 
> As a way to learn a bit more about how the new interfaces work I will try creating an example of this type.



> Just had a first-pass look through the documentation, particularly (and a quick look through the code). First impression is that everything is described in a technical sense though what is lacking are sufficient examples of a sparse-supporting UDF running on a sparse-generating DataSet, especially a UDF which can support different sparse APIs (or CPU + GPU). This might be a good case for another example notebook.
> 
> As a way to learn a bit more about how the new interfaces work I will try creating an example of this type.

Here's a potential example notebook, let me know if you want me to squash and push it into the PR: https://github.com/matbryan52/LiberTEM/blob/cfd408539531881538a800a686bd5c17bb75e07e/examples/sparse_udf.ipynb

src/libertem/udf/base.py
```diff
@@ -57,26 +63,226 @@
...
     return tmp_dtype
 
 
+class TileConverter:
+    def __init__(self, tile: DataTile):
+        self._tile = tile
+        # prime cache with identity
+        self._cache = {get_backend(tile.data): tile.data}
+
+    def get(self, backend):
```
matbryan52, line 95: I had a bug occur when writing the example notebook where a BACKEND_ALL-supporting udf + a normal MemoryDataSet(dense_array) but with a CUDA-only context, caused the data to convert to from dense numpy cupyx.scipy.sparse.csr_matrix rather than dense cp.array as I would have expected. I can't reproduce it unfortunately, and wondered if this is explicitly handled anywhere ? For a big array sparsification could be quite costly (or even undefined).

uellue:
> > As a way to learn a bit more about how the new interfaces work I will try creating an example of this type.
> 
> Here's a potential example notebook, let me know if you want me to squash and push it into the PR: https://github.com/matbryan52/LiberTEM/blob/cfd408539531881538a800a686bd5c17bb75e07e/examples/sparse_udf.ipynb

Awesome! Yes, feel free to push it. I really like it!

src/libertem/udf/base.py
uellue: Hmm I'll have to check the conversion efficiency matrix. It will choose the conversion based on that, but afair there's also some logic to consider the array format and device class when creating the execution plan.

Nice catch, I'll try to reproduce and see what is going on! I agree that it is sketchy to convert from dense to sparse like that, and probably not what one would expect.

uellue: I think I found the source of the confusion. A dataset indicates preference by the order of backends. The memory dataset only preferred its native type, meaning the remaining ones were in random order, without taking conversion efficiency into account. I am now including a change where it sorts the backends by a cost metric, which prefers same backend, same sparsity and same device class. Within these categories it sorts by the conversion cost of sparseconverter. This requires an update of sparseconverter, too, to have an interface to get the conversion cost.

setup.py
```diff
@@ -166,6 +166,8 @@ def find_version(*file_paths):
         'typing-extensions',  # backwards-compatibility for newer typing constructs
         'opentelemetry-api',
         'tblib',
+        'tomli',
+        'sparseconverter',
```
sk1p, line 170: Should have a lower constraint `>=0.3.0` probably

uellue: 👍

src/libertem/udf/base.py
uellue: Thx, nice catch! Should be fixed now.

src/libertem/utils/devices.py
uellue: Should be addressed in #1353 , resolving here

src/libertem/io/dataset/memory.py
uellue: Nice catch! This particular instance is now replaced. Didn't test if scipy.sparse generally works, though.

uellue: I looked into it and scipy.sparse matrices don't work with MemoryDataSet in general, only sparse.pydata.org arrays. Not sure if this should generate a different error by LiberTEM or if it should just fail naturally.

src/libertem/udf/base.py
uellue: It SHOULD be resolved now. :-)

src/libertem/io/dataset/base/tiling_scheme.py
```diff
@@ -188,6 +188,8 @@ def validate(
     ):
         sig_shape = shape[1:]
 
+        return True
```
sk1p, line 191: Can we have a comment here that justifies disabling the validation code below?

src/libertem/udf/base.py
```diff
@@ -1148,26 +1373,47 @@ def get_tiling_preferences(self) -> TilingPreferences:
...
+        This function should be wrapped around assignment to result buffers if
+        the array :code:`arr` might be sparse and/or have a flattened sig
+        dimension. It converts any of the supported sparse input formats into
+        the array format of the target buffer. If the argument is already matching the buffer
+        array or is not recognized, it is a no-op.
+
+        The result of array operations on a sparse input tile can't always be
+        merged into dense result buffers directly since arrays from the
+        :mod:`sparse` package are not converted to NumPy arrays automatically.
+
```
sk1p, line 1412: As this is one of the more important user-facing functions added, it would be great to have a "Parameters" section here that describes the types, and maybe a small example code snippet.

sk1p:
One more thing: the test case tests/udf/test_memleak.py has regressed - it runs successfully on `master` and fails on this branch. From stdout:

>Round 0
Memory use during UDF run: 0.
Memory increase worker: -400756736.
Memory increase executor: 3539.
Round 1
Memory use during UDF run: 0.
Memory increase worker: 399355904.
Memory increase executor: 43.
Round 2
Memory use during UDF run: 0.
Memory increase worker: 399331328.
Memory increase executor: 529.
Round 3
Memory use during UDF run: 0.
Memory increase worker: 399269888.
Memory increase executor: 1035.
Round 4
Memory use during UDF run: 0.
Memory increase worker: 399282176.
Memory increase executor: -85.

(the worker delta is about 6 times the masks size)

And the final error:

> \>     assert cumulative_worker_delta/rounds/worker_count \< sys.getsizeof(masks) * 0.1
> E       assert ((1196482560 / 5) / 4) < (66453648 * 0.1)
> E        +  where 66453648 = \<built-in function getsizeof\>(array([[[0., 0., 0., ..., 0., 0., 0.],\n        [0., 0., 0., ..., 0., 0., 0.],\n        > [0., 0., 0., ..., 0., 0., 0.],\n ...\n        [0., 0., 0., ..., 0., 0., 0.],\n        [0., 0., 0., ..., 0., 0., 0.],\n        [0., 0., 0., ..., 0., 0., 0.]]]))
> E        +    where \<built-in function getsizeof\> = sys.getsizeof

I can reproduce this very reliably, so something related to these changes is leaking memory (again).

docs/source/api.rst
```diff
@@ -197,7 +197,7 @@ used for computation:
     from libertem.executor.pipelined import PipelinedExecutor
     from libertem.utils.devices import detect
 
-    spec = PipelinedExecutor.make_spec(cpus=[0, 1, 2], cudas=[])
+    spec = PipelinedExecutor.make_spec(cpus=[0, 1, 2], cudas=[], cuda_info={})
```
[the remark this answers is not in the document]
sk1p:
```suggestion
    spec = PipelinedExecutor.make_spec(cpus=[0, 1, 2], cudas=[])
```

sk1p:
Another test failure: 

```
______________________________________________________ test_reshape_sync_offset[False-None-nav_shape1--182] ______________________________________________________

raw_csr_generated = <libertem.io.dataset.raw_csr.RawCSRDataSet object at 0x7f00e247f790>
mock_sparse_data = (<221x456 sparse matrix of type '<class 'numpy.float32'>'
	with 50573 stored elements in Compressed Sparse Row format>... 0., 0., ..., 1., 1., 1.],
       [0., 1., 0., ..., 0., 1., 0.],
       [0., 0., 1., ..., 0., 1., 1.]], dtype=float32))
lt_ctx = <libertem.api.Context object at 0x7f00e1d46f80>, sync_offset = -182, nav_shape = (14, 14), sig_shape = None, use_roi = False

[...]
>       assert_allclose(
            ref_result[1]['std'].raw_data,
            result[1]['std'].raw_data
        )
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=0
E       
E       Mismatched elements: 1 / 456 (0.219%)
E       Max absolute difference: 1.66533454e-16
E       Max relative difference: 4.31088651e-16
E        x: array([[4.948717e-01, 4.791574e-01, 3.499271e-01, 4.517540e-01,
E               4.948717e-01, 5.000000e-01, 5.000000e-01, 4.948717e-01,
E               4.948717e-01, 4.948717e-01, 4.948717e-01, 4.791574e-01,...
E        y: array([[0.494872, 0.479157, 0.349927, 0.451754, 0.494872, 0.5     ,
E               0.5     , 0.494872, 0.494872, 0.494872, 0.494872, 0.479157,
E               0.494872, 0.5     , 0.494872, 0.479157, 0.479157, 0.451754,...

tests/io/datasets/test_raw_csr.py:399: AssertionError
---------------------------------------------------------------------- Captured stdout call ----------------------------------------------------------------------
nav_shape (14, 14)
sig_shape None
sync_offset -182
roi None
4.728505609053946e-16
```

This is a bit brittle, it doesn't happen every time. Re-running it does generate a few failures though, sometimes - it seems to me that having `sync_offset=-182` is the common factor. 

Another one:

```
__________________________________________________________ test_reshape_sync_offset[True-sig_shape2-nav_shape5--228] __________________________________________________________

raw_csr_generated = <libertem.io.dataset.raw_csr.RawCSRDataSet object at 0x7fd73cfbcf10>
mock_sparse_data = (<221x456 sparse matrix of type '<class 'numpy.float32'>'
	with 50243 stored elements in Compressed Sparse Row format>... 1., 0., ..., 0., 0., 0.],
       [0., 1., 1., ..., 1., 0., 0.],
       [1., 0., 0., ..., 0., 1., 1.]], dtype=float32))
lt_ctx = <libertem.api.Context object at 0x7fd73be224d0>, sync_offset = -228, nav_shape = (84, 94, 21), sig_shape = (456,), use_roi = True

[...]


self = <libertem.io.dataset.memory.MemoryDataSet object at 0x7fd73be237f0>

    def get_sync_offset_info(self):
        """
        Check sync_offset specified and returns number of frames skipped and inserted
        """
        if not -1*self._image_count < self._sync_offset < self._image_count:
>           raise DataSetException(
                "sync_offset should be in (%s, %s), which is (-image_count, image_count)"
                % (-1*self._image_count, self._image_count)
            )
E           libertem.io.dataset.base.exceptions.DataSetException: sync_offset should be in (-221, 221), which is (-image_count, image_count)

src/libertem/io/dataset/base/dataset.py:59: DataSetException
```

src/libertem/io/dataset/base/tiling_scheme.py
uellue: Hm, I'm wondering how that ended up here 😅 looks like a quick band-aid fix that should have been removed again. I'll try to keep the validation working, and otherwise put a proper justification.

uellue: Nice catch! It turns out that the shape negotiation was in disarray, actually. It worked well because of a combination of bugs that led to an OK tile shape that got approved by the "True". I'm now working to find a good compromise using `get_max_io_size()`. Performance is not great yet for stddev, at least on my machine and Windows.

uellue:
Converted to draft since I still need to find a good compromise for tile shape negotiation. Performance is really poor on Windows, too.

src/libertem/udf/base.py
uellue: Good point!

src/libertem/io/dataset/base/tiling_scheme.py
uellue: Found a compromise that seems OK. The problem wasn't Windows, but a combination of not many cores, small L3 cache and less-than-ideal GPU support, I suspect way too small tiles.

PR #1209, opened 2022-02-14
Move all CI jobs to self-hosted Azure Agent

sk1p:
As we currently don't have any availability of cloud agents from the
Azure DevOps Pipelines free tier, as a workaround, we are using our own
agent for all tests. This means:

- We currently can't run Windows / Mac OS X tests in CI
- We can't run any tests that involve docker
    - To expand on this: running docker in docker requires, without
      additional configuration, that the outer container is started as
      privileged. This means that certain capabilities are no longer
      dropped by docker, and uid 0 in the inner container is basically
      equivalent in privileges to uid 0 outside.
    - You can also re-use the "host" docker daemon by mounting the
      docker socket into the container, but that too is a security
      issue, as then the container can start other containers which
      can mount arbitrary file system paths.
    - This means a) our `dist` tests aren't being run, and b) our docker
      image isn't built (which can be worked around by using buildah or
      similar, which may not be the worst workaround...)
- CI is no longer automatically run for PRs - it always needs to be
  triggered by a contributor, i.e. member of the LiberTEM github
  organization.

The pipeline definition is, for now, copied to
azure-pipelines-data.yml - this is probably easier than figuring out how
to temporarily make the other pipeline only manually-triggered.

# TODO

- [ ] Update contributor docs

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/azure-pipelines-data.yml
+++ b/azure-pipelines-data.yml
@@ -242,7 +242,7 @@ stages:
         displayName: 'Use Python 3.9'
         inputs:
           versionSpec: '3.9'
-      - bash: sudo apt install -y pandoc graphviz
+      - bash: apt install -y pandoc graphviz
         displayName: install required debian packages
 
       - task: Cache@2
@@ -305,7 +305,7 @@ stages:
         displayName: 'Use Python 3.9'
         inputs:
           versionSpec: '3.9'
-      - bash: sudo apt install -y pandoc graphviz
+      - bash: apt install -y pandoc graphviz
         displayName: 'install required debian packages'
 
       - task: Cache@2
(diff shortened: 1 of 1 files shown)


sk1p:
The "normal" pipeline is disabled for now, via the administrative GUI (by entering the GUI YAML editor, then selecting "Settings" in the "..." menu on the top-right:

image

sk1p:
Failing docs-check probably can be fixed by re-building our agent image, which I'm currently doing...

sk1p:
I'm going to merge this to see if the parts work that need access to credentials and are only running on the master branch :crossed_fingers:

PR #1197, opened 2022-01-28
Changelog, link check, contributor update for 0.9.0

uellue:
Work towards release

* Changelog update:
  * Summary on Dask integration
  * Merge individual snippets
  * Fix internal references
* Use `singularity exec` in documentation to steer users further away from the "footgun" `singularity run`.
* Link check and update
* Add contributors -- anyone else to consider? :-)

Based on #1193 to avoid merge conflicts, in particular reference names.

## Contributor Checklist:

* [x] I have added or updated my entry in [the creators.json file](https://github.com/LiberTEM/LiberTEM/blob/master/packaging/creators.json)
* [x] I have added [a changelog entry](https://github.com/LiberTEM/LiberTEM/tree/master/docs/source/changelog) for my contribution
* [x] I have added/updated documentation for all user-facing changes
* [N/A] I have added/updated test cases
* [N/A] I have included the [rebuilt production build of the client](https://libertem.github.io/LiberTEM/contributing.html?#building-the-client) (only if changes were made to the GUI)

## Reviewer Checklist:

* [ ] `/azp run libertem.libertem-data` passed

PR #1210, opened 2022-02-14
Make `partition_slice` available

uellue:
This was missing when experimenting with
combinations of `process_tile()` and `postprocess()` for SSB.

The correctness of the partition slice is mostly tested within the coordinate calculations already.
Basic correspondence check added as well.

This is based on #1197 for changelog purposes.

## Contributor Checklist:

## Reviewer Checklist:


uellue:
The test case fails, actually: Despite using `process_partition()`, the tile has the shape specified when creating the memory data set.

What is the correct behavior here? Feeding a tile that doesn't have the shape of the partition into `process_partition()` smells like a bug. Probably LiberTEM internals should raise an exception of the dataset produces a wrong kind of tile?

sk1p:
> The test case fails, actually: Despite using `process_partition()`, the tile has the shape specified when creating the memory data set.
> 
> What is the correct behavior here? Feeding a tile that doesn't have the shape of the partition into `process_partition()` smells like a bug. Probably LiberTEM internals should raise an exception if the dataset produces a wrong kind of tile?

Yeah, smells like a bug indeed. Adding a per-partition check for this sounds good to me.

Generally on the PR, I think we need to discuss this still: as we discussed today actually, it would be nice to allow for partitions to span arbitrary ROIs (probably even weirdly structured/interlaced - not directly describable by a flattened slice) instead of only slices. I agree that the the functionality in general needs to be there, but we may want to have a different interface actually.

Can we maybe postpone this to 0.10?

uellue:
Ok, that's a good point. Probably a "partition ROI". Feels similar to #764, actually!

sk1p:
> Ok, that's a good point. Probably a "partition ROI". Feels similar to #764, actually!

Indeed that's very similar :)

uellue:
Discussion with @sk1p : It boils down if self.meta.coordinates is available in get_task_data(), preprocess() and postprocess(). This would be a natural source of information on both the partition size and the position of parts of the partition within the dataset.

sk1p:
Closing in favor of #1397

PR #1212, opened 2022-02-15
Check frame and partition shapes

uellue:
The memory dataset's forced tile shape parameter showed that a faulty dataset can silently break the
contract of an UDF and feed it data that is not a full frame or a full partition. This was not caught previously
because most test UDFs are simple enough to tolerate several tiles per partition even if they declare `process_partition()`.

This PR adds tile shape consistency checks in the UDF runner and adapts tests accordingly.

The HDF5 dataset seems to be an example of invalid partition shapes in the wild, which causes `tests/test_local_cluster.py` to fail.

@sk1p could you maybe have a look at the HDF5 issue?

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/io/dataset/base/partition.py
+++ b/src/libertem/io/dataset/base/partition.py
@@ -221,7 +221,7 @@ class BasePartition(Partition):
     def set_corrections(self, corrections: CorrectionSet):
         self._corrections = corrections
 
-    def get_tiles(self, tiling_scheme, dest_dtype="float32", roi=None):
+    def get_tiles(self, tiling_scheme: TilingScheme, dest_dtype="float32", roi=None):
         """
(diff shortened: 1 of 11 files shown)


sk1p:
> could you maybe have a look at the HDF5 issue?

I'll have a look.

sk1p:
> could you maybe have a look at the HDF5 issue?

This should now be taken care of - it's a bit more than just fixing HDF5, as theoretically (and once #1159 lands, in practice) partitions can now differ in size from the globally negotiated tiling scheme. Now, the `TilingScheme` is updated in case of using `process_partition` to directly match the partition shape.

uellue:
Fix confirmed on my system! LGTM, and thank you for taking care of it. :-)

PR #1215, opened 2022-02-16
Correct shape of buffer views for tile depth 1

### LINKED ISSUE #1203: aux_data: Extra dimension added when tile depth is 1
## How to reproduce

```python
inline_ctx = Context.make_with('inline')
data = np.zeros((2, 17, 23))
auxdata = np.zeros((2, ))
test_ds_1 = inline_ctx.load('memory', data=data, num_partitions=1)
test_ds_2 = inline_ctx.load('memory', data=data, num_partitions=2)

class TestUDF(UDF):
    def __init__(self, testaux):
        super().__init__(testaux=testaux)
    
    def get_result_buffers(self):
        return {}
    
    def process_tile(self, tile):
        print(tile.shape)
        print(self.params.testaux.shape)

test_udf = TestUDF(
    testaux=TestUDF.aux_data(auxdata, kind='nav', dtype=auxdata.dtype)
)
print("tile depth 2")
inline_ctx.run_udf(dataset=test_ds_1, udf=test_udf)
print("tile depth 1")
inline_ctx.run_udf(dataset=test_ds_2, udf=test_udf)
```

## What happens

If the tile depth is 1, an extra dimension is added to the aux data.

```
tile depth 2
(2, 17, 23)
(2,)
tile depth 1
(1, 17, 23)
(1, 1)
(1, 17, 23)
(1, 1)
```

## What should happen

The shape of the aux data item is consistent, independent of tile depth.

```
tile depth 2
(2, 17, 23)
(2,)
tile depth 1
(1, 17, 23)
(1,)
(1, 17, 23)
(1,)
```

uellue:
Closes #1203

I'm not sure what the reason for the special treatment was, but it seems that removing it fixes this bug, and tests still pass.

The special treatment was part of the original implementation of tiled processing in 992fdb621

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -105,6 +105,9 @@ Bugfixes
   and curl if they are defined (:issue:`1138`, :pr:`1139`).
 * :code:`make_dask_array` now works correctly when a :code:`roi` is specified
   (:issue:`933`).
+* Correct shape of buffer views in :meth:`~libertem.udf.base.UDFTileMixin.process_tile`
+  when the tile has depth 1 (:pr:`1215`).
+
 
 Documentation
 -------------
--- a/src/libertem/common/buffers.py
+++ b/src/libertem/common/buffers.py
@@ -492,11 +492,7 @@ class BufferWrapper:
             # optimized builds for performance
             assert result_start < len(self._data)
             assert result_stop <= len(self._data)
-            # shape: (1,) + self._extra_shape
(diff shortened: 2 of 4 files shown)


sk1p (approved):
LGTM!

> I'm not sure what the reason for the special treatment was, but it seems that removing it fixes this bug, and tests still pass.

I also don't remember why it is this way, and I don't think the tests added in that commit covered that part... don't think it is worth to dig this info out right now :)

Issue #1204, opened 2022-02-09
Extend RawDataset to handle sets of raw files with per-file headers (and variants)


matbryan52:
Currently RawDataset can only handle a single raw binary file with an optional sync_offset.

It shouldn't be very hard to extend it to handle lists of raw files, each with their own header (or otherwise). This would handle datasets which are formatted as one-file-per-frame in a binary format.

There are a few additional cases to consider:

- One-file-per-frame, with or without a per-frame header
- One-header-per-file, multiple whole frames-per-file (without separate headers)
- Multiple whole frames-per-file, each with its own header (and potentially an extra whole-file header)
- A group of binary files which are essentially a continuous binary blob split across multiple files (for easier transfer/storage, perhaps). In this case I could imagine per-dataset, per-file or per-frame headers, or no headers at all! Also, consider that frames could be split across files.

PR #1224, opened 2022-03-10
Add DataSet implementation for groups of raw files

### LINKED ISSUE #206: Support for "directory full of binary files" kind of DataSet
Some software, for example in STEM simulation, uses a directory full of small binary files as the result. We need to test if we can support this in a high-performance way, and if we can, add support in LiberTEM.

Generally, we may want to use straight `read` calls, as `mmap` should have more overhead, and we need to copy anyways if we want to stack the images for processing.

It should have similar parameters to the all-in-one-file raw format.

- [ ] implement prototype
- [ ] do basic benchmarking with synthetic data via the Python API
- [ ] if it performs, clean up implementation and see if we want to add it to the UI

matbryan52:
Adds an extension of `RawDataSet` which can handle groups of files and files with frame headers/footers. Responds to #1204 .

Currently the implementation is not merged into `RawDataSet` or the usual LiberTEM `ctx.load` or web GUI endpoints. In a future release it would be possible to design a single class which handles both single-file and multi-file reading (in fact the implementation in this PR already handles single files), but it would entail a change to the RawDataSet API/interface (specifically the `path` argument which might change to plural). For now this class should be considered undocumented / testing-only until the interface is stabilized.

One limitation of the system is that when using `MMapBackend` it is impossible to have a frame header or footer which is not a multiple of `dtype.itemsize`. In this case the dataset raises a `DataSetException` and tells the user to change the backend.

This implementation also contains one or two optimizations specific to handling groups of files which might be useful elsewhere. Specifically:

- The method `check_valid` has been modified to only check 256 files at a time, this prevents an OSError due to too many open files
- The mechanism to infer the size of each file has been written to take advantage of parallel processing if this is provided by the executor (using `executor.map` on chunks of files). This prevents a big slowdown encountered when the class calls `os.stat` on thousands of files, particularly over a network filesystem.
- The number of partitions is adjusted to ensure fewer than 256 files-per-partition, to help avoid the OSError

Closes #206 (Dieter)

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/io/dataset/base/backend_buffered.py
+++ b/src/libertem/io/dataset/base/backend_buffered.py
@@ -183,7 +183,7 @@ class DirectBufferedFile(BufferedFile):
             # Make sure to keep a reference, otherwise
             # it will be closed
             self._fh = win32file.CreateFile(
-                self.path,  # fileName
+                str(self.path),  # fileName
                 win32file.GENERIC_READ,  # desiredAccess
                 win32file.FILE_SHARE_READ
                 | win32file.FILE_SHARE_WRITE
--- a/src/libertem/io/dataset/raw_group.py
+++ b/src/libertem/io/dataset/raw_group.py
@@ -79,14 +79,15 @@ class RawFileGroupDataSet(RawFileDataSet):
                          nav_shape=nav_shape, sig_shape=sig_shape,
                          **kwargs)
         self._path = None
-        self._paths = paths
+        self._paths: List[Union[str, pathlib.Path]] = paths
 
         self._file_header = file_header
         self._frame_header = frame_header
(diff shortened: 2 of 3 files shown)


matbryan52:
Data pipeline tests passed on linux 36=>39 before the push to add test skipping on Mac OSX.

uellue:
@matbryan52 should we try to get this into LiberTEM 0.11?

matbryan52:
> @matbryan52 should we try to get this into LiberTEM 0.11?

Definitely, the functionality for reading the files is already there in the normal RawDataSet, we just have to agree on the API design for taking multiple files.

PR #1099, opened 2021-09-02
Move files around to allow clean MIT licensing

uellue:
Closes #1031

## Contributor Checklist:

* [x] I have added or updated my entry in [the creators.json file](https://github.com/LiberTEM/LiberTEM/blob/master/packaging/creators.json)
* [x] I have added [a changelog entry](https://github.com/LiberTEM/LiberTEM/tree/master/docs/source/changelog) for my contribution
* [x] I have added/updated documentation for all user-facing changes
* [x] I have added/updated test cases
* [N/A] I have included the [rebuilt production build of the client](https://libertem.github.io/LiberTEM/contributing.html?#building-the-client) (only if changes were made to the GUI)

## Reviewer Checklist:

* [ ] `/azp run libertem.libertem-data` passed

PR #1245, opened 2022-04-29
Move files around to allow clean MIT licensing, take 2

### LINKED ISSUE #1031: License review of I/O code
The I/O code of LiberTEM is dual licensed under MIT and GPL, while the rest of LiberTEM is GPL-only.

Since we did a lot of development in the I/O code since the last check, we should have a look again if we depend on any GPL code and resolve that if possible.

uellue:
Re-do #1099 after the changes of the 0.9 release since this caused a massive merge conflict.

Marking as `priority` since we have not too many complex PRs in flight right now and this change is bound to create merge conflicts.

Fixes #1031 

The Executor API is not in `common`, too, since it is used heavily by the Dataset implementations.

Imports that are just there for type checking are not considered since that would also touch the UDFRunner. The affected methods are not implemented in the `common`part, but in subclasses.

## Contributor Checklist:

## Reviewer Checklist:


uellue:
azp run libertem.libertem-data

sk1p (changes requested):
Thank you for getting this rolling again. This looks good, let me collect some comments here:

- `libertem.common.numba.cache.prime_numba_cache` uses UDFs - I would propose to move that one to `libertem.web.dataset`, which is currently its only user.
- The tiling scheme `Negotiator` class does use the UDF interface, it would not work independently from it. It would need some refactoring to make sure this is no longer the case. For example, we could introduce another class, or function, that extracts the relevant information from a list of UDFs and feeds it into the `Negotiator`. Should we do this in this PR, or as a follow-up?
- `libertem.io.utils` imports `libertem.win_tweaks` - should that one move to `libertem.common`?

I think the `libertem.common.executor` -> `liberten.udf.base` import is ok-ish, as it is really only for type checking, and the use is otherwise opaque.

Should we add an item to the checklist for license adherence?

Btw. a helpful tool for reviewing was pydeps. Running `pydeps --show-deps` gives a json file with the import dependencies between modules, which can then be filtered, for example:

```python
import json
with open("deps", "r") as f: deps = json.load(f)
mit_keys = [k for k in deps.keys() if k.startswith('libertem.io') or k.startswith('libertem.common')]
for mit_key in mit_keys:
    print(mit_key, [k for k in deps[mit_key].get('imports', []) if not k.startswith('libertem.common') and not k.startswith('libertem.io') and k.startswith('libertem')])
```

src/libertem/common/udf.py
```diff
@@ -0,0 +1,36 @@
+from enum import Enum
+from typing_extensions import Protocol, TypedDict
+from typing import Union, Literal
```
sk1p, line 3: `typing` -> `typing_extensions` for `Literal`, as it was introduced only in Py3.9 IIRC

uellue: Thx! :-) I'm testing the async PR right now, after that I'll fix it.

sk1p:
LGTM, merging!

Issue #222, opened 2018-12-20
Support for numpy format


sk1p:
Docs: https://docs.scipy.org/doc/numpy-1.15.0/reference/generated/numpy.load.html#numpy.load

Supports mmap, which we will absolutely need to use. Also need to disable the pickling, as it is a trivial untrusted/remote code execution otherwise.

See also #221.

PR #1249, opened 2022-05-05
WIP: npy dataset, refs #222

sk1p:
# TODO
- [ ] write some tests
- [ ] rebuild GUI
- [ ] test on some real world data

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/client/src/dataset/components/DatasetOpen.tsx
+++ b/client/src/dataset/components/DatasetOpen.tsx
@@ -15,6 +15,7 @@ import HDF5ParamsForm from "./HDF5ParamsForm";
 import K2ISParamsForm from "./K2ISParamsForm";
 import MIBParamsForm from "./MIBParamsForm";
 import MRCParamsForm from "./MRCParamsForm";
+import NPYParamsForm from "./NPYParamsForm";
 import RawFileParamsForm from "./RawFileParamsForm";
 import SEQParamsForm from "./SEQParamsForm";
 import SERParamsForm from "./SERParamsForm";
@@ -177,6 +178,11 @@ const DatasetOpen = () => {
             const initial = formInitial && datasetType === formInitial.type ? formInitial : undefined;
             const info = formInfo && datasetType === formInfo.type ? formInfo : undefined;
             return renderForm(<MIBParamsForm {...commonParams} initial={initial} info={info} />);
+        }
+        case DatasetTypes.NPY: {
+            const initial = formInitial && datasetType === formInitial.type ? formInitial : undefined;
+            const info = formInfo && datasetType === formInfo.type ? formInfo : undefined;
+            return renderForm(<NPYParamsForm {...commonParams} initial={initial} info={info} />);
         }
         case DatasetTypes.BLO: {
             const initial = formInitial && datasetType === formInitial.type ? formInitial : undefined;
--- a/client/src/dataset/components/NPYParamsForm.tsx
+++ b/client/src/dataset/components/NPYParamsForm.tsx
@@ -1,0 +1,82 @@
+import { ErrorMessage, Field, FormikProps } from "formik";
+import * as React from "react";
+import { Button, Form } from "semantic-ui-react";
+import { Omit } from "../../helpers/types";
+import { DatasetInfoNPY, DatasetParamsNPY, DatasetTypes } from "../../messages";
+import { getInitial, getInitialName, adjustShapeWithBounds, parseShapeInCommaSeparatedString, validateSyncOffsetAndSigShape, withValidation } from "../helpers";
+import { OpenFormProps } from "../types";
+import BackendSelectionDropdown from "./BackendSelectionDropdown";
+import Reshape from "./Reshape";
+
+// some fields have different types in the form vs. in messages
+type DatasetParamsNPYForForm = Omit<DatasetParamsNPY,
+    "type"
+    | "path"
+    | "nav_shape"
+    | "sig_shape"> & {
+        nav_shape: string,
+        sig_shape: string,
+};
+
+type FormValues = DatasetParamsNPYForForm
+
+type MergedProps = FormikProps<FormValues> & OpenFormProps<DatasetParamsNPY, DatasetInfoNPY>;
+
+const NPYFileParamsForm: React.FC<MergedProps> = ({
+    values,
+    info,
+    isSubmitting,
+    handleSubmit,
+    handleReset,
+    isValidating,
+    onCancel,
+    setFieldValue,
+    setFieldTouched,
+    datasetTypeInfo,
+}) => (
+    <Form onSubmit={handleSubmit}>
+        <Form.Field>
+            <label htmlFor="id_name">Name:</label>
+            <ErrorMessage name="name" />
+            <Field name="name" id="id_name" />
+        </Form.Field>
+        <Form.Field>
+            <label htmlFor="id_io_backend">I/O Backend:</label>
+            <ErrorMessage name="io_backend" />
+            <BackendSelectionDropdown
+                value={values.io_backend}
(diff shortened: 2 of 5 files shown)


matbryan52:
Have started looking at this and writing tests.

One quick observation / question: I believe the arguments `nav_shape` + `sig_shape` are essentially ignored as the DataSetMeta object is instantiated with the shape values from the .npy header in combination with the `sig_dims` argument. Is this intentional or just incomplete? (One of my tests passes conflicting nav/sig/dims arguments to the loader and the final dataset always just matches the original npy file).

I can see a few ways to handle it:

- Setting these values during `ctx.load()` is allowed and later they are checked against the .npy header (probably raising a warning if non-matching). Optionally the `sig_shape` parameter could be used to set `sig_dims` if not explicitly provided. This prioritizes the npy file over user input.
- Allow user-provided `nav_shape` and `sig_shape` to essentially act like `np.reshape()` on the contents of the file, as long as their sum fits within the full shape from the .npy header (otherwise raise an Exception). This is more flexible but obviously needs careful and clear handling of priority of parameters. A potential use case here would be a camera that saves .npy files in a flat nav dimension form `(N, sigy, six)` and then by supplying `nav_shape` to the dataset we read the npy file as it if had a certain scan shape. This priorizes the user input over the npy file.

I am happy to implement either, but I wanted to check first in case you had a preference / behaviour in mind?

sk1p:
> Have started looking at this and writing tests.

Thank you for taking over!

> One quick observation / question: I believe the arguments `nav_shape` + `sig_shape` are essentially ignored as the DataSetMeta object is instantiated with the shape values from the .npy header in combination with the `sig_dims` argument. Is this intentional or just incomplete? (One of my tests passes conflicting nav/sig/dims arguments to the loader and the final dataset always just matches the original npy file).

It's just incomplete, sorry, added to the TODO above.

> * Allow user-provided `nav_shape` and `sig_shape` to essentially act like `np.reshape()` on the contents of the file, as long as their sum fits within the full shape from the .npy header (otherwise raise an Exception). This is more flexible but obviously needs careful and clear handling of priority of parameters. A potential use case here would be a camera that saves .npy files in a flat nav dimension form `(N, sigy, six)` and then by supplying `nav_shape` to the dataset we read the npy file as it if had a certain scan shape. This priorizes the user input over the npy file.

This is how it's handled in other `DataSet` impls, and exactly for the use case you are describing - that's the way to go.

matbryan52:
> > Have started looking at this and writing tests.
> 
> Thank you for taking over!
> 
> > One quick observation / question: I believe the arguments `nav_shape` + `sig_shape` are essentially ignored as the DataSetMeta object is instantiated with the shape values from the .npy header in combination with the `sig_dims` argument. Is this intentional or just incomplete? (One of my tests passes conflicting nav/sig/dims arguments to the loader and the final dataset always just matches the original npy file).
> 
> It's just incomplete, sorry, added to the TODO above.
> 
> > * Allow user-provided `nav_shape` and `sig_shape` to essentially act like `np.reshape()` on the contents of the file, as long as their sum fits within the full shape from the .npy header (otherwise raise an Exception). This is more flexible but obviously needs careful and clear handling of priority of parameters. A potential use case here would be a camera that saves .npy files in a flat nav dimension form `(N, sigy, six)` and then by supplying `nav_shape` to the dataset we read the npy file as it if had a certain scan shape. This priorizes the user input over the npy file.
> 
> This is how it's handled in other `DataSet` impls, and exactly for the use case you are describing - that's the way to go.

Great, that's what I implemented this morning and is in the current version. I took a bunch of tests from test_raw.py to apply in test_npy.py and it seems to correctly handle sync offset and various reshapings.

I'm looking at the conda-forge stuff right now but will do another pass over the tests here and also rebuild/test the web client.

matbryan52:
Notes on the web interface:

- On the Python side the `sync_offset` param in `NPYDatasetParams` was commented out. I've made it live again and it seems to work for the GUI side and with the current dataset implementation I don't see anything odd in the results. Do you remember why it was commented at the time ?
- There is a browser-side check for incompatible sig_shape here : https://github.com/sk1p/LiberTEM/blob/numpy-dataset/client/src/dataset/helpers.ts#L66 which prevents interpreting the frames in the dataset as anything other than the same number of pixels as the auto-detected frame size. The python implementation doesn't enforce this constraint right now. I don't think it's important (it's an edge use case) but I mention it just in case.
- A 3D numpy array `(N, sig_y, sig_x)` is auto-interpreted as a `(50, 1)` array (which is great), though for display purposes a `(1, 50)` array renders in the GUI much better. This could be fixed in `adjustShapeWithBounds` https://github.com/sk1p/LiberTEM/blob/numpy-dataset/client/src/dataset/helpers.ts#L30 for all datasets (in another PR) or specifically for the NPY dataset in `.detect_params`.
- There is a schema JSON linked in the params : `"$id": "http://libertem.org/NPYDatasetParams.schema.json"` but I can't find this. This might not just be for the NPY dataset, though, given that http://libertem.org doesn't seem to exist
- Are there any specific conditions for pushing the updated client JS files ? (npm version etc) I haven't ever pushed a web client update before, I've only built them locally for testing.

sk1p:
> On the Python side the sync_offset param in NPYDatasetParams was commented out. I've made it live again and it seems to work for the GUI side and with the current dataset implementation I don't see anything odd in the results. Do you remember why it was commented at the time ?

Possibly because it didn't work when I started working on the GUI - but don't remember the details. I'm fine with re-enabling it :+1:

> There is a browser-side check for incompatible sig_shape here : https://github.com/sk1p/LiberTEM/blob/numpy-dataset/client/src/dataset/helpers.ts#L66 which prevents interpreting the frames in the dataset as anything other than the same number of pixels as the auto-detected frame size. The python implementation doesn't enforce this constraint right now. I don't think it's important (it's an edge use case) but I mention it just in case.

Ok, yeah, I think we can touch this if anything specific comes up (in one direction or the other - enforcing the compatibility, or relaxing it).

> A 3D numpy array (N, sig_y, sig_x) is auto-interpreted as a (50, 1) array (which is great), though for display purposes a (1, 50) array renders in the GUI much better. This could be fixed in adjustShapeWithBounds https://github.com/sk1p/LiberTEM/blob/numpy-dataset/client/src/dataset/helpers.ts#L30 for all datasets (in another PR) or specifically for the NPY dataset in .detect_params.

I'm fine with solving this in a separate PR (for all datasets).

> There is a schema JSON linked in the params : `"$id": "http://libertem.org/NPYDatasetParams.schema.json"` but I can't find this. This might not just be for the NPY dataset, though, given that http://libertem.org doesn't seem to exist

That URI is just the identifier of the json schema you are currently looking at. It's not what the json-schema people call [Retrieval URI](https://json-schema.org/understanding-json-schema/structuring.html#retrieval-uri).

We were planning to use libertem.org some time ago, but then just stuck with github for hosting docs etc... theoretically these could be changed to at least use a domain we have under our control, i.e. libertem.github.io, which may be less confusing.

> Are there any specific conditions for pushing the updated client JS files ? (npm version etc) I haven't ever pushed a web client update before, I've only built them locally for testing.

npm should be one that uses the current lock file version (`lockfileVersion: 2`), so that would be npm 8.x I think. Generally, the `npm` shipped with the current nodejs LTS version (16.x at the time of writing) should be used.

Make sure to use the included `build_client` command of our `setup.py` file, which takes care of also copying the files where they belong. Then, it's important to `git add` the whole of `src/libertem/web/client/` to make sure all changes are committed, including new and deleted files.

matbryan52:
Tested in the web UI on a real dataset (MIB converted into .npy).

Branch pulled from github to a second (Windows) machine to check the client build, no issues found.

image

PR #1264, opened 2022-06-03
make_dask_array which ensures nav-aligned chunking

matbryan52:
Putting this up to have something to consider.

Still needs specific test cases, the current tests for `make_dask_array` already produced nav-aligned chunks so this is essentially a do-nothing modification for those tests. It is also not obvious how to check if the array is rechunked by dask as it is hard to inspect the task graph (the output of dask.array.reshape will always have nav-aligned chunking by definition!!!). As far as I know Dask doesn't provide a public API to get a list or tree of nodes needed to be computed, so we can't check for rechunk operations. (We could inspect the .visualize output [as SVG] for the word 'rechunk', though !!).

The way I did implemented this is aslo quite hacky, because there we have no interface to specify the partition slices (that I could find), so I monkeypatched the classmethod `Partition.make_slices` for the duration of the chunk generation. This should do the job for all file-based dataset formats (+MemoryDataSet) in LiberTEM except `HDF5DataSet` which uses a completely custom `get_partitions` without using `get/make_slices()`.

Such a slices interface could be made but given how each dataset is in principle free to define its own partitioning methods and partition class I find it hard to see how it could be made general.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/contrib/daskadapter.py
+++ b/src/libertem/contrib/daskadapter.py
@@ -1,26 +1,203 @@
 import dask
 import dask.array
 
+import numpy as np
+import contextlib
+from functools import partial
+from typing import TYPE_CHECKING, Tuple, Optional, List, Dict
 
-def make_dask_array(dataset, dtype='float32', roi=None):
+from libertem.io.dataset.base.partition import Partition
+from libertem.common.shape import Shape
+from libertem.common.slice import Slice
+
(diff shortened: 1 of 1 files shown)


uellue:
I'm putting this on 0.11 preliminarily. It looks like a significant improvement of the Dask integration. :-)

Issue #1271, opened 2022-06-20
Conda-Forge packaging


matbryan52:
Looking into packaging via condaforge, it seems we're all good to go except for two things:

- we depend on https://pypi.org/project/empyre/ which is also not distributed though conda-forge (or conda). According to the doc this is problematic  (see this FAQ entry, for example: https://github.com/conda-forge/staged-recipes#6-do-all-of-my-packages-dependencies-have-to-be-in-conda-forge-already ). As it is a Jülich project you could possibly publish empyre to conda-forge, otherwise we would have to port the code into LiberTEM (or make empyre optional). See this Issue : https://github.com/conda/conda-build/issues/548
- [optional] we don't distribute the tests directory with the PyPi .tar.gz, which means `conda-build` can't run them when compiling the feedstock. This is not blocking as they admit that not all test suites are appropriate or possible - in our case they might be too slow to run anyway! - and we can specify anything as tests including just a simple import or a short test script.

I've set up the recipe / fork / branch here as a project repo: https://github.com/LiberTEM/condaforge-staged-recipes/tree/libertem/recipes/libertem, next step would be to make a PR onto the parent repo for this branch, wait for it to be reviewed and then merged, and we're live.

Other notes:

- I've set @uellue @sk1p and myself as maintainers in the recipe, which should mean we are all able to handle updating the feedstock.
- There is a conda-forge bot which handles updating the feedstock when the PyPi version is updated. This generates a PR which we can then accept to push updates to the feedstock. If the dependencies of the package have changed then we need to update the PR with the new recipe, following the instructions here: https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests

sk1p:
Thanks for looking into this!

> * we depend on https://pypi.org/project/empyre/ which is also not distributed though conda-forge (or conda). According to the doc this is problematic  (see this FAQ entry, for example: https://github.com/conda-forge/staged-recipes#6-do-all-of-my-packages-dependencies-have-to-be-in-conda-forge-already ). As it is a Jülich project you could possibly publish empyre to conda-forge, otherwise we would have to port the code into LiberTEM (or make empyre optional). See this Issue : [pip dependencies  in conda recipe conda/conda-build#548](https://github.com/conda/conda-build/issues/548)

I think it's even possible to publish something to conda as a third party, right? At least that's my mental model of conda, it is more similar to a Linux distribution than it is to just a package index. Nevertheless, we can make this happen.

[...]

PR #1276, opened 2022-06-22
Include tests, Remove pyc / nbc / nbi files in source distribution

### LINKED ISSUE #1275: __pycache__ and *.pyc included in (PyPi) source tarball
While looking into #1271 I think I've found that the PyPi .tar.gz (and probably also the .whl) contain all the .pyc files.

from https://files.pythonhosted.org/packages/17/a4/20eb6a06158774138afbd4118e2800d1e0db54fd5a8136025cd723535ee3/libertem-0.9.2.tar.gz

image

From my understanding .pyc is OS+Python version specific so I think this just inflates the bundle size?

I can reproduce this when running `python -m build` and `python setup.py sdist` locally.

matbryan52:
In response to #1271 and #1275, adjusts `MANIFEST.in` to include:

- tests/
- pytest.ini
- conftest.py
- tox.ini
- test_requirements.txt

and excludes:

- *.pyc
- *.nbc
- *.nbi
- `__pycache__/`

Tested by installing from the .whl output of `python -m build` into fresh py39 and py310 environments. Confirmed that the source distribution does not contain any of the cache files.

Tested also that the web client files are correctly packaged.

Bundle size has dropped from 4.1 MB to 3.0 MB even when including 800 KB of test files !

This should allow running tests when packaging for conda-forge.

## Contributor Checklist:

## Reviewer Checklist:

Closes #1275

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,5 @@
 graft src
+graft src/libertem/web/client/
 include LICENSE
 include README.rst
 graft client
@@ -10,3 +11,4 @@ include conftest.py
 include pytest.ini
 include test_requirements.txt
 include tox.ini
+global-exclude *.pyc *.nbi *.nbc __pycache__
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -5,3 +5,8 @@ graft client
 prune client/node_modules
 prune client/build
 prune client/coverage
+graft tests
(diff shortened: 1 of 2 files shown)


sk1p (approved):
Tested this on a clean clone, seems to work fine, thanks!

PR #1278, opened 2022-06-22
Fix for CoM r / ri inversion when switching mode

### LINKED ISSUE #1191: Annular CoM: glitch when switching mode
When setting a small radius, and then switching the CoM mask shape from a disk cut-off to Annular CoM, the constraint `ri < ro` is not taken into account, and an all-zero mask is the result (and a bit of a broken state in the UI: you need to invert the handles again to make it work, and it's not that obvious).

As this is an edge case with very small radius and not likely to happen in practice, I suggest we leave this one for 0.10, unless we need to rebuild the GUI anyways for another reason.

matbryan52:
Fixes #1191. This is a minimal solution to the problem, I simply check if `r > ri` when switching to Ring mode and otherwise arbitrarily set `ri = r * 0.5` (which matches how the values initialize).

An alternative solution I thought of would be to update `ri` even when in Disk mode as a secondary effect of changing the `r` handle. It wouldn't display, of course, and would maintain the ratio of `r` to `ri` at the expense of losing the 'memory' of `ri` that we currently have when changing from Ring to Disk and back again. (This solution is also a bit more difficult to implement and adds an extra callback / operation per 'move' of the r-handle, and this callback would need to be disabled when actually in Ring mode).

I haven't pushed the updated client yet as I wanted to get your input @sk1p 

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/client/src/compoundAnalysis/components/CenterOfMassAnalysis.tsx
+++ b/client/src/compoundAnalysis/components/CenterOfMassAnalysis.tsx
@@ -317,6 +317,13 @@ const CenterOfMassAnalysis: React.FC<CompoundAnalysisProps> = ({ compoundAnalysi
                     <List.Item>
                         <List.Content>
                             <MaskShapeSelector selectedShape={maskShape} handleChange={(e, data) => {
+                                if (data.value === CoMMaskShapes.RING && ri >= r){
+                                    // Fixes r being set smaller than ri when the ri handle is hidden
+                                    // and the constraint is therefore no longer applied
+                                    // Only possible when switching to RING from DISK
+                                    // Otherwise maintains previous memory behaviour of r and ri
+                                    setRI(r * 0.5)
+                                }
                                 setMaskShape(data.value as CoMMaskShapes)
                             }} />
                         </List.Content>


matbryan52:
Video of the bug even when using a large `r` (relative to the displayed image size):

https://user-images.githubusercontent.com/78845903/175110387-5eb486ee-76b9-4e80-9c0a-7cffadf6a11b.mp4

After switching it appears that I drag the outer handle, but in fact it is the inner handle and the immediate result is to snap due to the application of `handle.apply_constraint`.

It seems like `r` is shared between Disk + Ring, while `ri` is only constrained when the handles of Ring are moved, not at initialization or when switching from Disk > Ring. An alternative solution would be calling `handle.apply_constraint()` on all handles when switching, though this still risks creating a zero-mask if `ri == r`.

sk1p:
> After switching it appears that I drag the outer handle, but in fact it is the inner handle and the immediate result is to snap due to the application of `handle.apply_constraint`.

Yep, exactly.

> It seems like `r` is shared between Disk + Ring, while `ri` is only constrained when the handles of Ring are moved, not at initialization or when switching from Disk > Ring.

Exactly - the constraint is applied when starting to drag and while dragging, not when rendering, which would be necessary for it to update on switching.

> An alternative solution would be calling `handle.apply_constraint()` on all handles when switching, though this still risks creating a zero-mask if `ri == r`.

Yeah, you would have to `setState(applyConstraint(...))` - which is not a good idea to do from outside of the component. I think the current solution works quite well - if you add a changelog entry and a rebuilt client, I'd be happy to merge as-is. :+1:

matbryan52:
> if you add a changelog entry and a rebuilt client, I'd be happy to merge as-is. +1

Thanks for reviewing!

PR #1279, opened 2022-06-23
Pick random port for libertem-server if default/choice not available

### LINKED ISSUE #1184: Pick free port like Jupyter
Right now, `libertem-server` will quit with an error if the default port 9000 is already occupied. If no port is specified and 9000 is already taken, LiberTEM could pick a free one. That's the behavior of Jupyter.

matbryan52:
Closes #1184 

Change the way the server binds to a port `server.bind_socket()` rather than `server.listen()` so that we can explicitly create the socket ourselves and set/access its port. Sets a random port if the default/choice by using the alias port 0 when binding to the socket. Required returning the socket object from server.py:main() so that server.py:run() can launch the correct address in the browser tab.

Tested running/using the server on Windows and Linux. I think there is no direct testing of the server.py:main/run, though, I will  try to look into that.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/docs/source/changelog/misc/pick_free_port.rst
+++ b/docs/source/changelog/misc/pick_free_port.rst
@@ -1,0 +1,6 @@
+[Misc] Server launcher picks free port if default/choice in use
+===============================================================
+
+* The function launching the `libertem-server` application will now
+  pick a free, random port if the default or chosen port is in use.
+  Closes #1184.
--- a/docs/source/changelog/misc/pick_free_port.rst
+++ b/docs/source/changelog/misc/pick_free_port.rst
@@ -1,6 +1,7 @@
-[Misc] Server launcher picks free port if default/choice in use
-===============================================================
+[Misc] Server launcher picks free port if default in use
+========================================================
 
 * The function launching the `libertem-server` application will now
-  pick a free, random port if the default or chosen port is in use.
-  Closes #1184.
+  pick a free, random port if the default port is in use. Choosing 
+  an explicit port will continue to raise an error.
+  Closes :issue:`1184`.
--- a/src/libertem/web/cli.py
+++ b/src/libertem/web/cli.py
@@ -61,4 +61,4 @@ def main(port, local_directory, browser, log_level, insecure, host="localhost",
         numeric_level = getattr(logging, log_level.upper(), None)
         if not isinstance(numeric_level, int):
             raise click.UsageError(f'Invalid log level: {log_level}.\n{log_values}')
-        run(host, port, browser, local_directory, numeric_level, token, preload)
+        run(host, port, browser, local_directory, numeric_level, token, preload, is_custom_port)
--- a/src/libertem/web/cli.py
+++ b/src/libertem/web/cli.py
@@ -48,7 +48,8 @@ def main(port, local_directory, browser, log_level, insecure, host="localhost",
     from libertem.common.threading import set_num_threads_env
     from libertem.common.tracing import maybe_setup_tracing
     token = get_token(token_path)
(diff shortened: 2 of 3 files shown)


sk1p:
I would agree with choosing a random port if the default is already in use - there is a potential issue with choosing a random port if the _explicitly_ stated port is in use: if we are automatically launching the LiberTEM server, for example for integration into jupyter via [libertem-jupyter-proxy](https://github.com/LiberTEM/LiberTEM-jupyter-proxy), the launch will succeed, but the proxy will be set up for whatever random software is listening on the conflicting port. In that case, I would say that failing to launch is better.

So, maybe we could only choose a random port in the default case, and continue with the original behavior if the port was chosen explicitly?

Is there a reason for only listening on IPv4? Previously, IPv6 was supported - I'd like to keep this functionality if possible!

uellue:
> So, maybe we could only choose a random port in the default case, and continue with the original behavior if the port was chosen explicitly?

Sounds reasonable! 👍

matbryan52:
OK, I will work on those changes, should all be possible.

> Is there a reason for only listening on IPv4? Previously, IPv6 was supported - I'd like to keep this functionality if possible!

Mainly laziness on my part, it simplifies unpacking the sockets (When I did not specify I had a single IPv4 socket on Ubuntu but both an IPv4 and IPv6 socket on Windows!). I will work something out.

matbryan52:
Those changes have been made, with some bonus additional handling of IPv6 hostnames.

New behaviour:
- No port specifier -> default to 9000 -> random port if already in use
- Port specified -> raise the address in use OSError raised by bind_sockets

I haven't been able to test a case where the hostname maps to two or more IP addresses (v4 and v6 at the same time, for example), which binds more than one socket, so for now I just launch the browser tab at the `hostname:port` of the first bound socket.

More generally testing all this functionality seems challenging, looking through ./test/server it seems like all the server api tests are done using the custom `class ServerThread(threading.Thread):` and not the actual CLI/server launcher functions. Is that right ?

NOTE: I also adjusted the 'localhost'/insecure check to also accept '127.0.0.1' and '::1' without error. @sk1p can you say if this is a security issue ?

sk1p:
> * No port specifier -> default to 9000 -> random port if already in use
> * Port specified -> Already in use -> raise the address in use OSError raised by bind_sockets

:+1:

> More generally testing all this functionality seems challenging, looking through ./test/server it seems like all the server api tests are done using the custom `class ServerThread(threading.Thread):` and not the actual CLI/server launcher functions. Is that right ?

That's right - the tests in `test/server` are IIRC a bit older than much of the CLI functionality - we should have added tests when the CLI gained more options.

> NOTE: I also adjusted the 'localhost'/insecure check to also accept '127.0.0.1' and '::1' without error. @sk1p can you say if this is a security issue ?

Sounds good!

docs/source/changelog/misc/pick_free_port.rst
```diff
@@ -0,0 +1,6 @@
+[Misc] Server launcher picks free port if default/choice in use
+===============================================================
+
+* The function launching the `libertem-server` application will now
+  pick a free, random port if the default or chosen port is in use.
+  Closes #1184.
```
sk1p, line 6: Nitpick: If you use :issue:\`1184\` here, it will link directly to the issue

sk1p (approved):
Looks good, apart from the nitpick on changelog.

docs/source/changelog/misc/pick_free_port.rst
matbryan52: Made that change, thanks!

sk1p:
Thanks!

PR #1283, opened 2022-07-18
Fix coverage for multiprocessing

### LINKED ISSUE #545: Fix coverage for subprocesses
As a follow-up to #474, we can also improve coverage collection for our worker processes. The plan would look a bit like this:

 * pin coverage.py to a specific version and install that into the docker container
 * run `dask-worker` under coverage.py - this includes
   * adjusting run command to something like `coverage run -p $(which dask-worker) ...`
   * mounting coverage settings into the container (currently in `setup.cfg`)
   * mounting a destination directory for the resulting `.coverage` file(s)
  * setup coverage.py [for subprocess support](https://coverage.readthedocs.io/en/coverage-5.0.2/subprocess.html) - possibly by mounting a `.pth` file into the site packages
 * merging of coverage files, both from the worker containers and from the top-level `pytest` run, via `coverage combine`

Note that this will only fix the coverage for the new distributed tests, not the ones using a localcluster!

sk1p:
Fixes #545 

These days, `pytest-cov` "just works" with threading and multiprocessing, but only if there is no additional direct configuration of `coverage`. This should improve coverage reporting from our integration tests, and is also a good preparation for #1267 

Snuck in a fix for npy tests not working well with `pytest-xdist`.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,9 +9,6 @@ exclude = .git,__pycache__,.tox,build,dist,node_modules,TOXENV
 [coverage:run]
 branch = True
 include = src/
-concurrency =
-    thread
-    multiprocessing
 
 [coverage:report]
 # Regexes for lines to exclude from consideration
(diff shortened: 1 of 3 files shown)


uellue:
LGTM, merging!

PR #1301, opened 2022-07-28
Delete continuous releases using the list endpoint

### LINKED ISSUE #414: github releases: left-over drafts of continuous releases
For some reason, there are sometimes draft releases that litter the releases tab (only for members of the LiberTEM repo, but still...). I already deleted some, but here is a recent example: https://github.com/LiberTEM/LiberTEM/releases/tag/untagged-38436bc187afc7aa3aa8

This is somewhat strange, as all releases are created with `draft: False`...

sk1p:
Probably fixes #414

For some reason, `repo.get_release("continuous")` does not return a release object, which is probably the reason we had leftover releases. List and filter instead.

Scheduled for 0.11 as I don't want to touch the release script shortly before 0.10.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/scripts/release
+++ b/scripts/release
@@ -318,12 +318,17 @@ def upload_to_github(verbose, files, token, release):
             "draft": False,
             "prerelease": True,
         })
-        # first, delete existing continuous tag and release:
+        # first, delete existing continuous tag and release(s):
         try:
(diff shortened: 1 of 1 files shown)


uellue:
Thx, merging!

PR #1317, opened 2022-08-17
WIP: Timepix3 prototype parser

matbryan52:
Adds a prototype notebook for parsing a raw Timepix3 file as sparse frames.

Still a WIP.

Links to #1207 

## Contributor Checklist:

## Reviewer Checklist:


matbryan52:
The PR has been updated with a prototype `DataSet` implementation and an accompanying notebook. I'm copying the last cell here to talk about its limitations:

- The `hits => dense => sparse` processing chain. Improving this will require some kind of low-level implementation for de-duplicating hits and summing their (corrected) ToT values.
- Not tested in detail for correctness when mapping timestamps to frames.
- Support for timer rollover is present but flaky at best.
- Mapping timestamps to frames is arbitrary because we have no external trigger/markers for the frames. 
    - Right now by default the (approximate) full span of time in the file is split evenly between the arbitrary number of frames we choose.
- No Time-Over-Threshold correction implementation. 
    - In the upstream library there is a `(1024, 256, 256, number_of_chips)` lookup table which is used to correct each-and-every decoded event. Inspecting the LUT it's uint16 therefore ~512 MB of data to share between all the workers somehow. I would imagine it's also sensor specific!
- No Time-Of-Arrival phase correction. 
    - I think this is related to certain columns / rows of each chip being slightly faster/slower than others and thefore requiring a constant offset. (Might also be chip/camera-specifc and dependent on the low-level configuration of the acquisition).
- No 'Marker Pixel' correction
    - It seems in the upstream library that one pixel of one chip has special properties and needs its hits correcting / deleting. As I am not sure how this works yet I've chosen not to implement it. 
    - I did notice one pixel in the above datasets seems to be 'hot', it is possibly linked.

uellue:
Regarding the sorting of array indices: CuPy can do this about two orders of magnitude faster than the CPU even with transfer to and from the GPU. It just requires quite a bit of GPU RAM. It seems that GPUs are extremely good at sorting large arrays of uint16.

With respect to #1207, it could be worth to implement sparse input on GPU after all. One could introduce a temporary convention that sparse input arrays are always flattened in sig dimension. Since both input tiles and result buffer views are guaranteed to be c-contiguous, one can just always flatten at negligible cost. I'll introduce a special 'flat sig' array format flag to indicate UDFs that accept such tiles.

matbryan52:
Closing as even as a prototype it's not the direction we've taken

Issue #1313, opened 2022-08-15
Reduce overhead of Slice and Shape calculations


uellue:
`get_contiguous_view_for_tile()` can take a significant portion of an UDF run if many `kind='sig'` buffers are involved. It seems that methods of `Slice` and `Shape` take a majority of that time.

# How to reproduce

Reproducer depends on #1207 

```python

import sparse
import numpy as np
import cProfile

from libertem.api import Context
from libertem.udf.sum import SumUDF
from libertem.udf.stddev import StdDevUDF

data = sparse.random((7000, 11, 13), format='coo').astype(np.float32)

ctx = Context.make_with('inline')
ds = ctx.load('memory', data=data, num_partitions=1, tileshape=(1, 11, 13))


def doit():
    return ctx.run_udf(dataset=ds, udf=[SumUDF(), StdDevUDF()])


doit()

with cProfile.Profile() as pr:
    doit()

pr.dump_stats('sparseds.prof')

```

```bash
pip install flameprof
python -m flameprof sparseds.prof > output.svg
```

# What happens

![sparseds](https://user-images.githubusercontent.com/25689052/184621658-5ad3d0f8-1a57-4d25-b89e-2b00f621f048.svg)

# What should happen

The calculations in `Shape` and `Slice` are slicing of short sequences and simple integer math. That means it should be possible to implement this efficiently enough so that they don't appear in a profile.

PR #1321, opened 2022-09-13
(Partially) optimise Shape/Slice

matbryan52:
Addresses #1313 at least for the method `BufferWrapper.get_contiguous_view_for_tile`.

I don't have an infallible way to prove the performance advantage yet, but comparing cProfiles for before/after this change, on the same data/tiling as in #1313 but as a dense `MemoryDataSet` + `InlineJobExecutor`:

Before:
image

After:
image

`get_contiguous_view_for_tile` doesn't show in the _After_ image, but the time spent in this function is:

image

compared to 44% of `run_for_partition` time before these changes.

Summary of changes / rationale:

- Where possible avoid calling the `Shape` and `Slice` constructors, which happens a lot throughout the code and adds overhead everywhere.
 - Adds `NavOnlyShape` and `SigOnlyShape` with optimised constructors + methods, as sub-classes of `Shape`
- Don't construct `Slice` objects to use as keys in `BufferWrapper._contiguous_cache`, only to subsequently convert the `Slice` back to a tuple of `slice` objects for slicing the data. Instead the keys are now a native tuple `(origin_tuple, shape_tuple, sig_dims)` which can still be cached, and converted directly into `tuple(slice)` (could be converted to `namedtuple`)
- `lru_cache` the conversion from the `key` tuple to `tuple(slice)` as the slices we need are repeated in each partition

These optimisations are focused on `get_contiguous_view_for_tile` but I think similar changes could be made in lots of places in the codebase where we repeatedly access `shape.sig`, `shape.dims` etc. One thing to do in the future would be to use `functools.cached_property` where possible (currently only py3.8 but it can be backported) as I believe the assumed behaviour is that Shapes and Slices are effectively immutable? Another possibility is actually subclassing `tuple` to implement `Shape`, which I tried here but found it didn't help much in this case. We should also look into optimisation of `UDFData._get_buffers` which is the next lowest hanging fruit.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/docs/source/changelog/misc/slice_shape_optim.rst
+++ b/docs/source/changelog/misc/slice_shape_optim.rst
@@ -1,0 +1,6 @@
+[Misc] Optimise Slice/Shape behaviour for tiles
+===============================================
+
+* Reduces overhead of (re-)constructing :code:`Slice` and :code:`Shape`
+  objects when slicing tiles, in particular focused on the method
+  :code:`BufferWrapper.get_contiguous_view_for_tile()`. (:issue:`1313`)
--- a/src/libertem/common/buffers.py
+++ b/src/libertem/common/buffers.py
@@ -557,12 +557,15 @@ class BufferWrapper:
         '''
         if self._kind == "sig":
             for key, view in self._contiguous_cache.items():
-                sl = key.get(sig_only=True)
+                sl, _ = self._slice_from_key(key, tuple())
+                _, _, sig_dims = key
+                sl = sl[-sig_dims:]
                 self._data[sl] = view
-                if debug and not disjoint(key, self._contiguous_cache.keys()):
-                    raise RuntimeError(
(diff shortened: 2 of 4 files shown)


matbryan52:
The notebook and integration_live test failures are not due to this PR, instead:

- ` RuntimeError: DectrisAcquisition has additional dependencies; please run `pip install libertem-live[dectris]` to install them.` => change the CI config?
- Notebook matplotlib size change in `hyperspy-integration.ipynb`:

```
 assert reference_output == test_output failed:

  '<Figure size... with 2 Axes>' == '<Figure size... with 2 Axes>'
  - <Figure size 1200x800 with 2 Axes>
  ?              ^^^^ ^^^
  + <Figure size 864x576 with 2 Axes>
  ?          
```

uellue:
Wow, that's quite an improvement! Many thanks, merging!

Issue #1282, opened 2022-07-18
`ValidationUDF`: Validate that all data of a dataset is seen exactly once


uellue:
Currently, the `ValidationUDF` only checks if the data it receives matches the reference. In the future it could also keep track which parts of the data it has seen to make sure none are skipped or processed twice.

sk1p:
Sadly not as easy as I first thought: adding a check in `get_results` doesn't work in a straight-forward way, as we don't know if we have the final result in our hands, or if it is only a partial result. We can't easily check a "should eventually be equal to" constraint in that way... so something like this fails:

```diff
diff --git a/tests/utils.py b/tests/utils.py
index 9dd04dd2..848b141f 100644
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -167,15 +167,23 @@ class ValidationUDF(UDF):
 
     def get_result_buffers(self):
         return {
-            # Just a buffer to "feel" the av shape
-            'nav_shape': self.buffer(kind="nav", dtype="float32"),
+            'seen': self.buffer(kind="nav", dtype=np.int64),
         }
 
     def process_tile(self, tile):
+        self.results.seen[:] += 1
         assert self.params.validation_function(
             self.meta.slice.get(self.params.reference), tile
         )
 
+    def get_results(self):
+        if self.meta.roi is None:
+            expected = self.meta.dataset_shape.size
+        else:
+            expected = np.sum(self.meta.roi)
+        assert np.sum(self.results.seen) == expected
+        return {}
+
```

Might need checks at the call site of the UDF, or maybe a small addition to the UDF API?

PR #1322, opened 2022-09-13
ValidationUDF tracks which data has been 'seen'

matbryan52:
Addresses #1282.

Overrides `UDF._do_get_results(self)` on `ValidationUDF` to verify that every frame is fully tiled and visited during the processing, taking into account the ROI (if provided).

~~As tiled processing visits each frame potentially multiple times, the raw 'seen'/'visited' values can be >= 1, here I've made the assumption that the tiling code is robust/correct so I don't verify that each pixel of each frame was seen, only that each frame was split in to the same number of tiles (or never seen at all in the partition).~~

~~Needs test cases to verify this actually raises when a dataset is not visited completely, which will require some kind of deliberately broken over-ride implementation of `get_tiles`.~~

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/tests/udf/test_simple_udf.py
+++ b/tests/udf/test_simple_udf.py
@@ -26,7 +26,8 @@ def test_validation(lt_ctx):
 
     udf = ValidationUDF(reference=data.reshape((-1, 16, 16)))
     res = lt_ctx.run_udf(dataset=dataset, udf=udf)
-    assert res['nav_shape'].data.shape == (16, 16)
+    assert res['seen'].data.shape == (16, 16)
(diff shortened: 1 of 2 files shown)


uellue:
@matbryan52 ist this ready for review? :-)

matbryan52:
> @matbryan52 ist this ready for review? :-)

I think so 👍 . It can always be made more strict but the implementation here is already an improvement (without changing prior functionality).

uellue (approved):
Looks good, many thanks!

PR #1328, opened 2022-09-29
DM4 DataSet + Fortran Array reader

matbryan52:
This adds a `DM4DataSet` class which is capable of detecting if the file is new-style C-ordering or old-style F/C-hybrid-ordering and reads it accordingly. The dataset has also been added to the web interface.

I looked around for an example of a new-style file and found one here https://zenodo.org/record/7041997 The new-style file contains an _int_ flag under:

```python
tags['ImageList'][dataset_key]['ImageTags']['Meta Data']['Data Order Swapped']
```

which is not present in old-style files. I am assuming this indicates that the data is fully C-ordered on disk. For files which have this `flag == 1` the file is read identically to a `RawDataSet` with a large file header.

If the flag is missing or 0 then the `DM4DataSet` assumes F-like ordering and uses a custom backend to read the file (over-rideable with a parameter).

The C/F files can be visualised on disk as a sequence of `nav_size` length arrays each containing all of the values for a single sig pixel - this is a very efficient layout for virtual imaging! The challenge adapting this to LiberTEM  is that each partition is required to cover the whole signal dimension, but only a fraction of the nav dimension, so we have to read very sparsely from the whole file in every partition. The worst case is building a frame, where we have to sample the whole file like this `raw_data[:, flat_nav::nav_size]`.

The custom backend breaks the data on disk into chunks which are similar to tile stacks with a `depth == nav_size` (they are chosen to align with tiling scheme boundaries where possible). Each partition reads all the tiles it can from a chunk before moving on to the next one. If it has to yield frames rather than tiles then chunks are still used, accumulating 'tiles' from each into a buffer, after which we can yield frames as normal.

A previous implementation tried to parallelize the read operations which caused problems with Dask worker memory (the memmaps were interpreted as a potential memory leak and the worker was killed). This new implementation only has one memmap open at a time, of a maximum size constrained to fit into a reasonable worker memory budget, and does no parallel reads (except via the normal Partition / Worker process mechanism). In an ideal world the reads from each parallel worker/partition would be synchronized, meaning the data is read once and subsequently taken from the page cache. This is probably true near the beginning of the processing but I am sure the workers go gradually out of sync over time.

Still some work to do, namely:

* [x] Support `sync_offset` in the C/F-ordered case
* [x] Adapt the read depth to attain a particular pre-allocated buffer size in bytes
* [x] See if there is any advantage to using `memmap` + `np.frombuffer` rather than `np.memmap`.
* [x] Look into unmanaged worker memory warnings from Dask when running this process-parallel on a big file. With the chunked read strategy I no longer leak memory (and each partition re-creates the memmaps), but it's something to guard against!
* [x] Propery parse a DM3/4 file containing an image stack (hologram), with web client support
* [x] More thorough benchmarking + tune tileshape / partition size / buffer size parameters
* [ ] Complete test cases
     * Misc tests on methods (e.g. `detect_params`...)
     * Add real data tests on an appropriate DM4 file
* [x] Reconcile this `DM4DataSet` with the existing `DMDataSet`, both of which support (in principle) .dm3/.dm4

When checking the performance so far, it's not bad when the file can fit into RAM (unsurprisingly!) as the memmaps are just soliciting the cache after a short time. On a beefy Windows machine I was able to process a 14 GB file at 1.1 GB/s with 12 workers. Small files on small laptops (Windows/Linux) are also fine. The same 14 GB file on a 4-core, 16 GB RAM machine takes about 4 minutes to process, whether I use Dask or Inline, which is expected. This performance is fine for scripting (we could throw a warning), but is a bit annoying in the web interface. I want to think about how we could make the web interface better for slow analyses (which we've mentioned before as a necessary modification).

NOTE, there is a single commit (reverted) in this PR which adds a mechanism to create F-ordered LiberTEM partitions by unrolling the nav dimensions differently. This is not actually a requirement of the dataset but might be a nice feature to consider?? It can be used to create a RawDataSet which will read an F-ordered array from disk using this custom backend.

## Contributor Checklist:

## Reviewer Checklist:


matbryan52:
Some more comments on the memmap + memory usage warnings when running in a `DaskJobExecutor`.

- Dask considers the sum of all accesses to a memmap object as part of 'unmanaged memory' - and hence will warn/kill workers if they keep the same big memmap open for too long while continuously reading from it.
  - I can get Dask to kill my workers even using a normal C-order `RawFileDataSet` + a big file if I constrain the number of partitions to 1 (on both Linux and Windows + MMapBackend).
  - On Windows, the memory use grows under 'working set (shared)', which as I understand it is memory which can be seen by the process but doesn't solely belong to the process (i.e. it's shared with any other process which wants to use the same memmap).
  - I think this is never a problem normally as we create/destroy the file memmap on each new partition, and partitions are sized at ~512 MB so we never load enough of the file to trigger Dask.
- These F-ordered files are really bad for this as to build a single frame / tile we have to traverse the whole file. If I don't re-create the memmap(s) periodically then Dask sees excessive memory leaking and kills workers.
  - The file structure also means that we have to read a full page into memory for each sig pixel in the tile stack, even if we only use a few values (== tile_depth) from the page. So we can hit the worker memory limit well before actually having 'read' the whole file!
-  Seeking / Reading the file directly avoids this unmanaged memory problem, though the overheads are large as the reads for one tile are extremely sparse.
    - One possibility would be to read tile stacks with a depth that matches the page size and as small a tile-shape as necessary, but this doesn't work well for frame-based processing.

I've looked at `BackendBuffered` to try to see how to adapt it, but for now the changes needed are escaping me (without substantially re-implementing it). It'd be good to talk again about this @sk1p when you have a moment. The other solution is to use something a bit more custom like the current PR which is much less effort, but not re-usable.

matbryan52:
So I did a bit of benchmarking to understand the behaviour of the custom reader, and the results confirm what we expect: the fewer passes we can do through the file, the faster it will be. Further tests (not shown) confirmed that reading by tiles favors as deep a stack as possible within a given tile byte size (i.e. smaller signal slices, larger depth).

The effect of number of partitions (inline executor):

image

The size of the buffer we read into and the size of the file chunk we read from at one time (inline executor):

image

And the number of workers (dask executor):

image

The last one is particularly interesting to me because it shows that for simple work (here just `SumSigUDF`) we're completely limited by the disk / memory cache, in a way which dominated the overhead from communication / creating extra tasks. I would imagine though that if the task burden is on the computation itself then the scaling behaviour would be more advantageous towards having extra workers.

Based on all this I adapted the dataset to use large partitions (up to 2 GiB but still respecting the number of cores/workers), a max memmap of 1 GiB (which is necessary to limit apparent RAM usage), and a buffer of 128 MiB to accumulate tiles. A future modification would be to adapt these numbers at runtime to the available resources on the worker (starting by recognizing if we even have to adhere to these limits at all to avoid triggering Dask).

matbryan52:
A (growing) list of Zenodo datasets which could be used to test this PR in CI:

- https://zenodo.org/record/4307783
  - provides a 1GiB C-ordered 4D-STEM file
- https://zenodo.org/record/7041997
  - Al_irradiated.dm4 - a 3D-STEM image (linescan) in C-order
  - several larger 4D stem files in C-order
- https://zenodo.org/record/2580185
  - The dm4 file is a F/C-ordered EELS-STEM image (3D)
  - also contains a 2D .dm3 file
- https://zenodo.org/record/4964416
  - Contains .dm3 format TEM stacks (timeseries_14.dm3 among others)
- https://zenodo.org/record/4508831
  - Contains several dm4 format 2D TEM images

I'm still looking for:
- dm3 format EELS-stem image (always F/C, I presume) (I have these locally I just have to ask permission)
- dm4 format EELS-stem with C-order, if this exists as an option in GMS
- A small 4D-STEM dm4 file in F/C order (I have these locally I just have to ask permission)

sk1p:
> * dm4 format EELS-stem with C-order, if this exists as an option in GMS

We'll try to acquire one of these - right now, our GMS version is most likely too old.

sk1p:
I've decided to tackle this together with the other `load`/`glob`/schema/... related things in the next milestone. If you agree, we can  focus on these issues in 0.12.

matbryan52:
> I've decided to tackle this together with the other `load`/`glob`/schema/... related things in the next milestone. If you agree, we can focus on these issues in 0.12.

OK by me. Though if we wanted to get something released here it could be just the functionality for DM4 files in C-order, which uses the normal I/O backends, and raise NotImplemented if we don't see the right file flag. It's a low hanging fruit which adds a new format?

matbryan52:
Closing this in favor of a version only supporting C-ordered files, for now, with a view to getting some of the functionality merged in time for 0.11. In practice the F-ordered file reader has poor scaling performance with the LiberTEM default behaviour, and requires careful choice of number of workers etc to be nice to use (even if it will eventually get results on these files).

PR #1334, opened 2022-10-12
Handle bad input / errors for Web local cluster creation

matbryan52:
In the web UI it is possible to send negative integers or an empty string for the `numWorker` argument in local cluster creation. It seems like this could be handled on the React side but given there is already a mechanism to signal errors from Python concerning cluster creation/connection, I've added the necessary try/except and exceptions to notify about the bad inputs. As a bonus if we have other errors during creation of `DaskJobExecutor` this should at least propagate the exception to the web interface.

## Contributor Checklist:

## Reviewer Checklist:


sk1p (approved):
Looks good to me! I'll take care of the CI issue and get this merged.

PR #1337, opened 2022-10-18
`make_slices` obeys `num_partitions` (avoids hanging frames)

### LINKED ISSUE #1335: `Partition.make_slices()` can create very small 'final' partitions for odd shapes
The `make_slices` method can sometimes generate oddly sized partitions if the input shape is not simply divisible by `num_partitions`

```python
    @classmethod
    def make_slices(cls, shape, num_partitions, sync_offset=0):
        ...
        num_frames = shape.nav.size
        f_per_part = max(1, num_frames // num_partitions)
        c0 = itertools.count(start=0, step=f_per_part)
        c1 = itertools.count(start=f_per_part, step=f_per_part)
        
        for (start, stop) in zip(c0, c1):
            if start >= num_frames:
                break
            stop = min(stop, num_frames)
            ...
```

For example: `shape = (8190, ...)` with `num_partitions = 4` generates slices for 5 partitions of size `(2047, 2047, 2047, 2047, 2)`, which is fairly sub-optimal! (and doesn't obey the `dataset.get_num_partitions()` method).

We're already importing `numpy` in this file, so an easy solution is the following using `np.linspace`, though the same functionality can be written using `itertools`:

```python
        num_frames = shape.nav.size
        boundaries = tuple(
                np.linspace(
                        0,
                        num_frames,
                        num=num_partitions + 1,
                        endpoint=True,
                        dtype=int
                    )
                )
        for (start, stop) in zip(boundaries[:-1], boundaries[1:]):
            ...
```

which would generate `(2047, 2048, 2047, 2048)`.

What I might have missed is a particular reason for the odd behaviour above ?

matbryan52:
Closes #1335 . Uses the implementation with `np.linspace`.

This required making a change to `MemoryPartition.get_tiles()` to propagate `tiling_intent` where available, which I think was actually a bug/oversight.

I also had to modify a number of test cases (including core tests in `tests/udf`) because they had been coded both for the old implementation of `make_slices` and the old implementation of `MemoryPartition.get_tiles()`. It seems like there were a few cases which previously were expected to raise with `pytest.raises` but now actually pass as with the other parametrizations of those tests. These changes should be thoroughly reviewed!

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/examples/Introduction to UDFs.ipynb
+++ b/examples/Introduction to UDFs.ipynb
@@ -85,7 +85,7 @@
        " {'name': 'Data kind', 'value': 'u'},\n",
        " {'name': 'Layout', 'value': '(1, 1)'},\n",
        " {'name': 'Partition shape', 'value': '(2075, 256, 256)'},\n",
-       " {'name': 'Number of partitions', 'value': '33'},\n",
+       " {'name': 'Number of partitions', 'value': '32'},\n",
        " {'name': 'Number of frames skipped at the beginning', 'value': 0},\n",
        " {'name': 'Number of frames ignored at the end', 'value': 0},\n",
        " {'name': 'Number of blank frames inserted at the beginning', 'value': 0},\n",
--- a/nbval_sanitize.cfg
+++ b/nbval_sanitize.cfg
@@ -17,3 +17,7 @@ replace: DEFAULT_REPR
 [figure_size]
 regex: Figure size [0-9]+x[0-9]+
 replace: FIGURE
+
+[num_partitions]
+regex: 'name': 'Number of partitions', 'value': '[0-9]+'
+replace: NUM_PART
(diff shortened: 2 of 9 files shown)


sk1p:
The failing test is a notebook that included the number of partitions in its output. This should be easy to fix; if you like, you can also add an ignore pattern to [nbval_sanitize.cfg](https://github.com/LiberTEM/LiberTEM/blob/master/nbval_sanitize.cfg) to make the test independent of future partitioning changes

sk1p:
> I also had to modify a number of test cases (including core tests in `tests/udf`) because they had been coded both for the old implementation of `make_slices` and the old implementation of `MemoryPartition.get_tiles()`. It seems like there were a few cases which previously were expected to raise with `pytest.raises` but now actually pass as with the other parametrizations of those tests. These changes should be thoroughly reviewed!

Looks correct, and indeed the tiling intent not being propagated was the issue.

PR #1338, opened 2022-10-18
Choose square `nav_shape` if detected `num_frames` is square (web-only)

### LINKED ISSUE #1309: Guess a square nav_shape if a square number of frames detected in the web interface
For datasets which define their sig shape but not their nav shape (e.g. non-integrated Merlin), if a square number of frames is detected we could guess a square scan grid given how common that choice is in reality.

matbryan52:
Fixes #1309 

Required updating tests for MIB and MRC `detect_params`. Quite difficult to test this for the other datasets affected as we don't necessarily have data recorded on a square nav grid (apart from mocking the data).

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/docs/source/changelog/misc/square_nav_auto_web.rst
+++ b/docs/source/changelog/misc/square_nav_auto_web.rst
@@ -1,0 +1,7 @@
+[Misc] Guess a square nav_shape in the web UI if no other information
+=====================================================================
+
+* If a dataset does not declare its navigation shape then guess
+  a square nav_shape in the Web UI if the number of detected frames
+  is a square number. This can occur for the MIB, TVIPS, SEQ, MRC
+  and K2IS dataset formats. (:issue:`1309`)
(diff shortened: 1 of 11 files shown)


matbryan52:
> Test failure could be related to our [quite low `cleanup_timeout` of 0.5s](https://github.com/LiberTEM/LiberTEM/blob/332d731da082c4156e3752d8aa298b05de336273/tests/executor/test_pipelined.py#L32) - I _think_ if enough tasks of `FailEventuallyUDF` choose to sleep, the timeout could be hit. as we have 32 partitions in the test case, and each task can sleep for max 0.1 seconds, the total run time is bounded by something like 3.2s
> 
> Increasing this value should only influence the run time of tests that handle exceptions, or where the workers for some reason take a long time to die on `close()`. Could you try to maybe put a value of 5s and rerun tests?

Seems to have done the trick, though I guess this is a random failure rather than consistent. Might be worth an issue ? (I don't 100% understand the effect of `cleanup_timeout` so I'd prefer if you made it!)

PR #1343, opened 2022-10-24
Make sure the `Context` cleans up after itself

### LINKED ISSUE #1339: `PipelinedExecutor` shutdown errors (cloudpickle teardown)
I'm finding that `PipelinedExecutor` is failing to shutdown correctly (hangs when the interpreter exits) with the following MWE. Difficult to diagnose but the tracebacks after `KeyboardInterrupt` seem to imply that an error was encountered on the main process during `PipelinedExecutor.close()`, specifically in the call to `SimpleMPWorkerQueue.put` to send `SHUTDOWN`:

One one machine (running current `master`):
```
KeyboardInterrupt: 
Exception ignored in: <function PipelinedExecutor.__del__ at 0x7f17011bf5b0>
  File "/home/bryan/Workspace/lt_fork/src/libertem/executor/pipelined.py", line 914, in __del__
  File "/home/bryan/Workspace/lt_fork/src/libertem/executor/pipelined.py", line 892, in close
  File "/home/bryan/Workspace/lt_fork/src/libertem/common/executor.py", line 490, in put
  File "/home/bryan/miniconda3/envs/lt_fork/lib/python3.10/site-packages/cloudpickle/cloudpickle_fast.py", line 73, in dumps
  File "/home/bryan/miniconda3/envs/lt_fork/lib/python3.10/site-packages/cloudpickle/cloudpickle_fast.py", line 633, in dump
AttributeError: 'NoneType' object has no attribute '__newobj__'
```
on another (ignore the line numbers in this one):
```
KeyboardInterrupt
Exception ignored in: <function PipelinedExecutor.__del__ at 0x7f74b1bcba60>
  File "/home/mat/Workspace/libertem_dev/libertem_fork/src/libertem/executor/pipelined.py", line 934, in __del__
  File "/home/mat/Workspace/libertem_dev/libertem_fork/src/libertem/executor/pipelined.py", line 910, in close
  File "/home/mat/Workspace/libertem_dev/libertem_fork/src/libertem/common/executor.py", line 490, in put
TypeError: 'NoneType' object is not callable
```

On both machines the call inside `put::490` is to the `cloudpickle` module. My guess is that during teardown the module is already gone by the time we get to the stage of closing down our workers, and as a result we cannot use `cloudpickle` to send the final `SHUTDOWN` message. An option would be a `.put_safe` method on `SimpleMPWorkerQueue` which only supports native python types without `cloudpickle`, and a corresponding second `except:` block in `get` which tries to use the object from the queue if either `cloudpickle` is torn down or it can't decode the object taken from the queue (in which case we assume it's a native python object).

The MWE:

```python
import tempfile
import pathlib
import numpy as np

import libertem.api as lt

if __name__ == '__main__':
    nav_shape = (32, 32)
    sig_shape = (4, 4)
    data = np.ones(nav_shape + sig_shape, dtype=np.float32)

    ctx = lt.Context.make_with('pipelined')

    with tempfile.TemporaryDirectory() as tpath:
        path = pathlib.Path(tpath) / 'test.raw'
        data.tofile(path)

        ds = ctx.load('raw',
                      path=path,
                      nav_shape=nav_shape,
                      sig_shape=sig_shape,
                      dtype=data.dtype)
        ds.set_num_cores(8)

        sum_a = ctx.create_sum_analysis(ds)
        sum_r = ctx.run(sum_a, progress=True)
```

sk1p:
Fixes #1339

It registers an atexit handler with a weak reference to the `Context`, and calls `Context.close` at process exit. Usually, using

    with Context(...) as ctx:
        ...

or explicitly calling `Context.close` for cleanup is preferred, but it is not obvious, and also not consistent with notebook usage, where a context manager is not practical.

Tests TBD - not sure yet how to properly test this. Sketch: maybe with a small function ran via `multiprocessing.Process(target=...)`, and two cases; 1) manual cleanup before interpreter shutdown, 2) automatic `atexit` cleanup. Both should complete within reasonable time limits.

Maybe the `__del__` of `PipelinedExecutor` should be completely removed, as it can lead to these confusing error messages?

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -251,7 +251,10 @@ def default_raw(tmpdir_factory, default_raw_data):
         io_backend=MMapBackend(),
     )
     ds.set_num_cores(2)
-    yield ds
+    try:
+        yield ds
+    finally:
+        lt_ctx.close()
 
 
 @pytest.fixture(scope='session')
--- a/docs/source/changelog/bugfix/context-cleanup-atexit.rst
+++ b/docs/source/changelog/bugfix/context-cleanup-atexit.rst
@@ -1,0 +1,5 @@
+[Bugfix] Context cleanup at process exit
+========================================
+
+* Automatically close :code:`Context` at process exit, so we don't run code when the interpreter is possibly in half-torn-down status (:pr:`1343`).
+
--- a/src/libertem/api.py
+++ b/src/libertem/api.py
@@ -4,6 +4,8 @@ from typing import (
 )
 from typing_extensions import Literal
 import warnings
+import weakref
+import atexit
 
 from opentelemetry import trace
 import numpy as np
@@ -107,6 +109,7 @@ class Context:
             executor = self._create_local_executor()
         self.executor = executor
         self._plot_class = plot_class
+        self._register_at_exit()
 
     @classmethod
     def make_with(cls, executor_spec: ExecutorSpecType, *args, **kwargs) -> 'Context':
@@ -1365,3 +1368,16 @@ class Context:
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.close()
+
+    def _register_at_exit(self):
+        """
+        Register at-exit handler, to make sure the executor is closed
+        """
+        weak_ctx = weakref.ref(self)
+
+        def _exit():
+            if weak_ctx() is None:
+                return
+            weak_ctx().close()
+
+        atexit.register(_exit)
--- a/src/libertem/executor/dask.py
+++ b/src/libertem/executor/dask.py
@@ -522,9 +522,16 @@ class DaskJobExecutor(CommonDaskMixin, BaseJobExecutor):
 
     def close(self):
         if self.is_local:
-            if self.client.cluster is not None:
-                self.client.cluster.close(timeout=30)
+            # Client.close won't close the Cluster itself because
+            # we provided an external dd.SpecCluster
             self.client.close()
+            # Manually close the cluster if not yet torn down
+            # use getattr just in case cluster is already gone
+            if getattr(self.client, 'cluster', None) is not None:
+                self.client.cluster.close(timeout=30)
+        # NOTE: distributed already registers atexit handlers for
+        # both clients and clusters, this is here to allow manual closure
+        # followed by creation of a new Executor without accumulating clusters
 
     @classmethod
     def connect(cls, scheduler_uri, *args, client_kwargs: Optional[dict] = None, **kwargs):
--- a/src/libertem/executor/pipelined.py
+++ b/src/libertem/executor/pipelined.py
@@ -907,6 +907,12 @@ class PipelinedExecutor(BaseJobExecutor):
             self._pool.close_resp_queue()
             self._closed = True
 
+    def __del__(self):
+        # `self` may be already partially garbage-collected; only close
+        # if "enough" of `self` still exists:
+        if hasattr(self, '_closed') and not self._closed:
+            self.close()
+
     def _run_function(self, fn: Callable[..., T], worker_idx, *args, **kwargs) -> T:
         self._validate_worker_state()
         qs = self._pool.get_worker_queues(worker_idx)
--- a/src/libertem/executor/pipelined.py
+++ b/src/libertem/executor/pipelined.py
@@ -907,12 +907,6 @@ class PipelinedExecutor(BaseJobExecutor):
             self._pool.close_resp_queue()
             self._closed = True
 
-    def __del__(self):
-        # `self` may be already partially garbage-collected; only close
-        # if "enough" of `self` still exists:
-        if hasattr(self, '_closed') and not self._closed:
-            self.close()
-
     def _run_function(self, fn: Callable[..., T], worker_idx, *args, **kwargs) -> T:
         self._validate_worker_state()
         qs = self._pool.get_worker_queues(worker_idx)
--- a/tests/executor/test_pipelined.py
+++ b/tests/executor/test_pipelined.py
@@ -465,7 +465,12 @@ def _teardown_manual():
 
 def test_auto_teardown():
     subprocess.run([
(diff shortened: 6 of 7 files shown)


matbryan52 (approved):
From what I understand about how `weakref` works I say :+1: , but I admit I need to experiment a bit more to get a better handle on the gotchas.

sk1p:
> From what I understand about how weakref works I say +1 , but I admit I need to experiment a bit more to get a better handle on the gotchas.

Yeah, I feel this needs to be refined a bit and "tested in production" to see if we are missing any edge cases.

I was proposing:

> Maybe the `__del__` of `PipelinedExecutor` should be completely removed, as it can lead to these confusing error messages?

I'm adding test cases now, and interestingly, removing `__del__` causes a hang again. I'll need to debug this a bit and see if there is anything missing with this approach.

sk1p:
Testing is a little fragile - the LiberTEM-live tests hang after finishing, and py36/py38 run into the timeout condition when automatically cleaning up.

sk1p:
New failures are related to the new pytest release - pytest-benchmark doesn't have a direct dependency on `py`, but imports it, so now that pytest no longer depends on `py`, things are breaking

sk1p:
Tests are now finally green. One remaining question is: do we want to have the `atexit` at `Context`, or at executor level? There can be cases where the `Context` doesn't own the executor, or where we have an executor without a `Context` (if people want to be fancy like that...). @matbryan52 @uellue let me know if you have an opinion on this, otherwise I'd say this is good to go.

matbryan52:
> Tests are now finally green.

Well done!

> One remaining question is: do we want to have the `atexit` at `Context`, or at executor level? There can be cases where the `Context` doesn't own the executor, or where we have an executor without a `Context` (if people want to be fancy like that...). @matbryan52 @uellue let me know if you have an opinion on this, otherwise I'd say this is good to go.

I feel this might be an edge-case on an edge-case, but the safest route to me would be to have a `cleanup_exec` or `own_exec` or `manage_exec` kwarg on `Context.__init__()` which is auto-set True when we use `lt.Context()` or `lt.Context.make_with(...)` (i.e. we 'own' the executor and so we can do what we want to it), and when false we expect the externally supplied executor to be cleaned up / managed otherwise.

matbryan52:
Just giving this a quick test, and in the environment I'm running (3.10.8 / distributed 2022.10.2) calling `DaskJobExecutor.close()` will cause a bunch of tracebacks on normal script / interpreter `exit()`. I assume this was the case before, too, but now we are actually taking the time to call `.close()` before exiting we are seeing the errors. 

```
2022-11-03 16:56:20,587 - distributed.client - ERROR - 
ConnectionRefusedError: [Errno 111] Connection refused

The above exception was the direct cause of the following exception:

  File "/home/mat/.pyenv/versions/libertem_fork/lib/python3.10/site-packages/distributed/comm/core.py", line 291, in connect
    comm = await asyncio.wait_for(
  File "/home/mat/.pyenv/versions/3.10.8/lib/python3.10/asyncio/tasks.py", line 445, in wait_for
    return fut.result()
  File "/home/mat/.pyenv/versions/libertem_fork/lib/python3.10/site-packages/distributed/comm/tcp.py", line 511, in connect
    convert_stream_closed_error(self, e)
  File "/home/mat/.pyenv/versions/libertem_fork/lib/python3.10/site-packages/distributed/comm/tcp.py", line 142, in convert_stream_closed_error
    raise CommClosedError(f"in {obj}: {exc.__class__.__name__}: {exc}") from exc
distributed.comm.core.CommClosedError: in <distributed.comm.tcp.TCPConnector object at 0x7fa91f3c95a0>: ConnectionRefusedError: [Errno 111] Connection refused
...
...
asyncio.exceptions.CancelledError
```

I can reproduce this without LiberTEM just by running:

```python
import dask.distributed as dd

if __name__ == '__main__':
    client = dd.Client()
    client.cluster.close() 
```
so it's something upstream. Looking more into it it looks like `dask.distributed` also registers `atexit` handlers for shutdown, and other people have reported the kind of same errors before, e.g. with this issue: https://github.com/dask/distributed/issues/6087

Still looking around to see if there is a way to handle these errors quietly.

sk1p:
> I can reproduce this without LiberTEM just by running [...]

Does it work to somehow close the client before closing the cluster? Maybe the `close` of `DaskJobExecutor` is just wrong? Looking [at `Client.close`](https://github.com/dask/distributed/blob/aa1c6d858fc5a3e6b6a3276686f469ef25af14eb/distributed/client.py#L1699-L1707) closing the client does imply shutting down the cluster, under certain circumstances, but we are building our own `SpecCluster`, so maybe something like this (pseudo-Python) could work:

```python
cluster = client.cluster
client.close()
cluster.close()
```

matbryan52:
From `dask.distributed` we see they register their own `atexit` handler for all `SpecClusters`:

https://github.com/dask/distributed/blob/1d0701b51c1dc354fb22af9b27b8f722f02daf63/distributed/deploy/spec.py#L671-L677
```
@atexit.register
def close_clusters():
    for cluster in list(SpecCluster._instances):
        if getattr(cluster, "shutdown_on_close", False):
            with suppress(gen.TimeoutError, TimeoutError):
                if getattr(cluster, "status", Status.closed) != Status.closed:
                    cluster.close(timeout=10)
```

So perhaps an option is to remove the `self.client.cluster.close(timeout=30)` from the `DaskJobExecutor.close()`. This solves the problem when closing a script or interpreter using a single `is_local` cluster inside a `Context`.

The one difficulty is if (for some reason) the user wants to close a `Context`and then do other things (including creating other Contexts / Clusters). The way `Client.close()` is written means that our own `dd.SpecCluster` won't be torn down until the interpreter exits - as the client doesn't 'own' the cluster we provide to it. It's possible that the cluster will be torn down by dask when it is gc'd but I can't find an explicit mechanism in the code yet.

sk1p:
@matbryan52 thanks for looking into the dask-related fix! I'll go ahead and merge

PR #1348, opened 2022-11-03
Generate notebooks from tox command

matbryan52:
Adds a command `tox -e notebooks_gen` which regenerates the example notebooks into a folder `examples/generated/`. This is fine for all notebooks except `live-plotting.ipynb` which requires us to use an interactive matplotlib backend (currently `nbagg`). For now I can't think of an easy way to auto-generate this particular notebook. The new notebooks go in `generated/` so that the command can be run multiple times before a 'final' copy + commit of the updated notebooks is made in `examples/`.

The tox config essentially copies `tox -e notebooks` but also installs `pyopencl` and `graphviz` to support `hyperspy-integration.ipynb`. This won't be robust to graphviz install issues (but does work on the server where we have access to the example data).

I also set 
```
    PYTHONWARNINGS=ignore
    DASK_LOGGING__DISTRIBUTED=60
    KMP_WARNINGS=0
```
to suppress a lot of INFO prints from dask and numba. So far this isn't a problem when running the notebook tests but could be in the future.

This was used to generate the notebooks in #1341 .

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ data
 junit.xml
 .vscode-server
 *.sif
+examples/generated/
--- a/tox.ini
+++ b/tox.ini
@@ -206,7 +206,7 @@ extras=
     hdbscan
     bqplot
 commands=
-    jupyter nbconvert --allow-errors --to notebook --output-dir "{toxinidir}/examples/generated/" --execute "{toxinidir}/examples/hyperspy-integration.ipynb"
+    jupyter nbconvert --allow-errors --to notebook --output-dir "{toxinidir}/examples/generated/" --execute "{toxinidir}/examples/*.ipynb"
 passenv=
     TESTDATA_BASE_PATH
     # HyperSpy expects this on Windows
--- a/tox.ini
+++ b/tox.ini
@@ -191,3 +191,27 @@ whitelist_externals=
     cat
 passenv=
     HOME
+
+[testenv:notebooks_gen]
+deps=
+    -rtest_requirements.txt
+    libertem-blobfinder
+    hyperspy
+    pyxem
+    scikit-image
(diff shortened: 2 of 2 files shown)


sk1p (approved):

sk1p:
Thanks, this is really helpful!

> Adds a command `tox -e notebooks_gen` which regenerates the example notebooks into a folder `examples/generated/`. This is fine for all notebooks except `live-plotting.ipynb` which requires us to use an interactive matplotlib backend (currently `nbagg`). For now I can't think of an easy way to auto-generate this particular notebook. 

I think for some notebooks, it's more important that they have good behavior at runtime rather than a nice preview in nbviewer / github etc. Live plotting is such an example, as using matplotlib there is not really fun..

> The new notebooks go in `generated/` so that the command can be run multiple times before a 'final' copy + commit of the updated notebooks is made in `examples/`.


> The tox config essentially copies `tox -e notebooks` but also installs `pyopencl` and `graphviz` to support `hyperspy-integration.ipynb`. This won't be robust to graphviz install issues (but does work on the server where we have access to the example data).

Sounds good! What's actually the reason `pyopencl` and `graphviz` aren't installed in the `notebooks` environment? Are they optional dependencies?

I think the error message from the Python bindings for a missing `graphviz` installation should be good enough.

> I also set
> 
> ```
>     PYTHONWARNINGS=ignore
>     DASK_LOGGING__DISTRIBUTED=60
>     KMP_WARNINGS=0
> ```
>[...]

:+1:

A useful addition to this workflow is to check the difference after generating a new set of notebooks, using `nbdime diff-web examples/<file> examples/generated/<file>`. Gives output like this:

image

(in this case, failing because it was running without data access)

Issue #441, opened 2019-09-25
Allow giving navigation and signal shape for HDF5 files


uellue:
Since HDF5 is a widely used container format, LiberTEM should allow a maximum of flexibility to load and interpret the data.

In particular, it should be possible to "fold" the data in a HDF5 dataset into appropriate navigation and signal dimensions, since not all software saves data with the correct shape.

Particular example: I tried unsuccessfully to open a file that contains data with a flattened navigation dimension.

sk1p:
> Particular example: I tried unsuccessfully to open a file that contains data with a flattened navigation dimension.

Agreed, and that's a valid use-case! Thanks for the issue!

AnandBaburajan:
I'm interested in working on this issue (Project 4: Allow reshaping datasets into a custom shape) as part of GSoC 2020.

sk1p:
See also: https://github.com/LiberTEM/LiberTEM/issues/553

uellue:
As a comment, this is also relevant for the EMPAD dataset. Right now, the `scan_size` parameter is overwritten by information from the XML file if an EMPAD dataset is opened from XML. Users might want to open data with their own `scan_size`, contrary to information in the XML file.

AnandBaburajan:
For `HDF5`, I tried different methods including accessing indices and slices & h5py's `read_direct`. Reading with indices was too slow. Using slices and h5py, I was finished with `nav_shape` reshaping, `sig_shape` reshaping and -ve `sync_offset` (check [this](https://github.com/LiberTEM/LiberTEM/pull/793/commits/b9ef81d3092d980a4eb750b79ff70fe4855f6397) commit). But the solution doesn't look maintainable and `sig_shape` reshaping is slow. So I'll try [memory mapping](https://github.com/LiberTEM/LiberTEM/issues/72) next.

AnandBaburajan:
See also: #984

sk1p:
> See also: #984

Indeed - let me know if you have questions about the changes in that PR - right now, it's still fresh in my mind :)

AnandBaburajan:
> Indeed - let me know if you have questions about the changes in that PR - right now, it's still fresh in my mind :)

[...]

PR #1364, opened 2022-12-05
HDF5 `nav_shape` reshaping and `sync_offset`

matbryan52:
This adds HDF5DataSet `nav_shape` reshaping and `sync_offset`, which respond partially to #441 .

As I mentioned there, because the navigation dimension is always flat in LT (internally) it's quite easy to let the `DataSet` display one shape to the rest of the architecture while treating the array in the file as having a different shape. This code does not touch anything related to choice of partition size, chunking, tileshape etc, essentially it just changes the shape we use when reshaping flat result buffers into final results.

`sync_offset` and reshaping into a different size nav space are also supported by adjusting the positions of partitions relative to the underlying `slice_nd` which access the hdf5 data.

`sig_shape` reshaping will be quite a lot more complicated to implement as we have to constrain tile shapes to be compatible under reshaping (and handle corrections). It might not be so hard when the array on disk has a flat signal dimension as if we constrain tiles to occupy full 'sig-reshaped rows' (in-nD) then any reshape from the flat form to a tile is valid.

~~If this is something we want to include then I would need to add the `nav_shape` option in the web interface too, before merging.~~

## Contributor Checklist:

## Reviewer Checklist:


matbryan52:
> That would be great. Maybe the parameter form component for reshaping then needs to be configured to not show or disable the offset and `sig_shape` parts. Let me know if you need help with that!

Disabling certain inputs is within my TS-fu, but what I can't currently work out is how to update or clear the values of, say, `nav_shape` if I change the selected HDF5 dataset (i.e. in response to changing the dropdown). It would be nice if (where possible) the values update to the defaults for the chosen dataset (or were blanked if not possible), but I can't see a setter (only the setter from field to values that get sent back to Python).

> I think the idea behind sig reshaping is mostly completeness (support for any ND reshape, if possible) and having a symmetric and consistent API.

As an aside, do we (generally) expect to be able to support setting a shape which doesn't match the underlying data size ? e.g. `nav_shape` (10, 10) in an on-disk dataset of 110 frames, with (optionally) a `sync_offset` to choose where the 100 desired frames come from? The implementation here requires the detected nav_size and desired nav_size are identical, and sync offset is applied as normal (as shifting before/beyond the full space).

matbryan52:
Some observations:

- the HDF5DataSet accepts a `min_num_partitions` argument, but I don't think the value is used anywhere in the codebase
- We only support frame-by-frame processing when using an ROI. I don't see why this is the case necessarily (though I'm not volunteering to change it right now! :D )
- With this change I've removed a call to `io.dataset.base.roi._roi_to_nd_indices`, leaving only one call in the whole codebase at `io.dataset.base.coordinates.get_coordinates`

matbryan52:
I think the web client support is now complete, we have the `Reshape`component on the HDF5 page, a `customValidation` function which will reject a `sig_shape` not matching the dataset native sig shape, and the form fields now update with the `ds_path` dropdown.

I haven't enable the sync_offset / frame info display part of the Reshape component, as I think making it update correctly with the dropdown box will require some new plumbing (it's not a Formik field that I can see).

Obviously we'll need to test this manually for the release. Here I generated some dummy files with multiple arrays of various shapes and they all seem to work as expected, but there are a lot of combinations to look through!

@sk1p if you could take a quick look at the TS that would be appreciated!

client/src/dataset/components/HDF5ParamsForm.tsx
```diff
@@ -90,13 +98,27 @@ const HDF5ParamsForm: React.FC<MergedProps> = ({
...
             {warning}
+            <Reshape navShape={values.nav_shape} sigShape={values.sig_shape} syncOffset={values.sync_offset} hideInfo setFieldValue={setFieldValue} setFieldTouched={setFieldTouched} />
             <Button primary type="submit" disabled={isSubmitting}>Load Dataset</Button>
             <Button onClick={onCancel} >Cancel</Button>
             <Button type="button" onClick={handleReset}>Reset</Button>
         </Form>
     )
 }
 
+const getInfoItemForDSPath = (ds_path: string, info?: DatasetInfoHDF5) => {
```
sk1p, line 109: Would be good to have a return type annotation here:

```suggestion
const getInfoItemForDSPath = (ds_path: string, info?: DatasetInfoHDF5): DatasetInfoHDF5Item => {
```

client/src/dataset/components/HDF5ParamsForm.tsx
```diff
@@ -90,13 +98,27 @@ const HDF5ParamsForm: React.FC<MergedProps> = ({
...
+    const dsItemsByPath: {
+        [k: string]: DatasetInfoHDF5Item
+    } = {};
+    info?.datasets?.forEach(dsItem => dsItemsByPath[dsItem.path] = dsItem);
+    // I think this will fail if datasets is undefined or empty
+    // yet the only possible value for ds_path comes from the dropdown
+    // menu which is itself defined by info.datasets so by definition
+    // it is valid!
```
sk1p, line 117: There's actually a case where this is not true: for some files, it can take a very long time to list the contained structure (groups, datasets, ...). Especially for matlab-generated HDF5 files, this can be the case, where often many small items are written to the file. To not block on this kind of file, we implemented a timeout and basically give up listing the datasets. Then, there is no drop-down and you have to manually enter the hdf5 dataset path.

Here is the current handling on the client side for this case:

https://github.com/LiberTEM/LiberTEM/blob/6042a5d3a28b6ec8177a392a5b736eaf76a3406e/client/src/dataset/components/HDF5ParamsForm.tsx#L64-L70

In that case, we can't really validate that the given shape(s) and offset is correct, as we don't have the info on them. As it is an edge case, I guess we can get away with handling errors / validation when the load button in pressed?

client/src/dataset/components/HDF5ParamsForm.tsx
```diff
@@ -25,14 +26,16 @@ const HDF5ParamsForm: React.FC<MergedProps> = ({
...
+    setFieldTouched,
 }) => {
     const dsItemsByPath: {
         [k: string]: DatasetInfoHDF5Item
     } = {};
     info?.datasets?.forEach(dsItem => dsItemsByPath[dsItem.path] = dsItem);
 
     const dsPathOptions = info?.datasets?.map(dsItem => {
-        const shape = dsItem.shape.join(",")
+        const raw_nav_shape = dsItem.raw_nav_shape.join(", ")
```
sk1p, line 37: Usually in TypeScript, the convention is to use `camelCase` for variable names, except of course if you are talking to the Python API.

client/src/dataset/components/HDF5ParamsForm.tsx
```diff
@@ -114,5 +136,15 @@ export default withValidation<DatasetParamsHDF5, DatasetParamsHDF5ForForm, Datas
         sig_shape: parseShapeInCommaSeparatedString(values.sig_shape),
         sync_offset: values.sync_offset,
     }),
+    customValidation: (values, { info }) => {
+        const ds_info = getInfoItemForDSPath(values.ds_path, info)
```
sk1p, line 140: `dsInfo` for keeping with the convention

sk1p (commented):
Looks good overall - there's an edge case if the detection times out, so the validation needs a slight adjustment

client/src/dataset/components/HDF5ParamsForm.tsx
matbryan52: OK, understood. I wish I could generate a test HDF5 file to see how that code path behaves! I guess I could force `isTimeOut`.

I've added some explicit handling of now that we actually use the dataset `info` after the creation of the dropdown/field. The changes now mean that the user's input in the Field is used to key into the info object which can result in `undefined` info items, which we then try to access for things like `nav_shape` (=> uncaught exception). I've checked explicitly for `undefined` but it might be better to refactor the handling of `Field`-type inputs.

client/src/dataset/components/HDF5ParamsForm.tsx
matbryan52: Good spot, I've adjusted a few more names too. What's worse is that I mixed the two in my additions!

client/src/dataset/components/HDF5ParamsForm.tsx
matbryan52: Thanks, I've added that in.

client/src/dataset/components/HDF5ParamsForm.tsx
```diff
@@ -114,5 +142,19 @@ export default withValidation<DatasetParamsHDF5, DatasetParamsHDF5ForForm, Datas
         sig_shape: parseShapeInCommaSeparatedString(values.sig_shape),
         sync_offset: values.sync_offset,
     }),
+    customValidation: (values, { info }) => {
+        const dsInfo = getInfoItemForDSPath(values.ds_path, info)
+        if (dsInfo === undefined) {
```
matbryan52, line 147: @sk1p this is one of the guards on `undefined` dsPath

matbryan52: The other is inside `onDSPathChange`.

client/src/dataset/components/HDF5ParamsForm.tsx
sk1p: OK, looks good!

client/src/dataset/components/HDF5ParamsForm.tsx
sk1p: :+1:

PR #1371, opened 2022-12-21
Explicitly add content of client/ in MANIFEST.in

matbryan52:
Avoids extremely long package build times (and pyroma evaluation time) by explicitly avoiding ever adding `node_modules` to the source dist.

Previously we had these two lines in MANIFEST.in:

```
graft client
prune client/node_modules
```

which during building would first add, then remove node_modules. On some systems this can take up to a minute, and is done even during `pyroma` which means it happens on every commit (with pre-commit enabled).

NOTE: on my system it seems like `package-lock.json` is included in the dist file, which can be quite large (> 1 MB). If we are building from a clean checkout this would not happen but it's something to be aware of (or we can exclude it explicitly).

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,11 +1,12 @@
 graft src
-graft src/libertem/web/client/
+graft src/libertem/web/client
 include LICENSE
 include README.rst
-graft client
-prune client/node_modules
-prune client/build
-prune client/coverage
+graft client/src
+graft client/public
+graft client/types
+include client/*.*js*
+include client/README.md
 graft tests
 include conftest.py
 include pytest.ini


sk1p:
Verified that wheel and sdist still contain the same list of files, thanks!

> NOTE: on my system it seems like `package-lock.json` is included in the dist file, which can be quite large (> 1 MB). If we are building from a clean checkout this would not happen but it's something to be aware of (or we can exclude it explicitly).

Ok, it's also included in the sdist, but as it compresses well it wasn't caught before. I'm fine with keeping it for now.

PR #1377, opened 2023-01-10
Add dataset detect based on file suffix

### LINKED ISSUE #1370: `io.dataset.detect()` first try the dataset type matching the file extension, if possible
In `io.dataset.detect()` we infer the dataset type by iterating over the `filetypes` dictionary  until `cls.detect_params` doesn't raise for that particular definition.

We could instead prioritize the dataset which matches the file extension, if such a match is possible, which would save time when using `ctx.load('auto')` and also when detecting a dataset in the web GUI. 

This could be made to support datasets which have multiple file types, and also one file extension matching multiple dataset types, if at runtime we generate the possible matches to search first, and then fall back to the current behaviour if these don't exist and/or fail.

matbryan52:
Closes #1370 

`io.dataset.detect` now uses the file extension to choose which dataset class to try first, when a compatible extension is present. This does not match multiple file types for a dataset, nor will it try multiple datasets as priority for one file type. To improve this would require refactoring / importing all the dataset classes in advance which could be costly. In reality, only a few datasets support multiple filetypes which don't match their key in `ctx.load`, so there's not a lot to gain from doing this now.

A temporary improvement could be to add aliases to `io.dataset.filetypes`, for example:

```python
   ...
    "mib": "libertem.io.dataset.mib.MIBDataSet",
    "hdr": "libertem.io.dataset.mib.MIBDataSet",
   ...
```

but better would be a proper restructuring of how a dataset registers itself and the extensions it supports.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/docs/source/changelog/misc/detect_with_filetype.rst
+++ b/docs/source/changelog/misc/detect_with_filetype.rst
@@ -4,5 +4,4 @@
 * The function :code:`detect` to automatically determine dataset type
   will now use the file suffix as a hint to choose its search order.
   This may lead to faster responses in the web client when configuring
-  a new dataset. As an additional feature, :code:`MemoryDataSet` will
-  now be auto-detected without error. (:pr:`1377`)
+  a new dataset. (:pr:`1377`)
--- a/docs/source/changelog/misc/detect_with_filetype.rst
+++ b/docs/source/changelog/misc/detect_with_filetype.rst
@@ -1,0 +1,7 @@
+[Misc] Dataset detection uses file suffix if possible
+=====================================================
+
+* The function :code:`detect` to automatically determine dataset type
+  will now use the file suffix as a hint to choose its search order.
+  This may lead to faster responses in the web client when configuring
+  a new dataset.
--- a/docs/source/changelog/misc/detect_with_filetype.rst
+++ b/docs/source/changelog/misc/detect_with_filetype.rst
@@ -4,4 +4,5 @@
 * The function :code:`detect` to automatically determine dataset type
   will now use the file suffix as a hint to choose its search order.
   This may lead to faster responses in the web client when configuring
-  a new dataset.
+  a new dataset. As an additional feature, :code:`MemoryDataSet` will
+  now be auto-detected without error. (:pr:`1377`)
--- a/src/libertem/io/dataset/__init__.py
+++ b/src/libertem/io/dataset/__init__.py
@@ -1,5 +1,5 @@
 import typing
-from typing import Dict, List, Union
+from typing import Dict, List, Union, Optional, Tuple, Any
 import pathlib
 from functools import lru_cache
 import importlib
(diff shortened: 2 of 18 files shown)


sk1p:
> In reality, only a few datasets support multiple filetypes which don't match their key in `ctx.load`, so there's not a lot to gain from doing this now.

What do you think about putting these formats to the front of the detection, such that if there is no direct match, they are tried first? I think the list would be MIB (for .hdr), K2IS (for .gtg), EMPAD (.xml), HDF5 (.h5 and other custom endings I think?), DM (.dm3/.dm4). 

> but better would be a proper restructuring of how a dataset registers itself and the extensions it supports.

Agreed.

On my Laptop, importing all dataset modules takes about 900ms cold, which may be preferable to trying different detect methods. How about something like this:

```python
ext_map = {}
for typ_ in filetypes:
    cls = get_dataset_cls(typ_)
    for ext in cls.get_supported_extensions():
        ext_map[ext] = cls  # if there is overlap, might need to have a list as value here
# look up extension in `ext_map` etc.
```

matbryan52:
> > In reality, only a few datasets support multiple filetypes which don't match their key in `ctx.load`, so there's not a lot to gain from doing this now.
> 
> What do you think about putting these formats to the front of the detection, such that if there is no direct match, they are tried first? I think the list would be MIB (for .hdr), K2IS (for .gtg), EMPAD (.xml), HDF5 (.h5 and other custom endings I think?), DM (.dm3/.dm4).
> 
> > but better would be a proper restructuring of how a dataset registers itself and the extensions it supports.
> 
> Agreed.
> 
> On my Laptop, importing all dataset modules takes about 900ms cold, which may be preferable to trying different detect methods. How about something like this:
> 
> ```python
> ext_map = {}
> for typ_ in filetypes:
>     cls = get_dataset_cls(typ_)
>     for ext in cls.get_supported_extensions():
>         ext_map[ext] = cls  # if there is overlap, might need to have a list as value here
> # look up extension in `ext_map` etc.
> ```

Great, if you are happy with the additional import overhead for now then I'll put that in. One they are imported we can construct an optimal search order.

I also just discovered an unhandled case for both the old and new implementation of detect: `ctx.load('auto', np.ndarray((1, 2, 3, 4)))` just crashes with a MIBDataSet `AttributeError: 'numpy.ndarray' object has no attribute 'lower'`. In an ideal world I'd expect a `MemoryDataSet` (all the more the case with the new sparse support).

sk1p:
> Great, if you are happy with the additional import overhead for now then I'll put that in. One they are imported we can construct an optimal search order.

:+1:

> I also just discovered an unhandled case for both the old and new implementation of detect: `ctx.load('auto', np.zeros((1, 2, 3, 4)))` just crashes with a MIBDataSet `AttributeError: 'numpy.ndarray' object has no attribute 'lower'`. In an ideal world I'd expect a `MemoryDataSet` (all the more the case with the new sparse support).

Opened #1386 for this

matbryan52:
Thanks, I'll handle that and link the issue when I get back to the PR, hopefully today.

matbryan52:
This is a delicate one - the current state works without special-casing `MemoryDataSet` but can't accept `ctx.load('auto', ...)` with multiple positional arguments. I thought this wasn't supported anyway but I then found one test case that covers it `test_frms6::test_auto_open_corrections_posargs`:

```python
ds_corr = lt_ctx.load('auto', FRMS6_TESTDATA_PATH, True, None, None, (2, 3))
```

(seems a bit dangerous to me given the ambiguity of auto!)

If we want to keep that particular behaviour of the interface then it's probably best to just drop the idea of `ctx.load('auto', np.ndarray) -> MemoryDataSet`, which has never worked anyway! The ways I can think of to implement it require either convoluted changes in `MemoryDataSet` kwarg handling or some kind of introspection of how to map the positional args to the `__init__` signature of each dataset (`inspect.Signature.bind(*args, *kwargs)`) - fun but not a good use of time :D

sk1p:
> then found one test case that covers it test_frms6::test_auto_open_corrections_posargs:

Yeah, that comes from the fix for https://github.com/LiberTEM/LiberTEM/issues/936

> If we want to keep that particular behaviour of the interface then it's probably best to just drop the idea of `ctx.load('auto', np.ndarray) -> MemoryDataSet`, which has never worked anyway!

Agreed, let's keep the changes simple for now.

sk1p (approved):

sk1p:
Thanks!

PR #1378, opened 2023-01-11
Reset sync_offset field correctly in web UI

### LINKED ISSUE #1368: Reset button does not reset `sync_offset` field to zero
When using the Reset button in the Web UI, the `sync_offset` form field is not reset visually to zero, even though the underlying parameter which is sent to the dataset is. This can result in loading a dataset expecting a non-zero `sync_offset` but actually getting the behaviour of `sync_offset==0`.

matbryan52:
Closes #1368 

Simple fix, I believe this was broken because the value passed to the sync_offset `<Input>` was the React state variable rather than the Formik value, and as a result Formik wasn't aware of the value to reset.

## Contributor Checklist:

## Reviewer Checklist:


sk1p (changes requested):
This change has an issue, namely you can't enter negative sync offsets via the keyboard anymore (only with the small arrows of the input element). The reason is probably that now it's not the stringly version that is in the `value` of the field, but the parsed one, and parsing `"-"` doesn't yield a value...

matbryan52:
> This change has an issue, namely you can't enter negative sync offsets via the keyboard anymore (only with the small arrows of the input element). The reason is probably that now it's not the stringly version that is in the `value` of the field, but the parsed one, and parsing `"-"` doesn't yield a value...

Good spot, I hadn't tried keyboard input, only the arrows.

That said, I've just spent a good while looking at this and I don't really know how to fix it. I think Formik is already successfully resetting `values.syncOffset`, but the actual displayed value is from `offsetValue` defined in the call to `react.useState`. I thought about wrapping `handleReset` from Formik with something like:

```javascript
const handleResetWrap = () => {
   handleReset();
   setOffset(values.syncOffset.toString());
}

<Button type="button" onClick={handleResetWrap}>Reset</Button>
```

But to do this will (I think) require us to declare `const [offsetValue, setOffset] = ...` outside of the `Reshape` component, and copy this in in every dataset. 

There might also be the possibility to add to the `commonParams.onReset` callback in `DatasetOpen.tsx` and make the `offsetValue` a piece of top-level state, and pass it down into `Reshape` ?

matbryan52:
Sorry for the review request, not exactly what I wanted to do.

Anyway I think it's all good now, seems like the extra State was redundant ? I replaced it with the Formik-provided values.syncOffset and seems like both the input box and the sync_offset info box below both work as intended.

That said I don't know much about *why* exactly this works, I guess that the values are already somewhere state-ful and so correctly trigger re-rendering of the elements when changed or reset.

sk1p (approved):
Looks good to me!

sk1p:
> Sorry for the review request, not exactly what I wanted to do.

No worries!

> Anyway I think it's all good now, seems like the extra State was redundant ? I replaced it with the Formik-provided values.syncOffset and seems like both the input box and the sync_offset info box below both work as intended.
> 
> That said I don't know much about _why_ exactly this works, I guess that the values are already somewhere state-ful and so correctly trigger re-rendering of the elements when changed or reset.

Yeah - as I understand it, the extra state was only used to account for "invalid" states that don't parse as integers - so you had one state in the form library that was always the parsed value, the other one always being the raw string value. As we saw with the reset interaction, this leads to issues...

Issue #1366, opened 2022-12-09
How to get data out and other stupid questions


maclariz:
Sorry for being a complete idiot, but I need to be able to get 4DSTEM data out of libertem for processing in other applications.

I found that libertem reads the latest version of mib, which is handy.  And can get it to load my data, apparently in the correct shape.

But from here I need two things:

Quick and dirty visualisation (e.g. sum image, ADF image).  Cannot even get my head round how you do this after a frustrating hour in your documentation.  Everything well documented is way more complex than this, and unnecessarily complex for me.

Export to a numpy array or an hdf5 file that I can read with something else.

Can anyone help?

Ian

PR #1379, opened 2023-01-11
`Context.export_dataset` method to export to another file format

matbryan52:
**NOTE** The implementation has changed, leaving the below for the record...

Supports export to `npy` and `raw`, could be later extended. In particular in response to #1366 and likely other Issues.

User-facing is fairly minimal:

```python
class DataSet:
    ...

    def save_as(
        self,
        path: os.PathLike,
        format: Optional[Literal['raw', 'npy']] = None,
        progress: bool = False,
        save_dtype: Optional['nt.DTypeLike'] = None,
    ):
        ...
```

Works via `partition.get_macrotile()` so should be quite well-supported. Writes data serially partition-by-partition but this could of course be improved.

I did see some old code for writing data notably `io.writers.base.WriteHandle`, but I couldn't quite make sense of it. Instead I wrote the writer support into the file `common.writers`.

A further quality-of-life extension would be a method to actually load the data direct to an array in memory.

I'm expecting to discuss / re-work this before merging, of course, but I'm putting up the PR for discussion!

## Contributor Checklist:

## Reviewer Checklist:


uellue:
Thx for addressing this! :-)

Maybe one could also use the Dask integration? That already allows loading to memory, storing into anything that allows slice assignment, HDF5, TileDB, Zarr, NPY stacks: https://docs.dask.org/en/stable/array-creation.html#store-dask-arrays. Probably it just has to be documented properly?

Also, in LiberTEM-live we have https://github.com/LiberTEM/LiberTEM-live/blob/master/src/libertem_live/udf/record.py that performs parallel writing to an NPY file on the workers using an UDF. That approach could be the way to go for best performance. Instead of NPY one could also look into ZARR -- it just has to support parallel assignment to non-overlapping slices.

matbryan52:
> Maybe one could also use the Dask integration? That already allows loading to memory, storing into anything that allows slice assignment, HDF5, TileDB, Zarr, NPY stacks: https://docs.dask.org/en/stable/array-creation.html#store-dask-arrays. Probably it just has to be documented properly?

Dask integration is a good way to do this, I agree. The reason that I didn't go with that for this first implementation (even in a hidden way) is that right now we can't ensure that partitions are axis-aligned so saving via Dask could incur rechunking (or even loading all the data into memory). It is a good reason to revisit #1264 and do it properly.

It's also possible that the delayed functions will be executed on different workers / machines leading IPC for the array chunks, though this is probably a minor issue.

> Also, in LiberTEM-live we have https://github.com/LiberTEM/LiberTEM-live/blob/master/src/libertem_live/udf/record.py that performs parallel writing to an NPY file on the workers using an UDF. That approach could be the way to go for best performance. Instead of NPY one could also look into ZARR -- it just has to support parallel assignment to non-overlapping slices.

Great, I can look at implementing this approach, though we would have to upstream RecordUDF or similar rather than importing from `libertem-live` I think.

uellue:
Good point about the K2 IS and dedicated high-performance low-overhead writers! In that sense, it would be a dataset option in an UDF run and just saving or converting would be running an empty UDF set with the dataset saving option enabled?

matbryan52:
After discussion with @sk1p updated this to instead use a method on the Context:

https://github.com/LiberTEM/LiberTEM/blob/ea7ee918913f85c409f2b879ec3a1a82e348960e/src/libertem/api.py#L1516C1-L1526

with an at-this-time minimal interface. Also this PR now upstreams `RecordUDF` to take advantage of its performance.

To be discussed if this is something we want to add, but it is undeniably a very convenient way to work with some odd datasets (offline) with other tools !

src/libertem/api.py
```diff
@@ -1510,3 +1512,53 @@ def _exit():
             weak_ctx().close()
 
         atexit.register(_exit)
+
+    def convert_dataset(
```
sk1p, line 1516: Maybe: `export`, `export_dataset`

src/libertem/api.py
```diff
@@ -1510,3 +1512,53 @@ def _exit():
...
+        self,
+        dataset: DataSet,
+        path: os.PathLike,
+        progress: bool = False,
```
sk1p, line 1520: Discussion: `path`, `progress` could be keyword-only arguments

src/libertem/api.py
```diff
@@ -1510,3 +1512,53 @@ def _exit():
...
+        progress : bool, optional
+            Whether to display a progress bar for the export,
+            by default False
+
+        Raises
+        ------
+        ValueError : If the path suffix is not supported
+        """
+        path = pathlib.Path(path)
+        if path.suffix != '.npy':
```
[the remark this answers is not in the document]
matbryan52: Definitely, I chose not to add the format parameter at this stage as we don't support anything else.

src/libertem/api.py
matbryan52: +1

src/libertem/api.py
matbryan52: Changed to export_dataset for now!

src/libertem/api.py
```diff
@@ -1510,3 +1512,53 @@ def _exit():
...
+        dataset: DataSet,
+        path: os.PathLike,
+        progress: bool = False,
+    ):
+        """
+        Convert the dataset to another format on disk
+
+        At this time can only convert to numpy *.npy* format,
+        but future extensions are possible.
+
```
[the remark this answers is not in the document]
matbryan52: Added `overwrite=False` by default

sk1p (commented):
Looks good overall, I've added some comments in-line, mostly details.

src/libertem/api.py
```diff
@@ -1510,3 +1512,65 @@ def _exit():
...
+    ):
+        """
+        Export the dataset to another format on disk
+
+        At this time can only convert to numpy *.npy* format,
+        but future extensions are possible.
+
+        The written data will have any reshaping / sync_offset
+        properties of the DataSet effectively baked into
+        the new file.
```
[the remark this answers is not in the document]
matbryan52: I think in the short term it's not something that'll be missed and better to not commit to the interface, in the future we might not go via `run_udf` and in this case we'd be tied to including corrections! If someone asks for it we can easily add it.

matbryan52:
Thanks for the review again, sorry to have missed all those doc changes!

sk1p (approved):
LGTM!

PR #1381, opened 2023-01-12
Flaky progress bar test fix

matbryan52:
Responds to [ db48c49 ](https://github.com/LiberTEM/LiberTEM/commit/db48c497353af9debf2b3f2dbe77ec71ba34d3f6)

This is still a draft as I need to come up with a better test case, but putting this up to explain why the original test was flaky.

After looking at this in detail, I think the issue is twofold:

1. Partition Start + Tile Completion messages go via the Queue and are processed in the background thread, while Partition End messages are fully synchronous and are processed immediately after we get results from the task. (This is the fallback which means all partitions will be accounted for even if for some reason the comms break down). The test case does no real work, so occasionally a partition will complete in the main thread before we receive all of its start + intermediate messages in the background.

2. The test is written in a way which doesn't account for point 1, as it is checking that at a given point in time we had a precise number of partitions in flight and frames processed. If a partition was marked as complete before we get its intermediate messages => the intermediate messages are dropped and the sequence of states is different. I don't think any messages were ever 'lost' in this test, rather they were received and dropped or received later than a partition end message causing the state sequence to be different.

I will re-think the test to actually check the messages we received rather than the states after each message.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/common/progress.py
+++ b/src/libertem/common/progress.py
@@ -210,8 +210,8 @@ class ProgressManager:
         # For converting tiles to pseudo-frames
         self._sig_size = tasks[0].partition.shape.sig.size
         # Counters for part display
-        self._num_complete = 0
-        self._num_in_progress = 0
+        self._complete = set()
+        self._in_progress = set()
         self._num_total = len(self._counters)
         if reporter is None:
             reporter = TQDMProgressReporter()
@@ -231,8 +231,8 @@ class ProgressManager:
         return ProgressState(
             sum(self._counters.values()),
             self._total_frames,
-            self._num_complete,
-            self._num_in_progress,
+            len(self._complete),
+            len(self._in_progress),
             self._num_total,
         )
 
@@ -271,7 +271,13 @@ class ProgressManager:
         """
         if topic != 'partition_start':
             raise RuntimeError('Unrecognized topic')
-        self._num_in_progress += 1
+        t_id = message['ident']
+        if t_id not in self._complete:
+            # if not complete handles case task was completed
+            # before we can process its start message,
+            # completion is signalled in the main thread
+            # while start messages are processed in the background
+            self._in_progress.add(t_id)
         self.reporter.update(self.state)
 
     def handle_end_task(self, topic: str, message: Dict[str, Any]):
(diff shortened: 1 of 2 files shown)


matbryan52:
> `check_notebooks` failures look like changes from current matplotlib and/or jupyter releases, see also [#1207 (comment)](https://github.com/LiberTEM/LiberTEM/pull/1207#issuecomment-1380338776) (should be fixed in separate PR so we don't get a mess)

No worries, this test is currently skipped on master I think so there is no rush to merge it.

uellue:
Thank you, looks good to me! I'll merge after the tests pass with the new notebooks.

uellue:
...also, fix confirmed with a few test runs on my machine.

uellue:
Now I am confused. Why did the notebook tests succeed for the previous PR, but here it complains that it doesn't know the "metric" parameter? I changed that in the notebooks since "affinity" triggered a deprecation warning suggesting to use "metric". Did the runs use different environments, with this one using an older version?

matbryan52:
> Now I am confused. Why did the notebook tests succeed for the previous PR, but here it complains that it doesn't know the "metric" parameter? I changed that in the notebooks since "affinity" triggered a deprecation warning suggesting to use "metric". Did the runs use different environments, with this one using an older version?

I am guessing it's using a cached environment ?? 

We could perhaps add a Pipeline step to print out `pip freeze` before testing so we have a full listing of the installed packages during a test?

uellue:
Ok, I'm going to merge anyway since the failure doesn't seem to be related to progress bars.

uellue:
> We could perhaps add a Pipeline step to print out `pip freeze` before testing so we have a full listing of the installed packages during a test?

Good idea, I'm trying it in 52e28b6!

sk1p:
Exactly, the recreation doesn't work properly since we nowadays have multiple pipeline agents, and the `tox -r` for one environment isn't guaranteed to run on all agents.

> We could perhaps add a Pipeline step to print out `pip freeze` before testing so we have a full listing of the installed packages during a test?

:+1:

We could also always re-create the tox environments, and trade some additional CI time for a guaranteed up-to-date state.

PR #1382, opened 2023-01-12
Fix notebooks

uellue:
* Mitigation for https://github.com/dask/distributed/issues/7471 picked from #1207
* Rerun notebooks and remove ";". Somehow the notebook tests expect an output despite the ";". I like it better without ";" anyway.

## Contributor Checklist:

## Reviewer Checklist:


matbryan52:
";" in the notebooks is me, sorry! It seemed like we were catching and filtering output using `nbval_sanitize` when adding the ; would suppress the matplotlib stuff Strange that the notebook tests still expect output...

uellue:
> ";" in the notebooks is me, sorry! It seemed like we were catching and filtering output using `nbval_sanitize` when adding the ; would suppress the matplotlib stuff at the source. Strange that the notebook tests still expect output...

Yes, that looks like a bug in the test routines to expect an output there. Probably we were dealing with "opposing bugs" here that first require the ";" and then break with the ";". I'll work on a reproducer and report it to nbval. :-)

uellue:
https://github.com/computationalmodelling/nbval/pull/195

uellue:
I'll just go ahead and merge so that I can also merge #1381 without a guaranteed build failure. :-)

PR #1392, opened 2023-03-14
Comudf

uellue:
Create a convenient UDF for COM so that users don't have to juggle mask factories, plot functions etc for live processing.

Currently WIP: Behavior in connection with ROIs since np.gradient may produce values that are not NaN for pixels outside a ROI if all neighbors are within the ROI. To be discussed how to handle this. Currently the analysis behavior can only be reproduced in a UDF if result buffer type "single" is used so that values outside the ROI can be set.

## TODO

- [ ] Update documentation / notebooks
   - [ ] Using the rotation guess function
   - [ ] Applying rotation etc.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/analysis/com.py
+++ b/src/libertem/analysis/com.py
@@ -1,6 +1,6 @@
 import logging
 import inspect
-from typing import Dict, NamedTuple, Optional, Tuple, Type, Union, TYPE_CHECKING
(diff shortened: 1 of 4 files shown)


sk1p:
Work is continuing in #1458

PR #1397, opened 2023-03-21
Make sure coordinates are available in `get_task_data`

sk1p:
As an alternative to #1210, this keeps the option open to have different partition slicing schemes in the future, that cannot be modeled by a simple `Slice` object. For most applications, it should be sufficient to have the coordinate list available for the partition.

This does not yet add support for accessing coordinates in `preprocess`, as that is also called per-dataset, and the coordinate code would need substantial changes to support this.

## Contributor Checklist:

## Reviewer Checklist:


uellue (commented):
LGTM!

PR #1398, opened 2023-03-29
Add endpoint to set progress reporter

matbryan52:
This is a component of the progress bar PR ( #1341 ) which probably should have been included then. We added plumbing for the `UDFRunner` to inject other types of `ProgressReporter` (the object which actually displays the progress messages), but didn't expose any user-facing endpoints. This is because `UDFRunner` is implicitly instantiated by `ctx::_run_sync_wrap` with only the `udfs` argument. You can get around it by subclassing the particular Executor + UDFRunner you're using but it's heavy!

This commit is the simplest possible suggestion to solve this, we essentially leave it to the user to ensure that whatever they pass into Context will forever be viable as a way to display progress, either as an object which can be re-used or a class which can be forever re-instantiated by the context. A more robust approach might be a way to re-direct the progress messages only for the duration of `run`/`run_udf` (expose the kwarg everywhere it needs to be exposed).

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/api.py
+++ b/src/libertem/api.py
@@ -1385,9 +1385,9 @@ class Context:
         self.executor.close()
 
     def set_progress_reporter(
-            self,
-            reporter: Union['ProgressReporter', Type['ProgressReporter']]
-        ):
+                self,
+                reporter: Union['ProgressReporter', Type['ProgressReporter']]
+            ):
         self._progress_reporter = reporter
 
     def __enter__(self):
--- a/src/libertem/api.py
+++ b/src/libertem/api.py
@@ -1,6 +1,6 @@
 from typing import (
     TYPE_CHECKING, Any, List, Optional, Union, Iterable, Generator, Coroutine,
-    AsyncGenerator, overload, Tuple, Type
+    AsyncGenerator, overload, Tuple
 )
 from typing_extensions import Literal
 import warnings
@@ -21,6 +21,7 @@ from libertem.executor.dask import DaskJobExecutor
 from libertem.executor.delayed import DelayedJobExecutor
 from libertem.executor.integration import get_dask_integration_executor
 from libertem.common.executor import JobExecutor
+from libertem.common.progress import ProgressReporter
 from libertem.masks import MaskFactoriesType
 from libertem.analysis.raw import PickFrameAnalysis
 from libertem.analysis.com import COMAnalysis
@@ -40,7 +41,6 @@ if TYPE_CHECKING:
     import numpy.typing as nt
     from sparse import SparseArray
     from scipy.sparse import spmatrix
-    from libertem.common.progress import ProgressReporter
 
 tracer = trace.get_tracer(__name__)
 
@@ -115,7 +115,6 @@ class Context:
             executor = self._create_local_executor()
         self.executor = executor
         self._plot_class = plot_class
-        self._progress_reporter = None
         self._register_at_exit()
 
     @classmethod
@@ -609,7 +608,7 @@ class Context:
     def run(
         self, job: Analysis,
         roi: RoiT = None,
-        progress: bool = False,
+        progress: Union[bool, ProgressReporter] = False,
         corrections: Optional[CorrectionSet] = None,
     ) -> Union[np.ndarray, AnalysisResultSet]:
         """
@@ -670,7 +669,7 @@ class Context:
         udf: Union[UDF, Iterable[UDF]],
         roi: RoiT = None,
         corrections: Optional[CorrectionSet] = None,
-        progress: bool = False,
+        progress: Union[bool, ProgressReporter] = False,
         backends=None,
         plots=None,
         sync=True,
@@ -802,7 +801,7 @@ class Context:
         udf: Union[UDF, Iterable[UDF]],
         roi: RoiT = None,
         corrections: CorrectionSet = None,
-        progress: bool = False,
+        progress: Union[bool, ProgressReporter] = False,
         backends=None,
         plots=None,
         sync=True,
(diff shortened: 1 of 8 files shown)


matbryan52:
To me this is a very similar property to `_plot_class`, so wherever it ends up being moved could also be the place to define `_progress_reporter`.

matbryan52:
In my case I needed this because I wanted to display a persistent, re-usable progress bar from Panel, which is entirely possible using the `ProgressReporter` definition except I couldn't inject it where it was needed.

sk1p:
> To me this is a very similar property to `_plot_class`, so wherever it ends up being moved could also be the place to define `_progress_reporter`.

Agreed.

> In my case I needed this because I wanted to display a persistent, re-usable progress bar from Panel, which is entirely possible using the `ProgressReporter` interface except I couldn't inject it where it was needed.

Understood. One potential issue is that you can have multiple UDFs running concurrently, started from the same `Context` and thus using the same reporter instance. This is possible and allowed either using threads or using `async` code. In practice, if I refactor the web backend code to use `Context` internally, instead of directly using the `UDFRunner` interface, this would be something that happens in practice, too.

Is this an issue?

I don't remember, does the progress infrastructure associate the events with the cancellation ID that is generated in `run_for_dataset_sync`? Should we care about this case, or is it fixable later without changing external interfaces?

matbryan52:
>  One potential issue is that you can have multiple UDFs running concurrently, started from the same `Context` and thus using the same reporter instance.
> ...
> Is this an issue?

Good point! And yes it would certainly be an issue in the case of a single reporter instance, if the usage might mean multiple concurrent calls to `run_` (easy enough to avoid in a UI design...).

> I don't remember, does the progress infrastructure associate the events with the cancellation ID that is generated in `run_for_dataset_sync`? Should we care about this case, or is it fixable later without changing external interfaces?

I've pushed a change that provides the cancel_id in each message, so a properly defined Reporter could keep track of multiple in-flight runs.

matbryan52:
I had another thought about how to implement the API for this:

```python
def run_udf(..., progress: bool | ProgressReporter):
    ...
```

It makes the kwarg more complicated but it has the advantage that there is no persistent modification to the Context - by design it's restricted to one run, and there is no additional method to add. The user is also responsible for instantiating their ProgressReporter rather than us doing it for them with no arguments (which would be the current behaviour).

src/libertem/api.py
```diff
@@ -1379,6 +1384,12 @@ def _create_local_executor(self):
     def close(self):
         self.executor.close()
 
+    def set_progress_reporter(
+                self,
+                reporter: Union['ProgressReporter', Type['ProgressReporter']]
+            ):
+        self._progress_reporter = reporter
```
sk1p, line 1391: As this is a method on the `Context`, which is the API most users will be using, can we have a minimal docstring here? I think it's fine to describe that it is an advanced use-case meant for integration, and maybe that the usual way to get a progress bar is to pass `progress=True` to `Context.run_udf`, to redirect users that are lost and believe this is for the common use case.

Stylistic suggestion for consistency: in such a multi-line argument list scenario, I try to keep the closing parenthesis at the same indentation level as the `def`, something like this:

```suggestion
    def set_progress_reporter(
        self,
        reporter: Union['ProgressReporter', Type['ProgressReporter']]
    ):
        self._progress_reporter = reporter
```

(at some point, we should switch to `black` for code formatting I guess)

sk1p:
> I had another thought about how to implement the API for this:
> 
> ```python
> def run_udf(..., progress: bool | ProgressReporter = False):
>     ...
> ```

I thought about this, too. This would also mean we don't have the issue of needing to track which progress "item" came from where, as you can pass a new instance to each `run_udf` call. For integration this is probably the right design, as most cases will have a central `run_udf` call where this can be handled. If this were something users set for each `run_udf` call, I'd prefer the `set_progress_reporter` way.

> It makes the `kwarg` more complicated but it has the advantage that there is no persistent modification to the Context - by design it's restricted to one run, and there is no additional method to add. The user is also responsible for instantiating their `ProgressReporter` rather than us doing it for them with no arguments (which would be the current behaviour).

Yup, especially when integrating into libraries/apps, this is nice behavior.

> In a similar way I feel like `_plot_class` could work like this too - i.e. it is inferred from the `plots=` argument rather than registered in a permanent way?

I think the current design for plotting is the way it is for the use-case of notebooks with many manual `run_udf` calls - it can become tedious to repeat the plot class again and again. Or if you are in an environment where the "normal" plot classes based on matplotlib/bqplot don't work or aren't integrated well, you again must specify this for every `run_udf` call. GMS is such an example, and we probably will have more in the future.

src/libertem/api.py
matbryan52: I was in the process of writing the docstring when I decided to write the comment for `(..., progress: bool | ProgressReporter)`. Do you want to retain both approaches then?

sk1p: > I was in the process of writing the docstring when I decided to write the comment for `(..., progress: bool | ProgressReporter)`. Do you want to retain both approaches then?

Yeah, I wrote this comment before I saw yours. I'm fine with having only one approach, i.e. `run_udf`.

tests/common/test_progress.py
```diff
@@ -127,11 +127,12 @@ def test_progress_inline_fasttime(lt_ctx, monkeypatch):
     # through a call to finalize_task() on the main thread
     # This tests the fallback mechanism in case comms fail
     states = reporter._history
-    start_progress = ProgressState(0., 16, 0, 0, 2)
+    progress_id = states[0].progress_id
```
sk1p, line 130: It would be nice to add a test here (or in the general `Context` tests) that specifies a reporter via the `Context` API and checks that this reporter is actually used.

(sorry that my review is coming in bit-by-bit today, some things going on concurrently...)

matbryan52: The new API is now used in all the tests but one, and the injected reporter is used.

sk1p (approved):

PR #1401, opened 2023-04-07
Single file DM4 support (C-order only)

matbryan52:
Extracted from #1328 for only the C-ordered file support. Raises `DataSetException` if we don't find the c-order metadata flag, unless the `force_c_order` constructor argument is set.

Needs test cases on real data, which are available on Zenodo, I am thinking probably this one: https://zenodo.org/record/4307783 for a small c-ordered 4D-STEM file. There are more data links in this post: https://github.com/LiberTEM/LiberTEM/pull/1328#issuecomment-1284104419 , I still haven't found a public 4D .dm4 in F/C-ordering, only the 3D STEM-EELS file I linked in that comment - including this could at least be a way to test that we raise if we don't see the right flag. @sk1p how would I go about adding new test data?

## Contributor Checklist:

## Reviewer Checklist:


src/libertem/io/dataset/dm_single.py
```diff
@@ -0,0 +1,370 @@
...
+from libertem.common.math import prod
+from libertem.common import Shape
+from .base import BasePartition, DataSetException, DataSetMeta, File, IOBackend
+from .dm import DMDataSet, SingleDMDatasetParams, DMFileSet
+
+log = logging.getLogger(__name__)
+
+if typing.TYPE_CHECKING:
+    from numpy import typing as nt
+    from libertem.executor.base import JobExecutor
```
sk1p, line 19:
```suggestion
    from libertem.common.executor import JobExecutor
```

(to try to keep things license-clean)

sk1p (approved):
Overall, looks good to me!

One question, should we expose the parameters (`force_c_order`, `dataset_index`) to the GUI? `dataset_index` selection could work similarly to HDF5. Maybe better to postpone to after 0.11?

matbryan52:
> Overall, looks good to me!
> 
> One question, should we expose the parameters (`force_c_order`, `dataset_index`) to the GUI? `dataset_index` selection could work similarly to HDF5. Maybe better to postpone to after 0.11?

Agreed that ideally these would be in the UI too. `force_c_order` I can add quite quickly but dataset selection I think is more work, better for after 0.11. (I also think multi-dataset 4D-STEM .dm4 files are quite rare...)

matbryan52:
I think the test failures on the old DM dataset are related to this:

https://github.com/LiberTEM/LiberTEM/blob/8ac0e0b584e344ec01f3042ad3d85b688f1b9f95/tests/io/datasets/test_dm.py#L33

The new .dm4 file is being recognized and included in the default dataset. Perhaps best if we move the files that are meant to form a stack into a subfolder?

sk1p:
> I think the test failures on the old DM dataset are related to this:
> 
> https://github.com/LiberTEM/LiberTEM/blob/8ac0e0b584e344ec01f3042ad3d85b688f1b9f95/tests/io/datasets/test_dm.py#L33

Ah, right.

> The new .dm4 file is being recognized and included in the default dataset. Perhaps best if we move the files that are meant to form a stack into a subfolder?

Yeah, or as an easier fix, make the glob more specific, something like this maybe?

```python
 files = list(sorted(glob(os.path.join(DM_TESTDATA_PATH, '2018-7-17*.dm4')))) 
```

matbryan52:
> Yeah, or as an easier fix, make the glob more specific, something like this maybe?
> 
> ```python
>  files = list(sorted(glob(os.path.join(DM_TESTDATA_PATH, '2018-7-17*.dm4')))) 
> ```

All done. I went with your easier fix (though in the end there were a lot of instances of this line to change)!

If I could point to a particular flaw right now is that the web interface doesn't handle incompatible DM files very well - it just rejects creation of the dataset and continues auto-detecting as other types (from which I saw #1403 ), without explaining why this obviously-a-DM-file was rejected. A user might find that frustrating as there is no error message. That said the behaviour is not unique to this dataset - if for some reason my .hdr is corrupt my MIB dataset would be silently rejected too...

sk1p:
> All done. I went with your easier fix (though in the end there were a lot of instances of this line to change)!

:+1:

> If I could point to a particular flaw right now is that the web interface doesn't handle incompatible DM files very well - it just rejects creation of the dataset and continues auto-detecting as other types (from which I saw #1403 ), without explaining why this obviously-a-DM-file was rejected. A user might find that frustrating as there is no error message. That said the behaviour is not unique to this dataset - if for some reason my .hdr is corrupt my MIB dataset would be silently rejected too...

Yes, indeed - we may want to add an information header at the top of the "parameters for opening" dialog, which tells in a human-readable way why this dialog was opened - was it saved parameters from a previous open attempt, or a successful detection, or a fallback to the default (i.e. raw)?

sk1p (approved):
Looking good now, thanks!

PR #1404, opened 2023-04-14
Add guard against too-large files in RawCSRDataSet detect_params

### LINKED ISSUE #1403: RawCSR `load_toml` too much worker memory when auto-detecting
I just ran into a strange case when the web interface was running auto-detect on a DM4 file, seemingly the `load_toml` run_function in the RawCSR dataset overran the worker memory budget:

image

Could it be that it is trying to interpret a large binary file as TOML ?

matbryan52:
Closes #1403 

Guards against by accepting either files with a supported extension, or files less than 1 MB in size with any extension.

~~Need to add a regression test case.~~

## Contributor Checklist:

* [NA] I have added/updated documentation for all user-facing changes
* [NA] I have included the [rebuilt production build of the client](https://libertem.github.io/LiberTEM/contributing.html?#building-the-client) (only if changes were made to the GUI)

## Reviewer Checklist:

### DIFF
--- a/src/libertem/io/dataset/raw_csr.py
+++ b/src/libertem/io/dataset/raw_csr.py
@@ -232,7 +232,7 @@ class RawCSRDataSet(DataSet):
         try:
             _, extension = os.path.splitext(path)
             has_extension = extension.lstrip('.') in cls.get_supported_extensions()
-            under_size_lim = os.stat(path).st_size < 2**20  # 1 MB
+            under_size_lim = executor.run_function(os.stat, path).st_size < 2**20  # 1 MB
             if not (has_extension or under_size_lim):
                 return False
             conf = executor.run_function(load_toml, path)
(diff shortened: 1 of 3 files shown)


src/libertem/io/dataset/raw_csr.py
```diff
@@ -230,6 +230,11 @@ def supports_correction(self):
     @classmethod
     def detect_params(cls, path: str, executor: "JobExecutor"):
         try:
+            _, extension = os.path.splitext(path)
+            has_extension = extension.lstrip('.') in cls.get_supported_extensions()
+            under_size_lim = os.stat(path).st_size < 2**20  # 1 MB
```
sk1p, line 235: For the remote use-case, need to do fs access via the executor:

```suggestion
            under_size_lim = executor.run_function(os.stat, path).st_size < 2**20  # 1 MB
```

sk1p (changes requested):
Good catch, and thanks for the fix! See comments for a small issue, otherwise looks good.

src/libertem/io/dataset/raw_csr.py
matbryan52: Good spot, thanks.

uellue:
LGTM, thx, merging!

PR #1422, opened 2023-04-21
More docs improvements (for after release)

matbryan52:
Continuing with docs changes

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/io/dataset/empad.py
+++ b/src/libertem/io/dataset/empad.py
@@ -127,7 +127,7 @@ class EMPADDataSet(DataSet):
     Examples
     --------
 
-    >>> ds = ctx.load("empad", path='./path_to_file.xml', ...)
+    >>> ds = ctx.load("empad", path='./path_to_file.xml', ...)  # doctest: +SKIP
     """
     def __init__(self, path, scan_size=None, nav_shape=None,
                  sig_shape=None, sync_offset=0, io_backend=None):
--- a/src/libertem/io/dataset/empad.py
+++ b/src/libertem/io/dataset/empad.py
@@ -123,6 +123,11 @@ class EMPADDataSet(DataSet):
     sync_offset: int, optional
         If positive, number of frames to skip from start
         If negative, number of blank frames to insert at start
+
+    Examples
+    --------
+
+    >>> ds = ctx.load("empad", path='./path_to_file.xml', ...)
     """
     def __init__(self, path, scan_size=None, nav_shape=None,
                  sig_shape=None, sync_offset=0, io_backend=None):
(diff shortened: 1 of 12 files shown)


src/libertem/io/dataset/empad.py
```diff
@@ -123,6 +123,11 @@ class EMPADDataSet(DataSet):
     sync_offset: int, optional
         If positive, number of frames to skip from start
         If negative, number of blank frames to insert at start
+
+    Examples
+    --------
+
+    >>> ds = ctx.load("empad", path='./path_to_file.xml', ...)
```
sk1p, line 130: This either needs a `# doctest: +SKIP`, or better: an `autouse=True` fixture that references the real EMPAD test data. Something like we have here already: https://github.com/LiberTEM/LiberTEM/blob/f5a2417e549384b9cf1d371916540c1beeac2e03/conftest.py#L810-L813

This one makes sure this doctest works:

https://github.com/LiberTEM/LiberTEM/blob/f5a2417e549384b9cf1d371916540c1beeac2e03/src/libertem/io/dataset/raw.py#L74-L75

Could be an additional fixture, that calls `pytest.skip(...)` if the test data is not available.

Same probably applies to the other added examples.

matbryan52: Yeah, I figured I would have to look at this more closely. The reason for adding them is that the documentation for the keys listed in `io.datasets.__init__.py` is quite hard to spot, the only place I know of it is in the doc for https://libertem.github.io/LiberTEM/reference/api.html#libertem.api.Context.load under the filetype parameter.

I think I may also find a way to actually insert the keys somewhere explicit and clear on the formats page etc.

sk1p: I agree these keys need to be listed more prominently.

We do have `docs/source/autogenerated`, where we have automatically generated cli help, updated on all the docs tox targets. Could run something like

```bash
python -c 'from libertem.io.dataset import filetypes; print(", ".join(list(filetypes.keys())))'
```

... and put the result into an `.rst` file.

matbryan52: I went with the   `# doctest: +SKIP` solution for now, there would be a lot of work to get every example functional and connected to the right data. I'd prefer to get this merged for now, and the functionality is at least tested through the normal routes.

matbryan52:
@sk1p this is ready to go (or to review, but it's doc-only so hopefully not too bad!)

src/libertem/io/dataset/frms6.py
```diff
@@ -433,6 +433,11 @@ class FRMS6DataSet(DataSet):
     sync_offset: int, optional
         If positive, number of frames to skip from start
         If negative, number of blank frames to insert at start
+
+            Examples
```
sk1p, line 437: Indentation issue?

matbryan52: Good spot, fixed now

sk1p (approved):
LGTM

PR #1438, opened 2023-05-11
Forward `silence_logs` from `cluster_kwargs` to distributed workers

### LINKED ISSUE #1420: Preloading / Unloading workers can take a long time on huge systems + stderr spam
This is possibly an upstream issue, but creating our cluster and tearing it down (particularly in the case of a crash) can take a really long time of a machine with many cores. In my case on a 64 core machine the web interface can become responsive for a long time while the cluster preloads.

In addition the preload `stderr` spam can be extensive, which I think we could better configure for our usecase even if it's the upstream default.

matbryan52:
Fixes #1420

Silencing the `distributed.preloading` and `distributed.utils` output when creating our Context seems to be surprisingly tricky. 

The documented mechanism is through the `silence_logs` keyword argument to `dd.SpecCluster`, which we set by default to `logging.WARN`. This value is used in `dd.SpecCluster.__init__` to set the level on all `distributed` logging handlers, but evidently the `INFO` messages are still getting through from all the child processes. (might warrant an upstream issue?)

It also seems that the values we set with`dask.config.set` are not normally synchronized to the workers after they are spawned (https://github.com/dask/distributed/issues/3882), and in particular I don't think using this mechanism to change log levels works because logging is initialized from the default config immediately after the import of distributed.

In issues and the docs I've always seen it recommended to set logging level using dask config.yml files, which we don't have the option to use. We can however set the corresponding environment variable `DASK_DISTRIBUTED__LOGGING__DISTRIBUTED` and this is picked up by the child processes when they are spawned. Of course the consequence is that this sets the logging level for the life of the worker processes (which could be a good or bad thing...).

- [x] Check this works on Windows, given the processes are spawned differently.
- [x] Implement as a context manager so that the new envvar is popped after cluster creation (won't help the child processes but will mean further calls to `distributed` won't be impacted
- [x] Check error / warning messages are still received during preloading
- [x] Check if externally setting `DASK_DISTRIBUTED__LOGGING__DISTRIBUTED` is respected
- [x] Correctly handle when `dask.config` already has a value for `distributed.logging` (i.e. from a config.yaml file on the user's system, or a prior call to `dask.config.set`)
- [x] Test cases

## Contributor Checklist:

## Reviewer Checklist:


sk1p:
With a _very_ simple test case like this:

```python


if __name__ == "__main__":
    with Context() as ctx:
        pass
```

I can confirm that using this PR silences the log spam at startup. A robust test case probably has to start a script like that in a subprocess and check that the number of lines in `stdout`/`stderr` is ~small. For inspiration, there are a few test cases like that already for the pipelined executor:

https://github.com/LiberTEM/LiberTEM/blob/1352955a6f90d406bbc9a44e3f73142de0be4d34/tests/executor/test_pipelined.py#L470-L501

matbryan52:
@sk1p I've added a test case and rebased this on top of #1443 so I could use it in the test. Once we get that merged hopefully it'll be good to go.

matbryan52:
@sk1p  I think this is ready to go, if you think the testcase is sufficient. It launches a cluster and checks we don't have any distributed - <module> INFO messages in the stderr. Given that the number of messages might change due to upstream behaviour I didn't neecssarily want to check the *number* of messages we get (in practice it's zero!).

sk1p (commented):
> @sk1p I think this is ready to go, if you think the testcase is sufficient. It launches a cluster and checks we don't have any distributed - INFO messages in the stderr. Given that the number of messages might change due to upstream behaviour I didn't neecssarily want to check the _number_ of messages we get (in practice it's zero!).

Testcase looks good.

As this includes regenerating the notebook outputs, it is affected by the issue fixed in https://github.com/dask/dask/pull/10385 - maybe we can pin `dask` to `!= 2023.6.1` for now for the `notebooks` tox env, and patch the `ename` for the live-plotting notebook? Otherwise, we'll get failing tests again once the upstream issue is fixed.

Other than that, I agree that this is good to go!

matbryan52:
Seems to be all good with the pin and the manual fix, @sk1p

sk1p (approved):
LGTM!

PR #1441, opened 2023-05-24
Experimental API for dynamic UDF parameter updates

sk1p:
This allows to dynamically update UDF parameters while LiberTEM is running the UDF. Implemented for `dask`, `pipelined`, `inline`, and `concurrent` for now (not implemented for `delayed`).

With this patch applied, the following code works:

    result_iter = lt_ctx.run_udf_iter(dataset=default_raw, udf=[DynamicParamsUDF(latest_index=0)])
    with contextlib.closing(result_iter) as result_iter:
        for idx, part_res in enumerate(result_iter):
            # a list of dicts, with one dict for each UDF in the `run_udf_iter` call:
            result_iter.update_parameters_experimental([
                {"latest_index": idx + 1}
            ])

In combination with LiberTEM-live, this gives you interactivity while a scan is running.

## Discussion points

- How does this integrate with the UDF interface? It might be needed to declare UDF parameters, and how/if they can change over time
   - Possibly defer this for now? Or add a section to docs that UDF authors should document if parameters can vary
- This already makes it possible to implement tighter-than-per-scan feedback loops, if a certain remaining latency can be tolerated (upper bound `n_workers * processing_time_per_partition` basically)
- If we want to reduce the latency further, we would have to have out-of-band communication - for example, the UDF could actively query an "in-memory database" about the most recent values of certain parameters, or there could be a second set of queues
- How does this relate to per-frame metadata? I.e. "zipping" the normal data stream with a matching stream of metadata. This certainly needs a different interface on the sending side/data source, but can the UDF side be the same for both use cases?

## TODO

- [x] Update parameters for `UDF.merge` and other functions that run on the main node
- [x] Executor support - which ones are doable?
- [x] Un-flake the tests (maybe we can actually synchronize somehow instead of relying on flimsy timing)
- [ ] Move code: new classes might not be so happy in `libertem.udf.base` or `libertem.api`
- [x] Implement "patching"/partial updates of UDF parameters
- [x] Documentation

## Contributor Checklist:

## Reviewer Checklist:


sk1p:
@matbryan52 @uellue this also ties in with the pre-process-before-running ideas we discussed last month - in some cases, the pre-processing when changing parameters can happen concurrently with the running data acquisition, and once done, the new parameters are pushed to the workers.

matbryan52:
This is looking great, and surprisingly simple in the implementation!

From my understanding, the implication is that the `List[params]` passed to `result_iter.update_parameters` are all the `**kwargs` passed to the UDFs `__init__` and accessible normally under `self.params.<name>` within the UDF process methods? In this case params will always be a dict-like?

To cover the case of only wanting to do a partial update of the params, could we use `dict.update` between the existing scattered params and the scatter_update params, easiest to see in the `PipelineExecutor` case, where currently you replace fully:

```python
work_mem[key] = header["value"]
```

instead:

```python
for existing_kwargs, kwarg_update in zip(work_mem[key], header["value"]):
    existing_kwargs.update(kwarg_update)
```

Then an update like `result_iter.update_parameters([{}, {}, {'b': 55}])` would only update parameter `'b'` on the third UDF. 

As an alternative, since the `ResultsForDataSet` is provided on the main node and hence has the same UDF objects that the user supplied, could we even provide something like:

```python
class ResultsForDataSet:
   ...
   
   def update_udf_params(self, udf, **params):
        # Supply one-or-more param updates to a single UDF
        # Logic to lookup the index of `udf` within `params`
         ...
```

retaining the existing endpoint for a batch update of multiple sets of UDF params at once.

sk1p:
Thanks for having a look!

> From my understanding, the implication is that the `List[params]` passed to `result_iter.update_parameters` are all the `**kwargs` passed to the UDFs `__init__` and accessible normally under `self.params.<name>` within the UDF process methods? In this case params will always be a dict-like?

Yes - this is how we build the `self.params` thingy and how we separately scatter the parameters from the task object - we extract the `**kwargs` from `UDF._kwargs`, which are then put into an `UDFParams` object.

> To cover the case of only wanting to do a partial update of the params, could we use `dict.update` between the existing scattered params and the scatter_update params, easiest to see in the `PipelineExecutor` case, where currently you replace fully:
> 
> ```python
> work_mem[key] = header["value"]
> ```
> 
> instead:
> 
> ```python
> for existing_kwargs, kwarg_update in zip(work_mem[key], header["value"]):
>     existing_kwargs.update(kwarg_update)
> ```

Hmm - I have to think a bit about this. In general I agree that this is nice to have (especially if you have some large parameters!) - on the other hand, right now the executor doesn't know about the inner structure of the scattered object. Maybe the generic idea of this is a `scatter_update_patch` method for the executor, which assumes some structure for the object that is scattered. Something like an interface `Updatable` with a method `update(self, patch)` where `patch` is some opaque object which updates `self` in some way.

The above loop won't directly work, as `work_mem[key]` is  an `UDFParams` instance, which is not iterable. So in the concrete use-case for `UDFParams`, we would add a method `UDFParams.update(self, patch: List[dict])` which would implement the loop above, and which would be called on the worker by the respective executors.

Does that make sense?

> Then an update like `result_iter.update_parameters([{}, {}, {'b': 55}])` would only update parameter `'b'` on the third UDF.

One case this doesn't completely cover is optional parameters, i.e. going from a concrete value to not specifying said value. Then one has to specify the default value explicitly (could be `None`, but could also be something else). I'd be fine with this, if we properly document this.

> As an alternative, since the `ResultsForDataSet` is provided on the main node and hence has the same UDF objects that the user supplied, could we even provide something like:
> 
> ```python
> class ResultsForDataSet:
>    ...
>    
>    def update_udf_params(self, udf, **params):
>         # Supply one-or-more param updates to a single UDF
>         # Logic to lookup the index of `udf` within `params`
>          ...
> ```

True - I think I discussed this shortly with @uellue too, and I convinced him that list-of-dict is fine, as this is symmetric to how you get the results with `run_udf_iter`. But we still may want to re-consider this...

One nice thing about the list-of-dict API is that as a user, you don't necessarily need to keep track of the UDF instances. I'm thinking of a pattern like this:

```python
udf_parameters : List[dict] = get_current_parameters()
udf_classes = get_udf_classes()
udfs = [cls(**params) for cls, params in zip(udf_classes, udf_parameters)]
res_iter = ctx.run_udf_iter(dataset=ds, udfs=udfs):
    new_parameters = get_current_parameters()  # this might re-calculate based on user input, or update parameters based on latest results or whetever
    if new_parameters != udf_parameters:
        res_iter.update_parameters(new_parameters)  # ... possibly not `new_parameters` directly, but a patch between `udf_parameters` and `new_parameters`
        udf_parameters = new_parameters
```

What kind of pattern did you think of, where the `update(udf, params)` style would make most sense?

Btw this makes me think of something else - right now we only update the UDF instances on the workers, but the parameters of the UDF instance where the merge function is running, i.e. the main process, might also need to be updated in the same way...

> retaining the existing endpoint for a batch update of multiple sets of UDF params at once.

:+1:

> I haven't dug into it but I assume the update will work for params which are `BufferWrapper`, too ?

Probably - I think the only way these are handled differently is in per-task initialization, and when you access them form the UDF. But a good point that we should have at least a test case for this.

sk1p:
Another topic is executor support.

- `inline`: done, tested
- `pipelined`: done, needs integration test
- `concurrent`: eagerly creates futures for all tasks
- `dask`: does limit the number of in-flight futures, so new futures could (in theory) use the new parameters. Or we could try to control the scatter keys that are used by scattering a `dict`, but I'm not sure I'm reading [the API docs](https://distributed.dask.org/en/stable/api.html#distributed.Client.scatter) correctly...
- `delayed`: I think same issue as `concurrent`? Also IIRC no real support for partial results?

I think mostly the issue is that tasks are generated eagerly in most executors, meaning they are basically a closure over the initial parameters. But `concurrent` could be changed to not have too many in-flight futures, similarly to the dask executor, creating futures on demand and with updated parameters, if they changed.

Another question is, how useful is this feature for offline processing? Maybe if the processing is slow and you can already see that the results are incorrect?

sk1p:
Discussion with @uellue:

- Some UDFs currently rely on changing parameters (example: the record UDF which will probably fail completely if parameters aren't passing through the UDF constructor, which does something different on main node vs. on workers)
- Patching semantics make sense: users don't want to have parameters in the update dict which should not be changed
- It could be necessary to pass newly constructed UDF instances instead of only the parameters, so we know that the parameters have been through the constructor
   - Then we need to compare parameters with the old ones to implement patching, which gives rise to the question of either doing a shallow object-identity comparison vs doing a deep equality check. The latter can be quite expensive (dask uses hashing for this and puts a 500MB/s number to it...), but would possibly be expected by the user for efficient updates of large array parameters.

Hmm, there is also a connection to our discussion in Grenoble - the way things are pointing, the _outer_ parameters of UDFs should be JSON-serializable in the future, if this is possible at all. This is important if we want to build a live-processing server that runs any number of UDFs, remotely from the user interface. If we have a `preprocess` step, for example for baking trotter stacks for SSB, then this has to be re-run on parameter changes. If the outer parameters of the UDF are JSON serializable, it is also no big deal to do a deep comparison for efficient delta updates.

JSON-serializable parameters also aligns with extending the ideas at #1376 to support complete specification of a "batch job": a dataset plus a list of UDFs plus their parameters, all specified in a TOML (or YAML/JSON) file. I think we discussed something in this direction already before.

One possibility I discussed with @uellue: introduce this interface first as `update_parameters_experimental`, which is replaced in the future with a more fleshed-out `update_parameters` method. The replacement then supports declarative parameters, better aux data (however that looks like), etc. and incorporates any lessons learned from the low-level experimental method. 

I hope I captured the essence of our discussion @uellue - feel free to add anything I missed!

matbryan52 (commented):
Review comment to say that I've at least understood how all of the different executors implement this, and the checked the interfaces. Firstly, again, good job on how minimal some of the implementations are !

I still need more passes over the code to think about edge cases. I will say that the different behaviour according to Executor type, dependent on how the each executor schedules tasks (i.e. eagerly or not) will need some clear documentation!

src/libertem/udf/base.py
```diff
@@ -62,6 +62,52 @@
...
+
+    def update_parameters_experimental_full(self, parameters: List[Dict[str, Any]]):
+        log.debug("ResultsForDataSet.update_parameters_experimental_full: %s", parameters)
+        # update parameters on workers (via the executor):
+        self._params.update_parameters(parameters=parameters)
+        self._executor.scatter_update(self._params_handle, self._params)
+        # update UDFs on the main node:
+        for udf, params in zip(self._udfs, parameters):
+            udf._update_parameters(new_kwargs=params)
+
```
uellue, line 100: What is the status and purpose of this function? It is not mentioned in the documentation, and it doesn't seem to have test coverage?

src/libertem/udf/base.py
```diff
@@ -919,6 +965,31 @@ def __init__(self, *args, **kwargs) -> None:
...
+    def _update_parameters(self, new_kwargs):
+        """
+        Update the parameters _for this UDF instance_ to `new_kwargs`.
+
+        For properly supporting AUX data (i.e. `BufferWrapper`-typed parameters),
+        make sure to set the correct views afterwards.
+
+        :meta private:
+        """
+        self.params = UDFKwargsWrapper(new_kwargs)
```
uellue, line 977: See comment above, only point where this is called?

src/libertem/udf/base.py
```diff
@@ -1601,6 +1672,16 @@ def from_udfs(
         kwargs = [udf._kwargs for udf in udfs]
         return cls(kwargs, roi, corrections, tiling_scheme)
 
+    def update_from_udfs(self, udfs):
+        self._kwargs = [udf._kwargs for udf in udfs]
+
+    def update_parameters(self, parameters):
+        self._kwargs = parameters
```
uellue, line 1679: See above, not covered and apparently only used in one place?

src/libertem/api.py
```diff
@@ -68,6 +66,101 @@
...
+        """
+        Update parameters while the UDFs are running.
+
+        `parameters` should be a list of dicts, with one dict for each
+        UDF you are running.
+
+        The dicts should only contain items for those parameters you want to
+        update.
+        """
+        logger.debug("ResultGenerator.update_parameters: %s", parameters)
```
uellue, line 100: The function name in the log message is different from the actual function name.

Would introspection be an option to get the class and function name?

src/libertem/udf/base.py
uellue, line 100: This function is not documented and not covered by tests, if I see correctly? What is the status of it?

src/libertem/udf/base.py
uellue, line 977: See above, apparently not used?

uellue (changes requested):
Looks good, generally! Tested successfully in https://github.com/LiberTEM/Microscope-Calibration

If I understand correctly, there's an orphaned set of methods to perform a full update that is not documented and not covered by tests? Maybe remove it, comment it out, ...? If I understand correctly, it would allow updating something that cannot be patched? Otherwise the only difference to patching all parameters would be that it can eliminate parameters altogether instead of setting them to some value?

src/libertem/api.py
sk1p: Good catch, fixed. Introspection is not worth the hassle IMHO - IIRC you can globally configure the log format to include these automatically, if you really want to.

sk1p:
> If I understand correctly, there's an orphaned set of methods to perform a full update that is not documented and not covered by tests? Maybe remove it, comment it out, ...?

I'll remove them, good catch.

> If I understand correctly, it would allow updating something that cannot be patched? Otherwise the only difference to patching all parameters would be that it can eliminate parameters altogether instead of setting them to some value?

Yeah, exactly: there is one corner case that is not obvious with the patch API: if you want to go back to the default value for a parameter, you need to explicitly specify the default value, you can't just leave it out. But that should not be too common, so I went ahead and implemented the patching version, just didn't remove all the full update methods yet.

src/libertem/udf/base.py
sk1p: Removed.

src/libertem/udf/base.py
sk1p: Removed.

src/libertem/udf/base.py
sk1p: Removed.

uellue (approved):
LGTM, nice! 👍

PR #1448, opened 2023-06-23
Properly expose UDF cancellation

sk1p:
Previously, the `JobCancelledError` was bubbling up to the user, which is not really part of the public API, and also caused a queue drain in the `PipelinedExecutor`. This PR replaces it with a properly documented `UDFRunCancelled` exception.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/udf/base.py
+++ b/src/libertem/udf/base.py
@@ -43,7 +43,9 @@ from libertem.common.async_utils import async_generator_eager
 from libertem.executor.inline import InlineJobExecutor
 from libertem.common.executor import (
     Environment, JobExecutor, TaskCommHandler, NoopCommHandler, TaskProtocol,
+    JobCancelledError,
 )
+from libertem.exceptions import UDFRunCancelled
 
 if TYPE_CHECKING:
     from numpy import typing as nt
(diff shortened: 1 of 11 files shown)


src/libertem/udf/base.py
```diff
@@ -2388,25 +2390,28 @@ def run_for_dataset_sync(
...
                     udfs=self._udfs,
                     damage=damage
                 )
-        if not any_result or not iterate:
-            yield self._make_udf_result(
-                udfs=self._udfs,
-                damage=damage
-            )
+        except JobCancelledError:
+            raise UDFRunCancelled
```
matbryan52, line 2414: Could consider a descriptive error message ? If we change `any_result` to an integer which we increment each time through to loop:

```python
        any_result = 0
        try:
            for part_results, task in result_iter:
                any_result += 1
                with tracer.start_as_current_span("_apply_part_result -> UDF.merge"):
                    self._apply_part_result(
                        udfs=self._udfs,
                        damage=damage,
                        part_results=part_results,
                        task=task
                    )
                if iterate:
                    yield self._make_udf_result(
                        udfs=self._udfs,
                        damage=damage
                    )
            if not any_result or not iterate:
                yield self._make_udf_result(
                    udfs=self._udfs,
                    damage=damage
                )
        except JobCancelledError:
            raise UDFRunCancelled(f'UDF run cancelled after {any_result} partitions')
```

Unfortunately we can't show the expected number of tasks because it's not exposed on the `result_iter`, I think.

matbryan52 (approved):

src/libertem/udf/base.py
sk1p: Good point, I'll improve the error message

sk1p: This is now implemented in 2ebac2bcf6ab42d82767f30c3adf3998aafc606e, thanks for the feedback!

matbryan52: Great, looks good.

PR #1451, opened 2023-06-27
Improved scheduling for the `PipelinedExecutor`

sk1p:
As [discussed before](https://github.com/LiberTEM/LiberTEM/pull/1267/files#r927542348) , an easy improvements on the round-robin scheduling implemented before, is sending tasks always to the process with the shortest request queue. This is implemented in these changes.

This should improve GPU performance, too, although I have not benchmarked this yet. With these changes, it should also be quite easy to preferentially send tasks to GPU workers when they are idle. Implementing proper resource support should now also be possible (see #1453).

In addition, this also speeds up the tests by a lot; on the system I tested a full `tests_pipelined` run goes from ~108s to ~28s - by using less worker processes, and by catching errors on startup faster. Also marked the "default spec" case as slow, as that will always try to start ~many workers on large machines.

This sits on top of #1443 and I'll rebase once that is merged.

## TODO

- [x] Rebase once #1443 is merged

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/docs/source/changelog/features/pipeline-scheduling.rst
+++ b/docs/source/changelog/features/pipeline-scheduling.rst
@@ -1,0 +1,4 @@
+[Feature] Improved scheduling for the pipelined executor
+========================================================
+
+* Schedule tasks on the worker with the smallest request queue size (:pr:`1451`).
--- a/src/libertem/common/executor.py
+++ b/src/libertem/common/executor.py
@@ -453,6 +453,9 @@ class WorkerQueue:
         """
         raise NotImplementedError()
 
+    def size(self) -> int:
+        raise NotImplementedError()
+
 
 class SimpleWorkerQueue(WorkerQueue):
     """
@@ -481,6 +484,9 @@ class SimpleWorkerQueue(WorkerQueue):
 
     def close(self, drain: bool = True, force: bool = False):
         pass
+
+    def size(self) -> int:
+        return self.q.qsize()
 
 
 class SimpleMPWorkerQueue(WorkerQueue):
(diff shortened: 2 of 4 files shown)


sk1p:
To keep this PR short (once it is rebased), I'm putting the work to support task resources into its own PR

sk1p:
Oof, looks like Mac OS doesn't support getting the queue size, so there needs to be a fallback to the previous behavior.

matbryan52 (approved):
Looks good to me. I noted there isn't really a good way to test the scheduling implementation -one way could be to play some games with `time.sleep()` in the first task and `os.getpid()` to figure out where each task executed. Not sure it's worth it, though!

sk1p:
Thanks for the review!

> Looks good to me. I noted there isn't really a good way to test the scheduling implementation -one way could be to play some games with `time.sleep()` in the first task and `os.getpid()` to figure out where each task executed. Not sure it's worth it, though!

Yeah - that has the potential to be a flaky test too, as the queue `size` method this is based on is "not reliable" in an unspecified way. In #1453 there are a few more tests, including [`test_device_classes_limit_to_cpus`](https://github.com/LiberTEM/LiberTEM/pull/1453/files#diff-db33431c53df549272cb75f13b9c5f686963076c91fc1c2b2de15da2d6aebc6eR185) where we do check if at least each CPU worker did get work at all. It's also complicated by the live case using the queue for more than just one message per task - meaning the scheduling should in theory work better when used with LiberTEM-live, which is hard to check from tests in LiberTEM. So I'm fine with not having an explicit test for scheduling behavior.

I'm going to merge this now; if we are inspired, we can improve the testing further in #1453

PR #1453, opened 2023-06-27
Resource constraints for the `PipelinedExecutor`

### LINKED ISSUE #1452: `PipelinedExecutor`: support for resource constraints
Update the `TaskProtocol` to include resource definitions and honor them when scheduling on `PipelinedExecutor`. Add tests for resource constraints - should [pass similar tests as the dask scheduler](https://github.com/LiberTEM/LiberTEM/pull/1267#discussion_r930174099)

sk1p:
Fixes #1452 - support for resource constraints in the `PipelinedExecutor`.

## TODO

- [x] Sits on top of  #1451 - rebase once that is merged
- [x] (Documentation and) changelog entry
- [x] Refactor a bit to not build worker resources on the fly from the `WorkerSpec` objects, and just keep a `libertem.common.scheduler.Worker` object around (possibly in addition to the `WorkerSpec`).
- [x] Maybe: move scheduling logic to  `libertem.common.scheduler.Scheduler`?
- [x] Updated tests

## Contributor Checklist:

## Reviewer Checklist:


sk1p:
This should now be ready for review.

I didn't add any new documentation, as this PR basically means the `PipelinedExecutor` can be used [as documented before](https://libertem.github.io/LiberTEM/udf/advanced.html#cupy-support) for CuPy-capable UDFs. Let me know I missed something.

src/libertem/executor/pipelined.py
```diff
@@ -128,22 +166,31 @@ def kill_worker(self, worker_info: PoolWorkerInfo, timeout: float = 5.0):
...
+            raise PoolStateError("Response queue not available")
         self._response_q.close(drain=False, force=True)
         self._response_q = None
-        assert all([e is not None for e in exitcodes])
+        if any([e is None for e in exitcodes]):
+            raise PoolStateError("At least one worker failed to be killed")
 
     @property
     def workers(self) -> List[PoolWorkerInfo]:
+        if self._workers is None:
```
matbryan52, line 187: Is it important that this raises? I would expect `self._workers is None` to return `[]`. Seems to match the 'somewhat permissive' form of `all_alive(self)`.

src/libertem/executor/pipelined.py
```diff
@@ -791,9 +900,26 @@ def _run_tasks_inner(
         # of exceptions, it becomes a bit harder to keep attribution of messages to
         # tasks, which is why we have a separate counter for now.
         in_flight = [0]
```
matbryan52, line 902: I can't recall why this needs to be a list ? Maybe for the use in the closure ? Could possibly get away using:

```python
in_flight = 0
...
def yield_result_if_found(block, timeout):
    nonlocal in_flight
    in_flight -= 1
    ...
...
in_flight += 1
```

src/libertem/executor/pipelined.py
```diff
@@ -816,17 +942,33 @@ def yield_result_if_found(block, timeout):
...
-                        assert len(id_to_task) == in_flight[0]
+                        if len(id_to_task) != in_flight[0]:
+                            raise RuntimeError(
+                                "state mismatch; `id_to_task` mapping should match `in_flight`"
+                            )
                 except WorkerQueueEmpty:
                     self._validate_worker_state()
 
             for task_idx, task in enumerate(tasks):
+                # important: call `schedule_task` before incrementing number of
```
matbryan52, line 953: If this is the case then should we only increment `in_flight` and add to `id_to_task` *after* the `'RUN_TASK'` message is sent (and maybe even after `'END_TASK'` too ?), in case for some reason sending the task fails (queue closes, process crashes etc...)

src/libertem/executor/pipelined.py
```diff
@@ -816,17 +942,33 @@ def yield_result_if_found(block, timeout):
...
+                # in-flight tasks, otherwise, if we run into an exception, we
+                # will wait for a response that never comes.
+                _worker_idx, worker_queues = schedule_task(
```
matbryan52, line 956: It's just naming, but `schedule_task` makes me think that this function call actually schedules the task, instead it performs a function more aligned with a name like `choose_worker`. The actual scheduling happens when we send `RUN_TASK` (for me).

src/libertem/executor/pipelined.py
```diff
@@ -65,6 +65,38 @@ class PoolWorkerInfo(NamedTuple):
...
 
 
+class PoolStateError(Exception):
+    """
+    The worker pool is not in the expected state to perform the requested operation
+    """
+    pass
+
+
+class ResourceError(RuntimeError):
```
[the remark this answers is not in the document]
sk1p: Yeah, the dask executor already raises a `RuntimeError` in some cases, I'll convert those to `ResourceError`, too. Moved to `libertem.executor.base` for now.

src/libertem/executor/pipelined.py
sk1p: Yup, that works, too

src/libertem/executor/pipelined.py
sk1p: Yup, moved things around a bit

src/libertem/executor/pipelined.py
sk1p: Naming is hard ;) For me, scheduling is more of a planning step, but I can understand that this can be a bit confusing. `choose_worker` IMHO also doesn't fully fit, as it is a different step than the `selector` function that is passed in (but includes calling said `selector`).. Hmm, I'll leave it as-is for now, until we can agree on/find a better name.

matbryan52: No worries, could just add a comment above the call saying that we are selecting the most appropriate worker for the task here.

matbryan52:
For me this is good to go, LGTM ?

PR #1458, opened 2023-07-04
Add COMUDF

matbryan52:
Adds `COMUDF`, continued from #1392 .

Notes:
- The default analysis parameters have been lifted into the definition of the `COMParams(NamedTuple)`, which makes the default functionality a bit clearer without digging into the code of `get_task_data`.
- Added a `@classmethod` `COMUDF.with_params(...)` to allow construction of a UDF with the desired `COMParams`, but mainly to give a clear place where the behaviour of each parameter is documented.
- Added a `raw_com` result buffer for the position of the CoM in the coordinate system of the frame - i.e. COMUDF returns the CoM, in addition to the *shift* in the CoM + field-derived results.
- Documented all of the result buffers in the docstring of `COMUDF`, again for visibility.

Two ideas:
- Any enthusiasm for renaming the UDF `CoMUDF` ? It would make it feel like less of a global variable, to me.
- I've always wondered about implementing a mode where the `raw_shifts` (and hence field++) are calculated from the *mean* of `raw_com` result, so that any error in the specification of `cy`/`cx` relative to the actual data centre doesn't give shifts / field which are all offset in one direction. Could also return the mean value and/or reference value in a result-only buffer for transparency. Any thoughts on that?

* [x] Check if this is still compatible with the `iCoM` repo

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/udf/com.py
+++ b/src/libertem/udf/com.py
@@ -8,6 +8,7 @@ from libertem import masks
 from libertem.corrections import coordinates
 from libertem.udf.masks import ApplyMasksEngine
 from libertem.common.container import MaskContainer
+from libertem.common.math import prod
 from libertem.udf.base import UDF
 
 
@@ -353,6 +354,22 @@ class COMUDF(UDF):
             raw_result, self.results.raw_mask_result
         )
 
+    def get_field_results(self, field_y, field_x):
+        '''
+        To be overwritten in subclasses, such as for iCoM results
+        in https://github.com/LiberTEM/LiberTEM-iCoM
+        '''
+        mag = magnitude(
+            y_centers=field_y, x_centers=field_x
+        )
+        div = divergence(y_centers=field_y, x_centers=field_x)
+        curl = curl_2d(y_centers=field_y, x_centers=field_x)
+        return {
+            'magnitude': mag,
+            'divergence': div,
+            'curl': curl,
+        }
+
     def get_results(self):
         com_params = self.get_params()
         raw_mask_result = self.results.get_buffer('raw_mask_result')
@@ -375,28 +392,20 @@ class COMUDF(UDF):
 
         raw_shifts = np.moveaxis(np.array(raw_shifts), 0, -1)
         field = np.moveaxis(np.array(field), 0, -1)
-        mag = magnitude(
-            y_centers=field_y, x_centers=field_x
-        )
-        div = divergence(y_centers=field_y, x_centers=field_x)
-        curl = curl_2d(y_centers=field_y, x_centers=field_x)
-        if self.meta.roi is not None:
-            raw_shifts = raw_shifts[roi]
-            field = field[roi]
-            mag = mag[roi]
-            div = div[roi]
-            curl = curl[roi]
-        else:
-            raw_shifts = raw_shifts.reshape((-1, 2))
-            field = field.reshape((-1, 2))
-            mag = mag.reshape(-1)
-            div = div.reshape(-1)
-            curl = curl.reshape(-1)
 
-        return {
+        nav_size = prod(self.meta.dataset_shape.nav)
+
+        results = {
             'raw_shifts': raw_shifts,
             'field': field,
-            'magnitude': mag,
-            'divergence': div,
-            'curl': curl,
         }
+
+        results.update(self.get_field_results(field_y=field_y, field_x=field_x))
+
+        if self.meta.roi is not None:
+            for key in results:
+                results[key] = results[key][roi]
+        else:
+            for key in results:
+                results[key] = results[key].reshape((nav_size, -1))
+        return results
--- a/src/libertem/udf/com.py
+++ b/src/libertem/udf/com.py
@@ -28,6 +28,13 @@ RegressionOptionsT = Union[np.ndarray, Literal[-1, 0, 1]]
 
 
 class CoMParams(NamedTuple):
+    """
+    Used to instantiate :class:`~libertem.udf.com.CoMUDF`.
+
+    See :meth:`~libertem.udf.com.CoMUDF.with_params`
+    for descriptions of each parameter. In most cases use this
+    method rather than instantiating this class directly.
+    """
     cy: Optional[float] = None
     cx: Optional[float] = None
     r: float = float('inf')
@@ -181,6 +188,14 @@ def coordinate_check(y_centers, x_centers, roi=None):
 
 
 class GuessResult(NamedTuple):
+    """
+    Container for CoM parameters inferred from results,
+    which can be passed to :meth:`~libertem.udf.com.CoMUDF.with_params`
+    to create a new :class:`~libertem.udf.com.CoMUDF` better-adapted
+    to the data.
+
+    Return type of :func:`~libertem.udf.com.guess_corrections`
+    """
     scan_rotation: int
     flip_y: bool
     cy: float
@@ -193,7 +208,8 @@ def guess_corrections(
     roi: Optional[Union[np.ndarray, Tuple[slice, ...]]] = None,
 ) -> GuessResult:
     '''
-    Guess corrections for center shift, :code:`scan_rotation` and :code:`flip_y` from CoM data
+    Guess values for center offset (:code:`(cy, cx)`), :code:`scan_rotation` and
+    :code:`flip_y` from existing CoM data.
 
     This function can generate a CoM parameter guess for atomic resolution 4D STEM data
     by using the following assumptions:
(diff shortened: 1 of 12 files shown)


matbryan52:
> Hi @matbryan52, :+1: for renaming! Would you mind if I take over the PR to finish it for the 0.12 release, or what are your plans with it? :-)

Go ahead! I'm still away this week so won't be able to work on it until the next. Feel free to push onto my branch if that's helpful.

uellue:
> * I've always wondered about implementing a mode where the `raw_shifts` (and hence field++) are calculated from the _mean_ of `raw_com` result, so that any error in the specification of `cy`/`cx` relative to the actual data centre doesn't give shifts / field which are all offset in one direction. Could also return the mean value and/or reference value in a result-only buffer for transparency. Any thoughts on that?

The correct compensation would be subtracting a vacuum reference. Subtracting the mean works if there's no long-range field. If we pass the beam through a constant electric or magnetic field, an equal shift of all points could be the actual result. On the other hand, I agree that the shift relative to the mean or even relative to a best-fit linear gradient (descan error) is often a relevant value. :+1: for additional result-only buffers!

uellue:
After discussion with @sk1p: Correct the field instead of adding more result buffers so that iCoMUDF works organically with the flat field correction, where it is probably particularly useful.

src/libertem/udf/com.py
```diff
@@ -0,0 +1,630 @@
...
+        non-NaN values even in :code:`ROI == False` pixels due to the behaviour
+        of :code:`np.gradient`.
+
+    Parameters
+    ----------
+    com_params : CoMParams
+        A :class:`CoMParams` instance containing the parameters
+        for this UDF. By default will create a CoMParams instance
+        which performs whole-frame CoM with results in the coordinates
+        of the frame # CHECKTHIS
```
sk1p, line 306: Aren't the results relative to the detector center? See `get_params` below.

src/libertem/udf/com.py
```diff
@@ -0,0 +1,630 @@
...
+
+    Examples
+    --------
+    >>> udf = CoMUDF()
+    >>> result = ctx.run_udf(dataset=dataset, udf=udf)
+    >>> result["magnitude"].data.shape
+    (16, 16)
+    """
```
sk1p, line 314: Hmm, could we have more examples here? One that uses `with_params` maybe?

src/libertem/udf/com.py
```diff
@@ -0,0 +1,630 @@
...
+        :meth:`apply_correction`
+    roi : Optional[numpy.ndarray]
+        Selector for values to consider in the statistics, compatible
+        with indexing an array with the shape of y_centers and x_centers.
+        By default, everything except the last row and last column are used
+        since these contain artefacts.
+
+    Returns
+    -------
+    GuessResult : relative to current values
```
sk1p, line 211: Can we have an example here how to pipe the result from the `CoMUDF` into `guess_corrections`? I think that would be a common use-case.

That would also mean this function needs to be added to the reference docs, or it needs to be exposed (as a static method?) on the `CoMUDF` itself.

src/libertem/udf/com.py
```diff
@@ -0,0 +1,630 @@
...
+            cy=cy, cx=cx, r=r, ri=ri, scan_rotation=scan_rotation, flip_y=flip_y,
+            regression=regression,
+        )
+        return cp
+
+    def get_task_data(self):
+        sig_shape = tuple(self.meta.dataset_shape.sig)
+        com_params = self.get_params()
+        if len(sig_shape) != 2:
+            raise ValueError('CoMUDF only works with 2D sig shape.')
```
sk1p, line 458: Should we have test cases for incompatible sig/nav shapes? Should be easy enough with a memory dataset.

src/libertem/udf/com.py
```diff
@@ -0,0 +1,630 @@
...
+
+            Please note that a regression calculated on the data itself may
+            distort valid results, such as the effect of long-range fields.
+            Furthermore, it may yield erronerous results due to aliasing on
+            periodic structures or boundary effects. Subtracting a reference
+            measurement or regression on a reference measurement instead of a
+            regression on the data may give more reliable results.
+        """
+        if ri >= r:
+            raise ValueError('Inner radius must be less than outer radius for annular CoM')
```
sk1p, line 391: Could we have a test case for this edge case, too?

src/libertem/udf/masks.py
```diff
@@ -1,11 +1,99 @@
...
-from libertem.udf import UDF
+from libertem.udf import UDF, UDFMeta
 from libertem.common.container import MaskContainer
 from libertem.common.numba import rmatmul
 
 
+class ApplyMasksEngine:
+    def __init__(self, masks: MaskContainer, meta: UDFMeta, use_torch: bool):
+        self.masks = masks
+        self.meta = meta
```
sk1p, line 12: Just a comment, nothing to fix right now: taking a `UDFMeta` instance here means this class is right now only usable in UDFs. I do really love that this code is now reusable, and in the future we might want to lift it up further to make it usable stand-alone (possibly by extracting the relevant parameters into our own parameter class, which is then easily constructable from ` UDFMeta`.

sk1p (changes requested):
Looks almost ready! I've added some comments inline, most of them are just about minor changes, and adding test cases for the edge cases (to make sure they continue to be checked properly in the future)

About the CI failure, a rebase to current `master` should fix the flakyness of the affected test case, by adjusting the timings to those used in the integration tests running at UDF level, which were running successfully since #1441 was merged, as far as I can see.

I've opened #1482 for the remaining work to surface the new features in the web GUI.

matbryan52:
@uellue I'm happy to cover the review comments in the morning as part of looking at the PR in detail - I think some of them concern bits of docstring that I myself wrote.

src/libertem/udf/com.py
matbryan52: Agreed for the example, even I am not 100% sure how to pipe from one to the other...

The `guess_corrections` function is already documented under `analysis.com`, though its docstring references the functions `center_shifts` and `apply_corrections` which are themselves undocumented, currently. I will see if I can write something complete with the new functionality offered with CoMUDF.

matbryan52: Do you have a good way to :autodoc: a NamedTuple or Enum (e.g. https://libertem.github.io/LiberTEM/reference/api.html#libertem.analysis.com.GuessResult ) ? I think there are Sphinx extension that can do it, but otherwise the output is quite unhelpful!

sk1p: Yeah, named tuples aren't properly supported by sphinx, yet: https://github.com/sphinx-doc/sphinx/issues/8717

I wouldn't be opposed to using [`autonamedtuple`](https://sphinx-toolbox.readthedocs.io/en/latest/extensions/more_autodoc/autonamedtuple.html) from sphinx-toolbox, if that is compatible with the rest of our sphinx extensions.

matbryan52:
@uellue Thanks for picking up the change to sphinx, seems like this is nearly ready to go?

I think you noticed that I added an `IntEnum` for `RegressionOptions`, hopefully that's an OK choice @sk1p  ? We don't have that pattern elsewhere right now, I think.

matbryan52:
Also an observation - we import and document `guess_corrections` under `libertem.analysis.com`, which means the example in the docstring doesn't actually match the documentation of that file (analysis api rather than udf api). We could choose to remove the `guess_corrections` and `GuessResult` members documented here:

https://github.com/LiberTEM/LiberTEM/blob/060613939719efbe3847b053f893a250bcb3984d/docs/source/reference/api.rst?plain=1#L21-L22

sk1p:
> @uellue Thanks for picking up the change to sphinx, seems like this is nearly ready to go?

Yup, should be ready to go if CI is green.

> I think you noticed that I added an `IntEnum` for `RegressionOptions`, hopefully that's an OK choice @sk1p ? We don't have that pattern elsewhere right now, I think.

Yes, `IntEnum` is :+1:. I'd also be fine with a regular enum but it doesn't really matter to me (int enum has benefit that you don't necessarily need to import the specific enum itself, if you know the constants, with the readability drawback that comes with it...)

uellue:
> @uellue Thanks for picking up the change to sphinx, seems like this is nearly ready to go?

Oh, sry I hope I didn't disrupt your work by pushing?

> I think you noticed that I added an `IntEnum` for `RegressionOptions`, hopefully that's an OK choice @sk1p ? We don't have that pattern elsewhere right now, I think.

With autoenum the documentation for that one looks pretty decent, I like it!

matbryan52:
> > @uellue Thanks for picking up the change to sphinx, seems like this is nearly ready to go?
> 
> Oh, sry I hope I didn't disrupt your work by pushing?

Not at all, it had slipped my mind to do that last step adding `sphinx_toolbox`.

sk1p:
> Also an observation - we import and document `guess_corrections` under `libertem.analysis.com`, which means the example in the docstring doesn't actually match the documentation of that file (analysis api rather than udf api). We could choose to remove the `guess_corrections` and `GuessResult` members documented here:

Yeah, I'd be fine with documenting the new import location only, the old ones are basically only in place for backwards-compatibility.

PR #1478, opened 2023-07-26
WIP: Update docker build infra

sk1p:
- Use github registry for now (easier auth and IIRC not as rate limited)
- Build using the official docker github actions
- Don't keep the requirements.txt around; generate from scratch
- Update base image (Python 3.11)

## TODO

- [x] Build and tag in build stage, then re-tag and push in release stage
- [x] Update docs for new image tags (ghcr.io)
- [ ] Clean up the build scripts (probably only need one for building locally?)
- [x] Add a README file that documents the build process (What is the `libertem-dev` image? What is `libertem/libertem:triage`? Why do these exist?)

## Contributor Checklist:

## Reviewer Checklist:


sk1p:
Hmm, uploading images as artifacts is very slow... when we are building master/stable branches or tags, we can push into the registry instead, maybe tagged with the revision, and then pull/re-tag/push in the `release`-part of the whole thing?

sk1p:
@uellue @matbryan52 I'll merge these changes once CI passes, as the building seems to work fine, but the actual `push`/`release` action can only be tested on the master branch (and the version tagging later with release candidates)

PR #1483, opened 2023-08-08
Improve tracing

sk1p:
- More visibility into tasks of the pipelined executor: among other details, trace queue puts, as they can take a bit of time for larger values
- Fix execution plan tracing event - can't have an `OrderedDict` as value, so format as `List[str]`

## Contributor Checklist:

## Reviewer Checklist:


sk1p:
Oof, `test_scatter_update` is still (again?) flaky. One solution I see is to increase the number of tasks (by a lot) - if this then takes too much time, I guess it needs to be marked as `slow`.

PR #1493, opened 2023-08-10
Catch `PermissionError` when doing `os.stat()` on file or directory

### LINKED ISSUE #1491: Web GUI fails to open folder containing shortcuts on Windows with bad permissions
I stumbled across a (very) edge case for the web GUI - trying to open a folder on Windows which contains a shortcut to a directory to which I don't have access, I get the following UI error. It is possible to close the error and continue working, but not to access the directory.

```
Error browsing directory: SyntaxError: Unexpected token '<', "<html><tit"... is not valid JSON
```

and in the terminal I get the following exception:

```
2023-08-10 14:50:54,569 - tornado.application - ERROR - Uncaught exception GET /api/browse/localfs/?path=C%3A%5CUsers%5Cmb265392%2F.. (::1)
HTTPServerRequest(protocol='http', host='localhost:9000', method='GET', uri='/api/browse/localfs/?path=C%3A%5CUsers%...%2F..', version='HTTP/1.1', remote_ip='::1')
  File "D:\...\envs\lt_release012\Lib\site-packages\tornado\web.py", line 1786, in _execute
    result = await result
             ^^^^^^^^^^^^
  File "D:\...\envs\lt_release012\Lib\site-packages\libertem\web\browse.py", line 19, in get
    listing = await executor.run_function(get_fs_listing, path)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\...\envs\lt_release012\Lib\site-packages\libertem\executor\base.py", line 87, in run_function
    return await sync_to_async(fn_with_args, self._pool)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\...\envs\lt_release012\Lib\site-packages\libertem\common\async_utils.py", line 58, in sync_to_async
    return await loop.run_in_executor(pool, fn)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\...\envs\lt_release012\Lib\concurrent\futures\thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\...\envs\lt_release012\Lib\site-packages\libertem\common\async_utils.py", line 27, in _wrapped
    return f(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^
  File "D:\...\envs\lt_release012\Lib\site-packages\libertem\executor\dask.py", line 598, in run_function
    return future.result()
           ^^^^^^^^^^^^^^^
  File "D:\...\envs\lt_release012\Lib\site-packages\distributed\client.py", line 327, in result
    raise exc.with_traceback(tb)
  File "D:\...\envs\lt_release012\Lib\site-packages\libertem\io\fs.py", line 82, in get_fs_listing
    s = os.stat(full_path)
    ^^^^^^^^^^^^^^^^^
PermissionError: [Errno 13] Access is denied: 'C:\\Users\\Username.old'
```

matbryan52:
Fixes #1491 

Adding a test case for this is nontrivial, still thinking about it.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/io/fs.py
+++ b/src/libertem/io/fs.py
@@ -80,7 +80,7 @@ def get_fs_listing(path):
         full_path = os.path.join(path, name)
         try:
             s = os.stat(full_path)
-        except FileNotFoundError:
+        except (FileNotFoundError, PermissionError):
             # this can happen either because of a TOCTOU-like race condition
             # or for example for things like broken softlinks
             continue
--- a/tests/io/test_io_fs.py
+++ b/tests/io/test_io_fs.py
@@ -3,7 +3,7 @@ from unittest import mock
 
 import pytest
 
-from libertem.io.fs import get_fs_listing, FSError
+from libertem.io.fs import get_fs_listing
 
 
 def test_get_fs_listing_permission_of_subdir(tmpdir_factory):
@@ -20,11 +20,14 @@ def test_get_fs_listing_permission_of_subdir(tmpdir_factory):
 
     # patch os.stat to fail for `sub2`:
     with mock.patch('os.stat', side_effect=mock_stat):
-        get_fs_listing(tmpdir)
+        listing = get_fs_listing(tmpdir)
+
+        dir_names = [
+            entry['name']
+            for entry in listing['dirs']
+        ]
+        assert 'sub2' not in dir_names
 
         # no direct access to `sub2`
         with pytest.raises(PermissionError):
             os.stat(no_access_dir)
-
-        with pytest.raises(FSError):
-            get_fs_listing(no_access_dir)
--- a/tests/io/test_io_fs.py
+++ b/tests/io/test_io_fs.py
@@ -20,11 +20,11 @@ def test_get_fs_listing_permission_of_subdir(tmpdir_factory):
 
     # patch os.stat to fail for `sub2`:
     with mock.patch('os.stat', side_effect=mock_stat):
-        listing = get_fs_listing(tmpdir)
+        get_fs_listing(tmpdir)
 
         # no direct access to `sub2`
         with pytest.raises(PermissionError):
             os.stat(no_access_dir)
-    
+
         with pytest.raises(FSError):
             get_fs_listing(no_access_dir)
--- a/tests/io/test_io_fs.py
+++ b/tests/io/test_io_fs.py
@@ -1,0 +1,30 @@
+import os
+from unittest import mock
+
+import pytest
+
+from libertem.io.fs import get_fs_listing, FSError
+
+
+def test_get_fs_listing_permission_of_subdir(tmpdir_factory):
+    tmpdir = tmpdir_factory.mktemp('get_fs_listing')
+    tmpdir.mkdir('sub1')
+    no_access_dir = tmpdir.mkdir('sub2')
+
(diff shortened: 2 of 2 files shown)


sk1p:
> Adding a test case for this is nontrivial, still thinking about it.

If you don't come up with something more sensible, one can think about mocking this - such that if `os.stat` is called on a specific file during the test, the mock raises a `PermissionError`...

matbryan52:
> > Adding a test case for this is nontrivial, still thinking about it.
> 
> If you don't come up with something more sensible, one can think about mocking this - such that if `os.stat` is called on a specific file during the test, the mock raises a `PermissionError`...

Yeah, I had the same thought... 

Have been busy with the release notes, if you need to get this merged ASAP go ahead and push a mocked test, I have to run for an hour or two right now.

sk1p:
> I have to run for an hour or two right now.


Same here, might take a stab at it later in the evening

matbryan52:
I had a try at this using the `pyfakefs` module (https://pytest-pyfakefs.readthedocs.io/en/latest/)

I can create a directory which I can't stat by essentially removing all permissions from it and its parent, but a downside of this is it also can't be listed.

```python
@pytest.mark.parametrize("fs", [[None, [libertem.io.fs]]], indirect=True)
def test_no_permission_stat(fs):
    fs.os = OSType.WINDOWS
    path = pathlib.Path("C:\\foo\\bar")
    fs.create_dir(path, perm_bits=0o000)
    fs.chmod(path, 0o000, force_unix_mode=True)
    # Parent exists and is a directory
    assert path.parent.exists()
    assert path.parent.is_dir()
    # Cannot stat the path itself
    with pytest.raises(PermissionError):
        os.stat(path)
```

I am not sufficiently expert in ACLs to know if this is possible (and if the mock FS supports this edge case!).

sk1p:
> I am not sufficiently expert in ACLs to know if this is possible (and if the mock FS supports this edge case!).
> 
> The fact we can't list the parent directory means this can't work with the `get_fs_listing` function (in addition I can't seem to correctly have `pyfakefs` patch `os` calls within our functions, though it should be able to).

`pyfakefs` looks handy! But in this case, maybe it's easier to go the low-level way of creating a temporary directory, `os.listdir`ing that, and injecting the `PermissionError` by mocking `os.stat` more manually. I've quickly thrown together a small test case for that, I'll push that in a bit.

sk1p:
> Huh, the test case for some reason doesn't work as expected on Windows - `get_fs_listing` on the path itself for which we mock doesn't `raise` like it does on Linux - could you have a look if that is reproducible locally, @matbryan52?

@uellue tracked this down to a Python 3.7 issue - we use `Path(...).resolve()`, which doesn't have to call `os.stat`. So instead of checking for a `FSError`, we probably just check that the branch is taken, i.e. `sub2` is not part of the result.

sk1p:
Latest push should take care of the issue.

matbryan52:
Thanks for the test case, looks good to me.

LGTM!

PR #1498, opened 2023-08-15
Switch to vite for building the GUI

### LINKED ISSUE #1171: Update frond-end libraries
`create-react-app` v5.0.0 was just released, this means we may want to update the front end stack soon: https://github.com/facebook/create-react-app/releases/tag/v5.0.0

### LINKED ISSUE #1495: GUI dependency update
> * GUI dependency update with `npm install`

This will have to wait until after the release. It turns out that one of our main dependencies (`create-react-app`) hasn't seen an update in a while and looks abandoned upstream, so we need to move to something else. The react docs themselves mention next.js, but we don't really need a full framework, we only need the build system parts, like `create-react-app` provided.

The currently best option I see is to move to [vite](https://vitejs.dev/), which is advertised more or less as a (faster? easier to setup?) replacement for the webpack build stack we currently use. As far as I can see, that won't be the one with least code changes, but manually managing a webpack config (+ the rest of the build stack) would probably be more work over time.

_Originally posted by @sk1p in https://github.com/LiberTEM/LiberTEM/issues/1476#issuecomment-1677491938_

sk1p:
Closes #1495, closes #1171 (and probably makes dependabot happy about a few deps!)

This should also improve DX quite a bit for web GUI development, as we should now have:

- Hot module reloading - simple JSX/CSS/... changes should be visible almost instantly after saving
- A lot faster dev startup: should be almost instant
- Less dependencies: before:  1583 packages, 342MB; after: 382 packages, 266MB (should be noticeably faster on Windows - also a lot less small files in `node_modules`!)
- Smaller bundle size - need to double check that this is correct!

## TODO

- [x] Use vitest for testing
- [x] Update linter / linter config and fix new linting issues
- [x] Update all outdated packages (as far as possible)
- [x] Update contributor documentation (default port changed, what else?)

## Contributor Checklist:

## Reviewer Checklist:


sk1p:
Testing via `vitest` needed a few more dependencies, so the numbers mentioned above may have changed a bit. I'm also happy to report that the total runtime of the `node` GitHub Actions job shrunk from ~2m20 to ~1m - so the fast user experience is not only accomplished with smoke and mirrors (& caches), but is a real improvement in install/build/test times.

sk1p:
I think the coverage reporting is not correct - possibly it doesn't contain files that don't have any coverage at all? Needs investigating.

sk1p:
> I think the coverage reporting is not correct - possibly it doesn't contain files that don't have any coverage at all? Needs investigating.

Should be fixed now.

PR #1509, opened 2023-09-11
`UDF.get_method()` used to choose implementation in `UDFPartRunner._run_tile`

### LINKED ISSUE #1508: UDFs with multiple implementations / `UDF.get_method()` discrepancy
As the UDF interface is currently written, a UDF has two ways to signal which processing method it implements:

https://github.com/LiberTEM/LiberTEM/blob/d839573b2752b26738df051177aaa8c5f6f3d964/src/libertem/common/udf.py#L67-L68

which is mostly used during tileshape negotiation. The default implementation is this:

https://github.com/LiberTEM/LiberTEM/blob/d839573b2752b26738df051177aaa8c5f6f3d964/src/libertem/udf/base.py#L1099-L1107

but also implicitly by implementing one of the processing `Protocol` (`UDFFrameMixin`, `UDFTileMixin`, `UDFPartitionMixin`). This second mechanism is used at runtime via an `isinstance` check in `UDFPartRunner._run_tile` to actually choose which method to call:

https://github.com/LiberTEM/LiberTEM/blob/d839573b2752b26738df051177aaa8c5f6f3d964/src/libertem/udf/base.py#L2104-L2120

I was caught out by this when working on #1304 writing a UDF which implements both `process_tile` and `process_frame` - I had assumed that by overriding `get_method` I could influence which processing method would be used based on runtime information (`self.params`, dataset shape etc) . But in fact because we order the `isinstance` checks as `tile`, `frame`, `partition`, this UDF always runs as `UDFTileMixin` independently of what `get_method` returns.

This also links to #1506 where a UDF might have multiple implementations depending on the runtime context (for performance or algorithmic reasons).

I think we could find a non-breaking way to both reconcile the possible discrepancy between `get_method` and the runtime protocol check, and at the same time allow a UDF to influence how it is used at runtime.

One possibility is to actually use `UDF.get_method()` inside `UDFPartRunner._run_tile`, and move the type checking onto the default implementation of `get_method()` itself, something like this:

```python
    def get_method(self) -> Literal['tile', 'frame', 'partition']:
        if isinstance(self, UDFTileMixin):
            return 'tile'
        elif isinstance(self, UDFFrameMixin):
            return 'frame'
        elif isinstance(self, UDFPartitionMixin):
            return 'partition'
        else:
            raise TypeError("UDF should implement one of the `process_*` methods")
```

then do a string comparison (or better still put our processing methods into an `Enum`) inside `UDFPartRunner._run_tile`.

We could also memoize the processing method somewhere during initialization, which would reduce the overhead from a lot of `isinstance` calls (small but in a large dataset with lots of tiles it could add up!).

There might be side-effects I haven't thought of, however! If not, I'll implement and use as the base for #1304.

matbryan52:
Closes #1508 

- Shifts the string identifiers "frame", "tile", "partition" onto `libertem.common.udf.UDFMethodEnum`
- Default `UDF.get_method()` performs the runtime protocol check previously in `UDFPartRunner._run_tile` and returns an enum member that `UDFPartRunner` matches against
- Additional checking added to verify `UDF.get_method()` return is valid (both is an enum member, and actually implements the desired Protocol)
- Additional test cases for invalid `UDF.get_method()` implementations

Notes:
- Returning an enum member from `get_method` is technically a breaking change, because `UDFMethodEnum.FRAME == 'frame'` is `False`, if we want to be completely backwards compatible we would need to use a `StrEnum` (either from the PyPi package or as a backport from 3.11). I think the API is/was not public though so no real requirement to do this!
- I had to add explicit `assert isinstance(udf, UDFFrameMixin)` inside `_run_tile` for each method in order to make `mypy` happy, this is just additional runtime overhead!
- The current implementation (nor the previous implementation) doesn't store the processing mode that the UDF wants to use, which adds a little bit of overhead, but also more importantly doesn't guarantee the processing mode can't change in the time between tiling scheme negotiation and actually processing tiles. On the one hand this would be a 'strange' thing for a UDF to do, but could be a hard to track down bug.
- I haven't added any documentation yet, would have to decide if this should be public facing?

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/common/udf.py
+++ b/src/libertem/common/udf.py
@@ -25,6 +25,12 @@ class TilingPreferences(TypedDict):
     total_size: Union[float, int]
 
 
+class UDFMethodEnum(Enum):
+    TILE = 'tile'
+    FRAME = 'frame'
+    PARTITION = 'partition'
+
+
 class UDFProtocol(Protocol):
     '''
     Parts of the UDF interface required for MIT code in LiberTEM
@@ -64,7 +70,7 @@ class UDFProtocol(Protocol):
     ND_BACKENDS = ND_BACKENDS  #: Set of backends that support n-dimensional arrays
     D2_BACKENDS = D2_BACKENDS  #: Set of backends that only support two-dimensional arrays
 
-    def get_method() -> Literal['tile', 'frame', 'partition']:
+    def get_method() -> Literal[UDFMethodEnum.TILE, UDFMethodEnum.FRAME, UDFMethodEnum.PARTITION]:
         raise NotImplementedError()
 
     def get_tiling_preferences() -> TilingPreferences:
--- a/src/libertem/udf/base.py
+++ b/src/libertem/udf/base.py
@@ -31,7 +31,7 @@ from libertem.common.buffers import (
     BufferKind, BufferUse, BufferLocation,
 )
 from libertem.common import Shape, Slice
-from libertem.common.udf import TilingPreferences, UDFProtocol
+from libertem.common.udf import TilingPreferences, UDFProtocol, UDFMethodEnum
 from libertem.common.math import prod
 from libertem.io.dataset.base import (
     TilingScheme, Negotiator, Partition, DataSet, get_coordinates
@@ -1096,13 +1096,17 @@ class UDFBase(UDFProtocol):
         else:
             raise ValueError(f"Backend can be {BACKENDS}, got {self._backend}")
 
-    def get_method(self) -> Literal['tile', 'frame', 'partition']:
-        if hasattr(self, 'process_tile'):
-            return 'tile'
-        elif hasattr(self, 'process_frame'):
-            return 'frame'
-        elif hasattr(self, 'process_partition'):
-            return 'partition'
+    def get_method(self) -> Literal[
+        UDFMethodEnum.TILE,
+        UDFMethodEnum.FRAME,
+        UDFMethodEnum.PARTITION
+    ]:
+        if isinstance(self, UDFTileMixin):
+            return UDFMethodEnum.TILE
+        elif isinstance(self, UDFFrameMixin):
+            return UDFMethodEnum.FRAME
+        elif isinstance(self, UDFPartitionMixin):
+            return UDFMethodEnum.PARTITION
         else:
             raise TypeError("UDF should implement one of the `process_*` methods")
 
--- a/src/libertem/udf/base.py
+++ b/src/libertem/udf/base.py
@@ -2117,12 +2117,15 @@ class UDFPartRunner:
         roi: Optional[np.ndarray],
     ) -> None:
         for udf in udfs:
-            if udf.get_method() == UDFMethodEnum.TILE:
+            udf_method = udf.get_method()
+            if udf_method == UDFMethodEnum.TILE:
+                assert isinstance(udf, UDFTileMixin)  # mypy!
                 udf.set_contiguous_views_for_tile(partition, tile)
                 udf.set_slice(tile.tile_slice)
                 udf.set_tile_idx(tile.scheme_idx)
                 udf.process_tile(device_tile)
-            elif udf.get_method() == UDFMethodEnum.FRAME:
+            elif udf_method == UDFMethodEnum.FRAME:
+                assert isinstance(udf, UDFFrameMixin)  # mypy!
                 tile_slice = tile.tile_slice
                 for frame_idx, frame in enumerate(device_tile):
                     frame_slice = Slice(
@@ -2143,7 +2146,8 @@ class UDFPartRunner:
                     udf.set_slice(frame_slice)
                     udf.set_views_for_frame(partition, tile, frame_idx)
                     udf.process_frame(frame)
-            elif udf.get_method() == UDFMethodEnum.PARTITION:
+            elif udf_method == UDFMethodEnum.PARTITION:
+                assert isinstance(udf, UDFPartitionMixin)  # mypy!
                 # Internal checks for dataset consistency
                 assert partition.slice.adjust_for_roi(roi) == tile.tile_slice
                 udf.set_views_for_tile(partition, tile)
--- a/src/libertem/udf/base.py
+++ b/src/libertem/udf/base.py
@@ -2094,7 +2094,8 @@ class UDFPartRunner:
             for udf in udfs:
                 udf.set_backend(backend)
                 udf.set_meta(meta)
-                # validate that one of the `process_*` methods is implemented
+                # validate that get_method returns a valid method type
+                # the presence of the method was checked on the main node
                 if udf.get_method() not in tuple(UDFMethodEnum):
                     raise UDFException('UDF.get_method() returned unrecognized')
                 udf.set_slice(pslice_for_roi)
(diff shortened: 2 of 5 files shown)


src/libertem/udf/base.py
```diff
@@ -2111,12 +2119,15 @@ def _run_tile(
         roi: Optional[np.ndarray],
     ) -> None:
         for udf in udfs:
-            if isinstance(udf, UDFTileMixin):
+            udf_method = udf.get_method()
+            if udf_method == UDFMethodEnum.TILE:
+                assert isinstance(udf, UDFTileMixin)  # mypy!
```
sk1p, line 2124: As an alternative, we can do something like this:

```python
                udf_tiled = typing.cast(UDFTileMixin, udf)
                udf_tiled.process_tile(device_tile)
```

(keeping the other method accesses like `udf.set_slice` as they are)

`cast` should be the identity function at runtime. Not the most elegant way (we don't actually check the type, we just say: trust me, basically), but maybe something to consider if we want to keep overheads low.

src/libertem/udf/base.py
```diff
@@ -2137,12 +2148,17 @@ def _run_tile(
...
                     udf.process_frame(frame)
-            elif isinstance(udf, UDFPartitionMixin):
+            elif udf_method == UDFMethodEnum.PARTITION:
+                assert isinstance(udf, UDFPartitionMixin)  # mypy!
                 # Internal checks for dataset consistency
                 assert partition.slice.adjust_for_roi(roi) == tile.tile_slice
                 udf.set_views_for_tile(partition, tile)
                 udf.set_slice(tile.tile_slice)
                 udf.process_partition(device_tile)
+            else:
```
sk1p, line 2158:
```suggestion
            else:  # pragma: no cover
```

src/libertem/common/udf.py
```diff
@@ -25,6 +25,12 @@ class TilingPreferences(TypedDict):
     total_size: Union[float, int]
 
 
+class UDFMethodEnum(Enum):
```
sk1p, line 28: Naming nit: how about just `UDFMethod`?

If we decide to make this a public interface, we may want to put these constants into the `UDF` class namespaces, too, like we have for array backend selection - this should reduce the number of imports needed for writing UDFs.

sk1p (changes requested):
> I think the API is/was not public though so no real requirement to do this!

Indeed, I'm fine with changing the type, no need for `StrEnum`.

> I had to add explicit assert isinstance(udf, UDFFrameMixin) inside _run_tile for each method in order to make mypy happy, this is just additional runtime overhead!

See the inline comment - the `assert` is the safest way, but as this is very close to the inner loop, where overheads do matter quite a bit, maybe a `cast` is justified here?

> I haven't added any documentation yet, would have to decide if this should be public facing?

Indeed. If this is needed for #1304 or similar UDFs, it has to become a public interface - otherwise we have a two-class system (like Apple with some of their system APIs, which can't be used by third-party apps...) and it becomes hard to reason what is actually part of the public API.

I guess this needs a small paragraph in the advanced UDF docs, and maybe some changes in the current "precedence" docs?

src/libertem/common/udf.py
matbryan52: OK by me, originally what I wrote but I was following the style of the other `Enums` in the file:

https://github.com/LiberTEM/LiberTEM/blob/675c98afd6101ea54d2c5e978d9d163ea93963b3/src/libertem/common/udf.py#L15-L20

sk1p: > OK by me, originally what I wrote but I was following the style of the other `Enums` in the file:
> 
> https://github.com/LiberTEM/LiberTEM/blob/675c98afd6101ea54d2c5e978d9d163ea93963b3/src/libertem/common/udf.py#L15-L20

I think they have the suffix because depth / size are either an integer, or a marker value from the `Enum`.

src/libertem/udf/base.py
matbryan52: That's a good solution, thanks for the suggestion. (I didn't know about `cast`).

On the subject of overheads, we could remove one `isinstance` call per tile (in the default/standard case) by moving the `get_method()` call one level up into _`run_udfs` and reusing the result for each `_run_tile`. Do you want me to do that, I think it would be consequence-free, or is it not worth the bother?

sk1p: > On the subject of overheads, we could remove one isinstance call per-UDF per-tile

Sounds good to me! Also good if people do whatever in their `get_method` impl...

matbryan52: It's not elegant, but it's done - could probably be merged with the execution_plan or better still frozen during `_prepare_run_for_dataset` : https://github.com/LiberTEM/LiberTEM/pull/1509/commits/ccc9f3af7416e52a5725fe1d803019fb5714a93b

tests/udf/test_simple_udf.py
```diff
@@ -770,3 +772,70 @@ def test_previous_attrs(lt_ctx, cls):
...
+    [
+        42,
+        UDFMethod.FRAME,
+        UDFMethod.PARTITION,
+        UDFMethod.TILE,
+    ]
+)
+def test_bad_get_method(lt_ctx, method):
+    ds = lt_ctx.load('memory', data=np.ones((2, 2, 4, 4)))
+    with pytest.raises(UDFException):
```
sk1p, line 796: I was a bit confused by this test case, maybe adding a comment here helps:

```suggestion
    # `UDFException` is thrown either for the invalid method `42`, or
    # because the UDF doesn't actually implement the matching `process_*` methods:
    with pytest.raises(UDFException):
```

docs/source/reference/udf.rst
```diff
@@ -29,6 +29,8 @@ Base UDF classes
     :members:
     :inherited-members:
 
+.. autoenum:: libertem.common.udf.UDFMethod
```
sk1p, line 32: I'm not sure what's the reason, but the `UDFMethod` enum is somehow not properly rendered in the return type position:

image

Just a note, and probably a bug in `enum-tools`?

matbryan52: It looks like `Literal[Enum.member]` is generally problematic : https://github.com/sphinx-doc/sphinx/issues/11473

What I've done is create `UDFMethodT` alongside the enum, which is used to type annotate where necessary. This is slightly less useful in the doc but at least renders correctly, and there is always a link to the `UDFMethod` doc itself within the docstring.

sk1p: Ok, another question: why do we have `Literal[TheEnum.Member1, TheEnum.Member2, ...]` instead of just `TheEnum` as type? That should only be needed if we somehow only support a subset of variants of the `enum`, right? Or is there another reason?

matbryan52: I was under the impression that that `def fn(...)-> MyEnum:` implies we return (an instance of) the Enum itself, not a member of the Enum (e.g. https://stackoverflow.com/a/67292548). I will see if I can find an official bit of doc...

sk1p: This is the documentation I was looking at: https://mypy.readthedocs.io/en/stable/literal_types.html#enums

This typechecks:

```python
from enum import Enum


class MyEnum(Enum):
    VariantOne = 1
    VariantTwo = 2


def f(my: MyEnum) -> MyEnum:
    return MyEnum.VariantTwo


f(MyEnum.VariantTwo)
```

I understood the Stackoverflow question as being about the actual string literals, like if we wanted to support strings like `"frame"` as return values

matbryan52: Clearly I misunderstood, I've switched it to returning `UDFMethod` and removed the alias. Thanks!

sk1p:
CI failure is unrelated, I'm re-triggering failed jobs now

sk1p (approved):
Looks good! CI is still failing, somewhere in `pyxem` → `pyFAI` → `numexpr`. As far as I can see, a new `numexpr` release should come soon (they are aware of the breakage), so I'm inclined to just go ahead and merge now.

matbryan52:
> Looks good! CI is still failing, somewhere in `pyxem` → `pyFAI` → `numexpr`. As far as I can see, a new `numexpr` release should come soon (they are aware of the breakage), so I'm inclined to just go ahead and merge now.

I did also see some logs like "No space left on device" during tests on failed CI runs ?

Thanks for all the reviews and feedback!

sk1p:
> > Looks good! CI is still failing, somewhere in `pyxem` → `pyFAI` → `numexpr`. As far as I can see, a new `numexpr` release should come soon (they are aware of the breakage), so I'm inclined to just go ahead and merge now.
> 
> I did also see some logs like "No space left on device" during tests on failed CI runs ?

Yeah, that's a curious issue with our ci setup - after some time, /dev/shm in some containers runs out of space - didn't manage to 	find time to track down the reason for that, and re-creating the containers is a quick workaround...

> Thanks for all the reviews and feedback!

You're welcome!

Issue #1513, opened 2023-09-20
Fix GUI static files path


sk1p:
Since #1498 the GUI build included in the repo is broken - it tries to load assets from paths that are different than before. Needs to be investigated and fixed, either in the vite config or by serving static files from different paths.

PR #1514, opened 2023-09-21
Add progress bar in GUI

### LINKED ISSUE #1500: GUI: integrate new progress reporter as progress bars
- In the UDF running engine of the web API, use a custom progress reporter
- Forward events from the progress reporter via the websocket connection
- Display progress bar in the web GUI, in some form or other
    - don't flash progress bar if operation finishes "fast" (under some threshold, let's say 200ms?)
    - ...but do show progress bar even before the first result comes in, which is currently a bit confusing

sk1p:
It's a combined progress bar for all currently running jobs (which also means it may jump backwards if one of the jobs finishes...)

Closes #1500 

(note that #1513 needs to be fixed before rebuilding the client, which will be done in a separate PR)

## Contributor Checklist:

## Reviewer Checklist:


matbryan52:
Great! I had a look over the code (for as much as I can follow the TS), and seems like it wasn't too hard to integrate with the `ProgressReporter`. Looking forward to seeing it running once the support PRs are merged.

sk1p:
> Great! I had a look over the code (for as much as I can follow the TS), and seems like it wasn't too hard to integrate with the `ProgressReporter`. Looking forward to seeing it running once the support PRs are merged.

Thanks for having a look! I enjoyed writing TypeScript after a while, the type system is really quite enjoyable.

I think the main gotcha was that the callbacks in the `ProgressReporter` are not running on the main thread, so we have to do some explicit gymnastics to dispatch the websocket message on the right thread and message loop. Once that was solved it was only a matter of state mgmt and viewing progress in the GUI.

sk1p:
> Looking forward to seeing it running once the support PRs are merged.

It's now rebased on top of the changes merged with #1513, and I've pushed a rebuilt client. Should now be testable without running vite.

client/src/progress/components/ProgressPie.tsx
matbryan52, line 1: I can't figure out if `ProgressPie` is used anywhere? If not then maybe we can remove it for now?

matbryan52 (approved):
I like it ! I can't really comment on the TypeScript, but it certainly seems consistent with my understanding of the rest of the code.

While testing I did notice that I think we have a problem with the `needs_rerun` mechanism - it seems to work for the Disk analysis but Ring and Point, for example, always re-calculate (for me). This is happening on current master, too, so I think it's not related to the progress bar.

client/src/progress/components/ProgressPie.tsx
sk1p: Good catch, that was from my first go at it - but only having the Pie is too subtle. I'll clean this up.

sk1p:
Thanks for having a look!

> I like it ! I can't really comment on the TypeScript, but it certainly seems consistent with my understanding of the rest of the code.

Yeah, review regarding the functionality is more important than reviewing the Typescript impl in this case ;)

> While testing I did notice that I think we have a problem with the `needs_rerun` mechanism - it seems to work for the Disk analysis but Ring and Point, for example, always re-calculate (for me). This is happening on current master, too, so I think it's not related to the progress bar.

Interesting - I'll have a look if I can reproduce this - maybe there's something inconsistent in how parameters are sent over or compared? I don't remember how I implemented this, so I'll need to have a look at the code.

PR #1520, opened 2023-10-02
Converter for `(sig, nav)` DM4 files

matbryan52:
Adds converter for `(sig, nav)` DM4 files to `(nav, sig)` NPY files so that these can be processed using LiberTEM.

Add a single user-facing function at `libertem.contrib.convert_transposed.convert_dm4_transposed`.

Uses `process_partition` and a modified `RecordUDF` to load blocks of the file, generally avoiding the multiple-traversal / page fault issue explored in #1328.

This  converts the data to a numpy file, but I think it would be possible to extend it to create a new, c-ordered DM4 file with the same tags / structure + the special flag to denote a c-ordered file.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/tests/io/test_convert_transposed.py
+++ b/tests/io/test_convert_transposed.py
@@ -1,5 +1,6 @@
 import pathlib
 import numpy as np
+import pytest
 
 from libertem.udf.raw import PickUDF
 from libertem.udf.sum import SumUDF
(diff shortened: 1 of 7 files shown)


tests/io/test_convert_transposed.py
```diff
@@ -0,0 +1,80 @@
...
+    assert ds_npy.shape.to_tuple() == converted_data.shape
+
+    for _ in range(3):
+        check_slice = tuple(
+            np.random.randint(0, n) for n in nav_shape
+        )
+        roi = np.zeros(ds_npy.shape.nav, dtype=bool)
+        roi[check_slice] = True
+        frame = lt_ctx.run_udf(ds_npy, PickUDF(), roi=roi)['intensity'].data.squeeze()
+        assert np.allclose(frame, converted_data[check_slice])
```
sk1p, line 63: I wonder, the test data used here is small enough, maybe we can have an `allclose(converted_data, load_numpy(convert_path))` here, instead or in addition?

sk1p (approved):
LGTM! Small comment on the test case :point_down:

tests/io/test_convert_transposed.py
matbryan52: Good idea, checking the whole data covers every possible analysis we could test with so it makes sense to just do that. Thanks!

PR #1550, opened 2023-11-09
Try harder to cancel the progress bar on cancellation

### LINKED ISSUE #1548: Progress bar doesn't reset when dataset is closed with task in flight
## How to reproduce

* Open a dataset and start an analysis with non-trivial runtime.
* Close the dataset while processing is still going on and the progress bar is shown.
* Open the dataset again and process again.

## What happens

* The progress bar gets stuck at the point where the dataset was closed.
* The progress bar moves forward with the new processing, but then resets to the previous state once the new processing finishes.

## What should happen

The pending tasks from a dataset are removed from the progress bar when a dataset is closed.

sk1p:
It can happen that a progress report comes in after a job has been successfully cancelled - so we can't rely on removing the progress state only on finishing events, we need to listen to the explicit 'end' progress hook.

This updates the web API to emit specific progress events.

Fixes #1548

## Contributor Checklist:

## Reviewer Checklist:


sk1p:
docs-check failure is the numpy intersphinx failing once again...

uellue:
Thx, fix confirmed, merging!

PR #1561, opened 2023-11-23
Optimise `MIBDataSet` to read headers in parallel + respect ulimit NOFILE

matbryan52:
First stab at optimizing the header-reading code of `MIBDataSet` to use a `ThreadPoolExecutor` when reading a 'large' number of files, particularly useful when opening a dataset of many small files over the network.

The number of workers for this executor is quite sensitive, and I need to do more testing. For now it's limited to `2` on linux and `0.5 * num_cpu` on Windows based on what I've seen so far. I'll try to do some benchmarking tomorrow (made a bit more tricky because of filesystem caching / remote storage server caching).

Also adds a specialised `get_num_partitions` which tries to guess a `num_partitions` value which won't break the `NOFILE` limit for a process (assuming we are running in a process-based executor) on linux (only). There's a lot to improve in this part, but it already has no impact whatsoever on a 'well-acquired' dataset, and lets you actually process a `single-file-per-frame` dataset on linux, which is already a win.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/io/dataset/mib.py
+++ b/src/libertem/io/dataset/mib.py
@@ -1173,11 +1173,11 @@ class MIBDataSet(DataSet):
         filesizes = {}
         for entry in os.scandir(os.path.dirname(self._path)):
             if entry.is_file() and entry.path.endswith('.mib'):
-                filesizes[entry.path] = entry.stat().st_size
+                filesizes[str(entry.path)] = entry.stat().st_size
 
         res = {}
         for path, header in zip(fnames, header_bytes):
-            res[path] = MIBHeaderReader._parse_header_bytes(header, filesizes[path])
+            res[path] = MIBHeaderReader._parse_header_bytes(header, filesizes[str(path)])
         return res
 
     def _filenames(self):
--- a/src/libertem/io/dataset/mib.py
+++ b/src/libertem/io/dataset/mib.py
@@ -1196,7 +1196,9 @@ class MIBDataSet(DataSet):
         if num_files > 16384:
             warnings.warn(
                 f"Saving data in many small files (here: {num_files}) is not efficient, "
-                "please increase the \"Images Per File\" parameter when acquiring data",
+                "please increase the \"Images Per File\" parameter when acquiring data. If ",
+                "this leads to \"Too many open files\" errors, on POSIX systems you can increase ",
+                "the maximum open files limit using the \"ulimit\" shell command.",
                 RuntimeWarning
             )
         self._filename_cache = fns
--- a/src/libertem/io/dataset/mib.py
+++ b/src/libertem/io/dataset/mib.py
@@ -779,13 +779,16 @@ class MIBHeaderReader:
             raise NotImplementedError(f"unknown dtype: {dtype}")
 
     def read_header(self) -> HeaderDict:
-        # FIXME: do this read via the IO backend!
-        with open(file=self.path, encoding="ascii", errors='ignore') as f:
-            header = f.read(1024)
-            filesize = os.fstat(f.fileno()).st_size
-
+        header, filesize = self._read_header_bytes(self.path)
         self._fields = self._parse_header_bytes(header, filesize)
         return self._fields
+
+    @staticmethod
+    def _read_header_bytes(path) -> Tuple[str, int]:
+        # FIXME: do this read via the IO backend!
+        with open(file=path, encoding="ascii", errors='ignore') as f:
+            filesize = os.fstat(f.fileno()).st_size
+            return f.read(1024), filesize
 
     @staticmethod
     def _parse_header_bytes(header: str, filesize: int) -> HeaderDict:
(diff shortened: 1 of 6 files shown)


matbryan52:
A few numbers, accessing a dataset of 65k small `.mib` files (65 KB each). Just running `ctx.load('mib', path='path/to/img.mib')`

From a Windows server (36 cores, 20 Gbit/s network connection):

- to a fairly slow SMB share:
    - This branch: ~15 seconds
    - Current master: ~3 min 10 seconds

- to a fast-ish NFS share:
    - This branch: ~15 seconds
    - Current master: ~3 min 20 seconds

- to local disk:

    - This branch: ~10 seconds
    - Current master: ~20 seconds

From a Linux server (64 cores, 10 Gbit/s network connection)

- to a fast-ish NFS share:

    - This branch: ~15-25 seconds
    - Current master: 40 seconds

- to the SMB share mounted via `gvfs` (which is definitely the bottleneck here! but the best I have):

    - This branch: 18 min 45 seconds 😱 
    - Current master: Longer than I was willing to wait

- to local disk:

    - This branch: ~3-4 seconds
    - Current master: 2 seconds

There should be minimal-to-no-change for a dataset with less then 512 files as I don't even create a thread pool in this case. We could bump this number higher if you think it's too low.

[Aside] : In the Linux case, current master, `run_udf` immediately crashes due to `OSError("Too many open files")`.

I'm working on getting a dataset like this to add to our test suite.

src/libertem/io/dataset/mib.py
```diff
@@ -1143,33 +1167,49 @@ def detect_params(cls, path, executor):
...
         if self._filename_cache is not None:
             return self._filename_cache
         fns = get_filenames(self._path, disable_glob=self._disable_glob)
-        if len(fns) > 16384:
+        num_files = len(fns)
+        if num_files > 16384:
             warnings.warn(
-                "Saving data in many small files (here: %d) is not efficient, please increase "
-                "the \"Images Per File\" parameter when acquiring data",
+                f"Saving data in many small files (here: {num_files}) is not efficient, "
```
sk1p, line 1198: Good catch!

src/libertem/io/dataset/mib.py
```diff
@@ -1230,6 +1270,29 @@ def get_base_shape(self, roi: Optional[np.ndarray]) -> Tuple[int, ...]:
         assert self.meta is not None and base_shape[-1] == self.meta.shape[-1]
         return base_shape
 
+    def get_num_partitions(self) -> int:
+        # Try to limit the partition to some fraction of the max_open files limit
+        limit_n_files = float('inf')
+        if platform.system() != "Windows":
+            import resource
+            # Windows has no problem opening huge numbers of files
+            limit_n_files, _ = resource.getrlimit(resource.RLIMIT_NOFILE)
```
sk1p, line 1279: Maybe we can subtract the already open files here? Should be accessible via `getrusage(resource.RUSAGE_SELF)`

Additional comment: on Linux, one of the main reasons to limit the number of open file descriptors is the `select` syscall, which we should not be using anyways. The hard limit of file descriptors is a lot higher than 1024, for example on my system:

image

1048576 files should be enough for everyone™ (another common value seems to be 262144 on CentOS machines; on older versions though the hard limit is often indeed smaller, I've seen something like 4096). So, what we could do, is increase the soft limit to be equal to the hard limit. This should be done "globally" in our worker processes, not as a per-dataset-type fix. We can include something like this in the worker setup code (dask, pipelined):

```python
_, hard_lim = resource.getrlimit(resource.RLIMIT_NOFILE)
resource.setrlimit(resource.RLIMIT_NOFILE, (hard_lim, hard_lim))
```

(aside: this is also something we should include in our testing setup, where I sometimes manually set the `ulimit` before running `pytest`)

For threaded workers, it becomes a bit harder, as we don't "own" the Python process itself, and might cause issues if we modify the process-wide limit. Do you have an opinion on this? Or should we assume that `select` usage is rare, and work on an opt-out basis?

If setting the rlimits is not enough (for systems with a low hard limit), instead of a per-dataset-type fix here, we should fix the backends to limit the number of concurrently open files instead - then you also don't need a separate fix for the raw `DataSet`.

sk1p (changes requested):
I like the "parallel stat and header reading" changes - for the "number of files per partition" thing I would prefer a more global fix, meaning 1) increasing the soft limit as far as possible, and 2) a fix in the I/O backend such that it only opens a limited number of files at once.

src/libertem/io/dataset/mib.py
matbryan52: Drive-by-fix, has always bugged me :D

src/libertem/io/dataset/mib.py
matbryan52: > So, what we could do, is increase the soft limit to be equal to the hard limit. 

I thought about this, but wasn't sure of what consequences it might create. If you're happy with it as an approach then I'll implement that instead.

> For threaded workers, it becomes a bit harder, as we don't "own" the Python process itself, and might cause issues if we modify the process-wide limit. Do you have an opinion on this?

I think that using processes is probably 90%+ of real usage so we're making an improvement. I don't know enough about the implications of a higher limit for the main Python process to have a strong opinion! I'll take a look at the implementation and see if we can even implement something global with the currently accessible info (I'm not sure if a dataset can know enough about the execution context at the moment that we call `get_partitions` ? Especially if we're running on an external cluster or on multiple-nodes etc etc).

matbryan52: The way I wrote the test this assumes the test files are in `TESTDATA_BASE_PATH`. It takes about 0.5 seconds to save the 16k files on Ptycho (cold, in isolation) so I could convert this fake datset into a fixture if you prefer? Would mean we check the file limit on a wider variety of configurations, too.

sk1p:
Huh, now _Windows_ is unhappy about the number of open files... curious. Maybe there is a limit in CI that's not present on "normal" systems?

matbryan52:
> Huh, now _Windows_ is unhappy about the number of open files... curious. Maybe there is a limit in CI that's not present on "normal" systems?

Yeah... I have a fear this trying to 'test' this, either for the fake raw DS or real MIBs, might be a flakiness nightmare ! (without resorting to some kind of mocking).

PR #1572, opened 2024-01-22
Suspend executor of `libertem-server` after period of inactivity

### LINKED ISSUE #1570: `libertem-server`: Shut down cluster if inactive
If LiberTEM is deployed as a long-running service, the workers and scheduler take up CPU cycles and memory when idle. This is an issue on shared machines, right now on Moellenstedt.

For that reason, the cluster should be shut down after a period of inactivity and restarted on demand.

sk1p:
Yaks shaved:

* Removed a lot of event loop trickery in the test setup, which should not be necessary anymore with the current version of `pytest-asyncio`.
* ... and added back different event loop trickery - partly because on Python 3.7 we have to use an older version of `pytest-asyncio`... and because Windows™
* Explicitly close the `Context` used in `libertem-server` (and a few other resources used), which should improve the shutdown experience
* Pinned `pyxem`, see #1574 

Closes #1570

## TODO

- [x] Verify that the event loop changes don't have unintended side-effects
- [x] Do some real-world testing, too
- [x] Allow setting a custom snooze timeout (CLI? Environment variable?)
- [x] Verify that all relevant activity does indeed reset the last activity timer
- [x] ~~If a browser tab is kept open, should the executor be kept alive as long as that tab is alive? Should there be a kind of "keep-alive" from the client?~~ I decided against this, as re-starting the executor is actually fast enough (takes less than a second on our fast workstation), so counting UDF runs as activity is sufficient. => revisit for #1575?
- [x] Configure snooze time in minutes instead of seconds
- [x] Documentation improvements: can take a while to `unsnooze`
- [ ] Revisit event loop blocking functions
  - [ ] `unsnooze` -> `get_executor`: slow on some machines, so it might need to be `async` (or can it run in a thread?)
  - [ ] `shutdown`: closing the executor could take a while?
  - [ ] `make_executor`

## Contributor Checklist:

## Reviewer Checklist:


sk1p:
@uellue @matbryan52 this is now ready for review; if one of you has some time to spare, I'd be grateful!

matbryan52:
> @uellue @matbryan52 this is now ready for review; if one of you has some time to spare, I'd be grateful!

Can do, should be able to look tomorrow morning.

uellue (changes requested):
Generally, seems to work on Win11, Python 3.11! Thx for implementing. Here a few observations:

* CuPy workers seem to JIT-compile sth in my case on the first run. It has to re-compile after unsnoozing.
* Unsnoozing takes 22 s on my test system, which feels like an eternity given the usual responsiveness. The GUI shows no response except the spinner on the button that was pressed. It would help if the progress bar on top would show in-flight tasks, since that usually appears right after pressing a button. Implementing #1575 would be best to improve the user experience.
* The cluster is also snoozed if a user is using the GUI. Can GUI activity be counted to reset the snooze timer? This would be useful for short snooze intervals, but probably not relevant for now.

The cluster had some issues to shut down when pressing Ctrl-C multiple times. It terminated eventually, but it took quite some time. Might be unrelated.

Discussion with @sk1p: The startup and warmup on Linux is considerably faster, perhaps because it uses `fork` instead of `spawn`? 

I'd say it can be used with snooze timeouts in the region of 30 min where it will rarely trigger while a user is active. Setting the parameter in seconds will encourage users to use shorter timeouts, though.

For the time being we can perhaps include more info in the documentation, for example the `--help` result of `libertem-server`?

* Recommended time between 600 s (10 min)  and 2000 s (33 min).
* Unsnoozing may take in the region of 30 s, in particular on Windows.

sk1p:
> * CuPy workers seem to JIT-compile sth in my case on the first run. It has to re-compile after unsnoozing.

Aren't the JIT results cached on disk? [They should be...](https://docs.cupy.dev/en/stable/user_guide/performance.html#kernel-compilation)

> * Unsnoozing takes 22 s on my test system, which feels like an eternity given the usual responsiveness. The GUI shows no response except the spinner on the button that was pressed. It would help if the progress bar on top would show in-flight tasks, since that usually appears right after pressing a button. Implementing [GUI: show cluster snooze status  #1575](https://github.com/LiberTEM/LiberTEM/issues/1575) would be best to improve the user experience.

Yup, the usual Windows slowness of spawning new processes, probably re-importing "the world", etc... agreed that the snooze status GUI change would be best, I would not change the progress bar behavior for this (it's currently mostly server-driven, and there are no tasks in flight yet when unsnoozing...)

> * The cluster is also snoozed if a user is using the GUI. Can GUI activity be counted to reset the snooze timer? This would be useful for short snooze intervals, but probably not relevant for now.

This could be possible, and easy or hard depending on how you define GUI activity. 

* "Tab is open" - this is easy, we just have a client-side loop and send a keep-alive message to the server every so often
* "The user is doing _something_ (moving the mouse, scrolling, ...)" - this is harder, as [the API built for this](https://developer.mozilla.org/en-US/docs/Web/API/IdleDetector) is experimental and non-portable...
* "The user is doing an action that affects the GUI state" - for example, changing parameters, opening data sets, ... - easy to implement, but covers less activity.

All of these would then have to send keep-alive messages to the server, debounced such that we don't spam, like one message for each mouse movement...

Would you agree that this could happen as part of the other GUI changes in #1575?

> The cluster had some issues to shut down when pressing Ctrl-C multiple times. It terminated eventually, but it took quite some time. Might be unrelated.

Yeah, I'm a bit unhappy how shutdown is handled right now, especially in the non-obvious cases (example: while the executor is starting up), and how this interacts with ctrl+c handling of the dask workers...

> I'd say it can be used with snooze timeouts in the region of 30 min where it will rarely trigger while a user is active. Setting the parameter in seconds will encourage users to use shorter timeouts, though.

I think we can do two things: 1) configure the snooze timeout in minutes, and 2) document recommended values.

sk1p:
> Can confirm the same on a Windows system with 36 workers. Upwards of a minute to come back from snooze (without GPUs involved). That said our system might be badly configured...

Oof.

matbryan52:
> Yeah, I'm a bit unhappy how shutdown is handled right now, especially in the non-obvious cases (example: while the executor is starting up), and how this interacts with ctrl+c handling of the dask workers...

Confirm this also. When shutting down during unsnoozing the UI hangs on the shutdown modal and the command line appears to show the new 'nanny' process bringing all the workers back up again, The outcome might depend on exact timing.

matbryan52:
An idea to handle slow startup issues in general, not for this PR though as it is quite a change - we could support scaling the number of workers a bit dynamically  : rather than waiting for the full cluster just create the nanny and one worker, and start responding to the user (or a run UDF job) while bringing other workers up into the pool. The architecture of UDFTask dispatch would have to change I think but we could look into it.

sk1p:
> An idea to handle slow startup issues in general, not for this PR though as it is quite a change - we could support scaling the number of workers a bit dynamically : rather than waiting for the full cluster just create the nanny and one worker,

Indeed, I was looking at the code that starts the `SpecCluster`, and wondered if it couldn't be a bit more lazy (i.e. only wait for the first worker, instead of `len(spec)`...

>  and start responding to the user (or a run UDF job) while bringing other workers up into the pool. The architecture of UDFTask dispatch would have to change I think but we could look into it.

Yeah - maybe we can make the code simpler and improve this issue at the same time - if we are less rigid in our mapping of tasks to workers, for example.

I was also thinking about "hitting the ground running" using a threaded executor at the beginning, and then switch over to a different one, but that can be a bit surprising (maybe acceptable for the web GUI?)

uellue:
> An idea to handle slow startup issues in general, not for this PR though as it is quite a change - we could support scaling the number of workers a bit dynamically : rather than waiting for the full cluster just create the nanny and one worker, and start responding to the user (or a run UDF job) while bringing other workers up into the pool. The architecture of UDFTask dispatch would have to change I think but we could look into it.

That's a good idea! It also resonates with a discussion with @sk1p to use the threaded executor as a "quick response" stand-by and spin up a process-based one in the background.

One could even implement a hybrid executor that is threaded when snoozed, and brings up Dask workers one by one, starting with GPU workers, to replace the threads.

matbryan52:
Merging as the backend is fully functional, and so that the frontend points raised in the review can be addressed ( see #1575 ).

tests/server/test_state.py
```diff
@@ -101,14 +101,14 @@ async def test_snooze_last_activity(async_executor):
         _ = executor_state.get_context()
         executor_state._update_last_activity.assert_called()
 
-        executor_state._update_last_activity.assert_called()
```
sk1p: Well spotted :+1:

PR #1579, opened 2024-01-31
Test on M1 / Mac OS 14

sk1p:
Use the just-announced public M1 runners: https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,11 @@ jobs:
       matrix:
         python_version: ["3.7", "3.9", "3.10", "3.11", "3.12"]
         os: [ubuntu-latest, windows-latest, macos-latest, macos-14]
+        exclude:
+          - os: macos-14
+            python_version: "3.7"
+          - os: macos-14
+            python_version: "3.9"
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       matrix:
         python_version: ["3.7", "3.9", "3.10", "3.11", "3.12"]
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        os: [ubuntu-latest, windows-latest, macos-latest, macos-14]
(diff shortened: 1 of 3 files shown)


sk1p:
@matbryan52 this fails on `test_many_files_read` - maybe related to the recent changes?

Also the M1 runners seem to be quite fast, which is a nice change from the previous situation where Mac OS was almost always the slowest.

matbryan52:
> @matbryan52 this fails on `test_many_files_read` - maybe related to the recent changes?
> 
> Also the M1 runners seem to be quite fast, which is a nice change from the previous situation where Mac OS was almost always the slowest.

Certainly, this is just a test for the `ulimit` increase during preloading workers. Seems the `macos-14` runner doesn't have the limits that the test expects. We could skip this on `macos-14` or even entirely if you like ? It's not really a robust test as-is across the board...

sk1p:
> Certainly, this is just a test for the `ulimit` increase during preloading workers. Seems the `macos-14` runner doesn't have the limits that the test expects. We could skip this on `macos-14` or even entirely if you like ? It's not really a robust test as-is across the board...

Hmm, the test should already be skipped in case the hard file limit is too low:

https://github.com/LiberTEM/LiberTEM/blob/6d93939629a7236da77bd7332a561178ede5d7ac/tests/io/test_many_files.py#L8-L14

Maybe the threshold is too close to the actual usage, and we open too many "other" files? I'll try to set the threshold to `2**15` and see if that "fixes" the issue (by just not running on mac os 14)

sk1p:
> Maybe the threshold is too close to the actual usage, and we open too many "other" files? I'll try to set the threshold to `2**15` and see if that "fixes" the issue (by just not running on mac os 14)

Huh, that didn't work at all. Weird - maybe there's indeed something wrong with setting the limits on the latest mac os...

sk1p:
Hmm, I'm not particularly happy about just skipping this test, but I think this might be an issue that is best debugged in an interactive session (why is it telling us that the hard limit is high enough, and then doesn't properly adjust the limit afterwards?). I'll see if I can get access to an ARM-based Mac, if I can reproduce this issue, but I'll merge for now.

Issue #1577, opened 2024-01-24
Revisit event loop blocking functions


sk1p:
Remainders from #1572:

  - [ ] `unsnooze` -> `get_executor`: slow on some machines, so it might need to be `async` (or can it run in a thread?)
  - [ ] `shutdown`: closing the executor could take a while?
  - [ ] `make_executor`

Also, the cluster spin-up could be more lazy, like only wait for a fraction of workers to come up, instead of all of them, and already start servicing real user requests.

sk1p:
> * [ ]  `unsnooze` -> `get_executor`: slow on some machines, so it might need to be `async` (or can it run in a thread?)
> * [ ]  `shutdown`: closing the executor could take a while?
> * [ ]  `make_executor`
> 
> Regarding `unsnooze` and `make_executor` - both are basically blocked by this issue:
> https://github.com/LiberTEM/LiberTEM/blob/5d4d2de2256113b3760fabede1a316716fadaa6b/src/libertem/web/server.py#L244-L252

This is now taken care of as part of #1578 - `shutdown` probably can't be async, as that is called too late in the whole process. `shutdown` is generally a bit broken right now - if it's called from a different thread than where the loop is running, we can't reliably wait for the cancellation to actually go through (and if the event loop is dead, it's pointless anyways)

uellue:
Is this issue still relevant?

PR #1580, opened 2024-02-04
Run `web_api` tests with `PYTHONASYNCIODEBUG=1`

sk1p:
Fix snooze task cancellation being called from the wrong thread (which was caught by this change).

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/web/state.py
+++ b/src/libertem/web/state.py
@@ -51,6 +51,7 @@ class ExecutorState:
             snooze_timeout and (snooze_timeout * 0.1) or 30.0,
         )
         self._snooze_task = asyncio.ensure_future(self._snooze_check_task())
+        self._loop = asyncio.get_running_loop()
         self._keep_alive = 0
         self.local_directory = "dask-worker-space"
         self.preload: typing.Tuple[str, ...] = ()
@@ -151,7 +152,7 @@ class ExecutorState:
         return self.context
 
     def shutdown(self):
-        self._snooze_task.cancel()
+        self._loop.call_soon_threadsafe(self._snooze_task.cancel)
         if self.context is not None:
             self.context.close()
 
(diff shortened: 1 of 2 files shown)


matbryan52:
Reproduced the nbval test failure outside of tox:

image

Seems to be slow coroutines being logged as warning by the new envvar, which is bizarre as the variable is only set during the `web_api` env ??

sk1p:
> Reproduced what is probably the nbval test failure outside of tox:

[...]

> Seems to be slow coroutines being logged as warning by the new envvar, which is bizarre as the variable is only set during the `web_api` env ??

Thanks for tracking this down. That is weird indeed - can you check `os.environ['PYTHONASYNCIODEBUG']` inside the notebook? It's either [this environment variable](https://docs.python.org/3/library/asyncio-eventloop.html#enabling-debug-mode), `loop.set_debug(True)` or the [Python development mode](https://docs.python.org/3/library/devmode.html#devmode) that should trigger these log messages...

The log messages themselves are related to long running tasks, see also #1577 - are there any other messages in there, maybe related to thread safety or other "async sins"?

> NOTE: there seem to be two CI failures, first the async notebook fails from cell [2], then the job got stuck and later cancelled when running binning.ipynb.

matbryan52:
Think it might just have been a red herring ? I started going down the rabbit hole of envvars bleeding through tox configs but actually it passed fine on re-run.

Maybe the worker had a hiccup which caused a log message in the notebook cell ?

I would say LGTM and monitor during other CI runs...

sk1p:
> Think it might just have been a red herring ? I started going down the rabbit hole of envvars bleeding through tox configs but actually it passed fine on re-run.
> 
> Maybe the worker had a hiccup which caused a log message in the notebook cell ?
> 
> I would say LGTM and monitor during other CI runs...

Thanks for looking into that. Yeah, I would concur. I'm quite certain that there _are_ issues with async, especially if `distributed` is not up to date, or in the shutdown code, but this change should only affect the web API, not the bare Python API etc.

I'll go ahead and merge!

Issue #256, opened 2019-01-28
Handle missing data


sk1p:
Sometimes, the acquisition may be canceled and the dataset may not contain as many images to fill the whole scan region. We should handle this case in some way. Related to #120 and maybe also our live acquisition plans.

In the FRMS6 format, the "signalFrames" header value will contain the real number of acquired frames. Maybe we can just ignore that we don't have data available, output a warning etc.

uellue:
What about displaying the missing data as black or checkerboard?

sk1p:
Black could work, as long as there is no need to distinguish between 0 and "no data". Checkerboard could be problematic for missing values that are more spread around, like randomly missing frames. I guess we should have general visualization settings, where the user can select the appropriate placeholder (could also be any color that is not part of and very different from the current colormap).

Previous thoughts on this, from #128:

> 
> How do we, generally, handle "missing" or invalid values in visualization? Examples:
> 
>  - complete frames may be missing, for example because the acquisition did not keep up (with UDP used in acquisition this is a real-world scenario!)
>  - values may fall out of the range of visualization functions, for example negative numbers with log scaling
>  - other examples?
> 
> For specific cases, we should try to find a good place holder (as with center of mass: the origin / reference center), but if we have real missing data, this is not so easy. Maybe a bright neon colored marker? Like [darktable clipping indicator](https://redmine.darktable.org/attachments/download/2582/overexposed-toggle_on.png) or for larger areas, [darktable raw over exposure indication](https://www.darktable.org/2016/10/raw-overexposed/rawoverexposed-2-1.jpeg)

sk1p:
This can now also come up if we visualize a result from running a UDF with a `roi` set. In that case, values masked out are `NaN` in the result. So, I propose:

 * augment `visualize_simple` to properly handle `NaNs`
 * add a method to query the dataset for missing data, which returns a bool mask array
 * as a postprocessing step, add `NaNs` in the places of missing data (udf: `run_for_dataset*`, for the old-style built-in jobs we may have to do this in each `Job`)

PR #1593, opened 2024-02-08
Valid data masking

### LINKED ISSUE #1449: Allow to influence "damage" of results
Use case: visualization of partial results  with `kind='single'` and possibly custom shape, but the UDF knows what parts of it are valid. Examples are SSB/WDD or possibly iCoM.

### LINKED ISSUE #1473: Make `damage` available in `get_results()`
When calculating derived results in `UDF.get_results()` it can be useful to know the current `damage`, i.e. which parts of a buffer contain valid data. Offset and linear gradient correction in CoMUDF #1458 would be a concrete example where only valid portions of a result buffer should be included in the correction.

When looking at `UDFRunner`, the damage is an additional result buffer that is kept outside of the UDF. It is only merged with the UDF results when the final results object is assembled.

Possibly the way how damage is handled in live plots could be an inspiration? Here we already have an infrastructure that is similar to `get_results()`, but aware of damage. One could consider keeping track of damage for each result buffer? SSB and WDD would be examples of `kind='single'` derived result buffers that are, in a way, actually `kind='nav'`.

sk1p:
Makes the "valid nav mask" available in UDFs, and propagates it to results.

Closes #1473, closes #1449

## TODO

- [x] improve test cases
- [x] implement `BufferWrapper.masked_data` and friends
- [x] update plotting to use the per-buffer valid data mask instead of "global" damage
- [x] figure out how this fits best with `roi`s
- [x] add `raw_masked_data`
- [x] documentation

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/docs/source/changelog/features/valid-data-masking.rst
+++ b/docs/source/changelog/features/valid-data-masking.rst
@@ -1,0 +1,10 @@
+[Feature] Valid data masks
+==========================
+
+* Make the mask of already processed navigation elements available
+  to the UDF in :meth:`~libertem.udf.base.UDF.merge` and
+  :meth:`~libertem.udf.base.UDF.get_results` (:issue:`1473`),
+  propagate this information to the result buffers
+  and allow UDFs to override this for results using the new
+  :meth:`~libertem.udf.base.UDF.with_mask` method
+  (:issue:`1449` :pr:`1593`).
--- a/docs/source/changelog/features/valid-data-masking.rst
+++ b/docs/source/changelog/features/valid-data-masking.rst
@@ -1,10 +1,10 @@
 [Feature] Valid data masks
 ==========================
 
-* Make the mask of already processed navigation elements available
-  to the UDF in :meth:`~libertem.udf.base.UDF.merge` and
-  :meth:`~libertem.udf.base.UDF.get_results` (:issue:`1473`),
-  propagate this information to the result buffers
-  and allow UDFs to override this for results using the new
-  :meth:`~libertem.udf.base.UDF.with_mask` method
+* Add endpoints to provide :meth:`~libertem.udf.base.UDF.merge`
+  and :meth:`~libertem.udf.base.UDF.get_results` with information
+  on which parts of the dataset have already been processed
+  (:issue:`1473`). The information is then propagated to the result
+  buffers, and UDFs can override the validity mask for results
+  using the new :meth:`~libertem.udf.base.UDF.with_mask` method
   (:issue:`1449` :pr:`1593`).
--- a/docs/source/udf/advanced.rst
+++ b/docs/source/udf/advanced.rst
@@ -187,7 +187,7 @@ Post-processing after merging
 
 If you want to implement a post-processing step that is run on the main node
 after merging result buffers, you can override
-:meth:`libertem.udf.base.UDF.get_results`:
+:meth:`~libertem.udf.base.UDF.get_results`:
 
 .. testsetup::
 
@@ -222,22 +222,23 @@ after merging result buffers, you can override
 
     ctx.run_udf(dataset=dataset, udf=AverageUDF())
 
-Note that :meth:`UDF.get_result_buffers` returns a placeholder entry for the
+Note that :meth:`~libertem.udf.base.UDF.get_result_buffers` returns a placeholder entry for the
 :code:`average` result using :code:`use='result_only'`, which is then filled in
-:code:`get_results`.  We don't need to repeat those buffers that should be
+:meth:`~libertem.udf.base.UDF.get_results`.  We don't need to repeat those buffers that should be
 returned unchanged; if you want to omit a buffer from the results completely,
 you can declare it as private with :code:`self.buffer(..., use='private')` in
-:code:`get_result_buffers`.
+:meth:`~libertem.udf.base.UDF.get_result_buffers`.
 
-:meth:`UDF.get_results` should return the results as a dictionary of numpy
+:meth:`~libertem.udf.base.UDF.get_results` should return the results as a dictionary of numpy
 arrays, with the keys matching those returned by
-:meth:`UDF.get_result_buffers`.
+:meth:`~libertem.udf.base.UDF.get_result_buffers`.
 
-When returned from :meth:`Context.run_udf`, all results are wrapped into
-:code:`BufferWrapper` instances. This is done primarily to get convenient
-access to a version of the result that is suitable for visualization, even if
+When returned from :meth:`~libertem.api.Context.run_udf`, all results are wrapped into
+:class:`~libertem.common.buffers.BufferWrapper` instances. This is done primarily to get convenient
+access to a version of the result that is suitable for visualization using
+:attr:`~libertem.common.buffers.BufferWrapper.data`, even if
 a :code:`roi` was used, but still allow access to the raw result using
-:attr:`BufferWrapper.raw_data` attribute.
+:attr:`~libertem.common.buffers.BufferWrapper.raw_data` attribute.
 
 The detailed rules for buffer declarations, :code:`get_result_buffers` and :code:`get_results` are:
 
--- a/docs/source/udf/advanced.rst
+++ b/docs/source/udf/advanced.rst
@@ -280,10 +280,10 @@ Valid data masking
 ------------------
 
 In various intermediate states, result buffers may be only partially filled
-with valid data. This comes from the fact that we split the data set into
-partitions, and then merge results together as they become available.
-Also, when doing live processing, you may have a direct mapping of acquired
-detector frames to result items, where naturally only positions where we have
+with valid data. This comes from the fact that we split the dataset into
+partitions and then merge results together only as they become available.
+When doing live processing we also have the constraint that not all frames
+have been acquired yet, and so naturally only positions where we have
 received and processed data can be valid in the corresponding result buffer.
 
 Handling partially valid data becomes important when running the UDF. For example,
@@ -331,9 +331,12 @@ Buffer :code:`kind` Default valid mask
 =================== ==================
 
 In your UDF, you can also customize the valid mask that should be returned.
-This is useful if there is no direct mapping between the already processed
-navigation elements and the valid results, or if you are using a :code:`kind='single'`
-buffer which is progressively built up, to override the default of all-valid.
+This is useful if the dataset might contain invalid frames discovered at runtime,
+or if intermediate results cannot be fully valid due to missing information (e.g.
+calculating gradients across partition boundaries). Another example is when using
+a :code:`kind='single'` buffer which is progressively computed; valid-by-default
+might be incorrect if not enough information has been seen (e.g. statistical measures
+or a count of values).
 
 Customizing the valid data mask is done in :meth:`~libertem.udf.UDF.get_results`
 with the :meth:`~libertem.udf.UDF.with_mask` method. For example:
@@ -341,6 +344,7 @@ with the :meth:`~libertem.udf.UDF.with_mask` method. For example:
 .. testsetup::
 
     from libertem.udf import UDF
+    from libertem.common.math import count_nonzero    
 
 
 .. testcode::
--- a/docs/source/udf/advanced.rst
+++ b/docs/source/udf/advanced.rst
@@ -359,6 +359,9 @@ with the :meth:`~libertem.udf.UDF.with_mask` method. For example:
                 nav_shape = (count_nonzero(self.meta.roi),)
 
             return {
+                # Emulate a kind='nav' result buffer using a kind='sig' buffer for demonstration.
+                # In the real world, such result buffers can be found in the implementations of ptychography
+                # and iCoM since they reconstruct the object in Fourier space
                 'custom_2d': self.buffer(kind='single', dtype=np.float32, extra_shape=nav_shape),
             }
 
--- a/docs/source/udf/advanced.rst
+++ b/docs/source/udf/advanced.rst
@@ -387,6 +387,12 @@ method, which is available in :meth:`~libertem.udf.UDF.merge` and
 already-processed navigation elements.
 
 .. versionadded:: 0.14.0
+    
+   In previous versions, only :attr:`libertem.udf.base.UDFResults.damage` was
+   available for inspecting the mask of already processed data.
+   This did not account for differences between buffers, and was only available
+   when iterating over results, i.e. with :meth:`~libertem.api.Context.run_udf_iter`.
+
 
 AUX data
 --------
--- a/docs/source/udf/advanced.rst
+++ b/docs/source/udf/advanced.rst
@@ -276,6 +276,114 @@ by the ROI.
     node, too. Views for aux data are set correctly on the main node. Previously,
     it was only executed on the worker nodes.
 
+Valid data masking
+------------------
+
+In various intermediate states, result buffers may be only partially filled
+with valid data. This comes from the fact that we split the data set into
+partitions, and then merge results together as they become available.
+Also, when doing live processing, you may have a direct mapping of acquired
+detector frames to result items, where naturally only positions where we have
+received and processed data can be valid in the corresponding result buffer.
+
+Handling partially valid data becomes important when running the UDF. For example,
+live plotting needs to know what regions of the results to include. Also, when using
+:meth:`~libertem.api.Context.run_udf_iter` you might only want to handle the already valid
+parts of the results.
+
+For UDF results, you can access
+:attr:`~libertem.common.buffers.BufferWrapper.valid_mask` to access
+the valid mask for that particular buffer. You can also use
+:attr:`~libertem.common.buffers.BufferWrapper.masked_data` 
+to access the data of the buffer as a numpy masked array.
+
+For example:
+
+.. testsetup::
+
+    from libertem.udf.sumsigudf import SumSigUDF
+
+.. testcode::
+
+    # use `run_udf_iter` to get partial results while the computation is running:
+    for res in ctx.run_udf_iter(dataset=dataset, udf=SumSigUDF()):
+
+        # get the BufferWrapper for the 'intensity' buffer of the first UDF:
+        buf = res.buffers[0]['intensity']
+
+        # boolean mask that contains True for valid elements:
+        mask = buf.valid_mask
+
+        # get a numpy masked array that is limited to the valid elements:
+        masked_data = buf.masked_data
+
+
+By default, the valid masks of results depend on the buffer kind:
+
+=================== ==================
+Buffer :code:`kind` Default valid mask
+=================== ==================
+:code:`'single'`    All-valid
+------------------- ------------------
+:code:`'sig'`       All-valid
+------------------- ------------------
+:code:`'nav'`       Parts which have been merged using :meth:`~libertem.udf.UDF.merge`
+=================== ==================
+
+In your UDF, you can also customize the valid mask that should be returned.
+This is useful if there is no direct mapping between the already processed
+navigation elements and the valid results, or if you are using a :code:`kind='single'`
+buffer which is progressively built up, to override the default of all-valid.
+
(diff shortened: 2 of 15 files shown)


uellue:
I like the API! Really minimal and simple. Regarding masked array support in other backends, I came up with the following:

* Dask: https://docs.dask.org/en/latest/generated/dask.array.ma.masked_array.html
* CuPy: https://github.com/cupy/cupy/issues/2225
* Sparse: https://github.com/pydata/sparse/issues/359

Since we have, as far as I am aware, only NumPy and Dask results, it should work. If we want to support other return backends, one could implement the thin wrapper alternative for them, essentially a "poor man's masked array" type. When the time comes we can think of an API that allows creating the right kind of array which is backwards-compatible with NumPy masked arrays.

Regarding ROI behavior, what about the following?

Internally, everything works on the ROI like before. That means the mask is relative to the nav space with ROI applied -- always of the same shape as the actual result buffer one is interacting with. Result buffer `.data` only sets the values within the ROI that are not masked, meaning masked parts of result buffers are treated the same way as areas outside the ROI.

uellue:
Hm, the `fill_value` parameter is interesting relative to our ROI considerations. One could also argue that masked areas should have the fill value and not the "outside of ROI" value.

Now I'm not so sure if we should use masked arrays directly.

## Pro

* We use an existing API of NumPy
* Potentially benefit from additional capabilities, see open question
* It generally matches what we want to express with the valid mask

## Con
* We should support all features of masked arrays in a predictable way if we allow users to pass them directly. With ROIs it leads to an ambiguous situation.
* The behavior of masked arrays is opposite of ROIs and damage -- "1" means invalid for masked arrays, but valid for ROI and damage.
* Not available in CuPy, which would be a likely target for `get_results()` implementations that benefit from GPUs, like iCoM

## Open questions

* Will the additional features of masked arrays be useful for the user, such as numerical operations on them?

## Conclusion

Looking at the implications, we might be better off to implement a thin wrapper type?

matbryan52:
> ## Open questions
> * Will the additional features of masked arrays be useful for the user, such as numerical operations on them?
> 
> ## Conclusion
> Looking at the implications, we might be better off to implement a thin wrapper type?

My view on this part is that if we go with a thin wrapper, we can always offer a method on the wrapper which gives the user a true `np.ma` array with the most useful properties for them (valid only, roi only, valid + roi etc), and it gives us more flexibility in terms of fill values because we can handle that ourselves. We also don't have to construct anything new unless the user actually wants to take advantage of `np.ma`, which might avoid some copies ?

We also retain the option to support non-masked-compatible backends, as you've said already.

Overall I like the implementation and idea behind it 👍

sk1p:
> The behavior of masked arrays is opposite of ROIs and damage -- "1" means invalid for masked arrays, but valid for ROI and damage.

Oof, yeah, I completely missed that one. I think that alone is enough to convince me that a thin wrapper can provide a better API.

> > Looking at the implications, we might be better off to implement a thin wrapper type?
> 
> My view on this part is that if we go with a thin wrapper, we can always offer a method on the wrapper which gives the user a true `np.ma` array with the most useful properties for them (valid only, roi only, valid + roi etc), and it gives us more flexibility in terms of fill values because we can handle that ourselves. We also don't have to construct anything new unless the user actually wants to take advantage of `np.ma`, which might avoid some copies ?
> 
> We also retain the option to support non-masked-compatible backends, as you've said already.

Thanks for the feedback, I'll work on an updated version of these changes. I'm thinking on what aspects have to be taken into account.

> Since we have, as far as I am aware, only NumPy and Dask results

Yeah, according to our documentation, `get_results` [is supposed to always return `np.ndarray`](https://libertem.github.io/LiberTEM/udf/advanced.html#post-processing-after-merging). This is actually wrong, as you correctly state, it can also be a `dask.array.core.Array`, with the delayed executor. I ran the test suite with the following patch, and at least for what we cover, `Union[np.ndarray, dask.array.core.Array]` is the true type there.

```patch
diff --git a/src/libertem/udf/base.py b/src/libertem/udf/base.py
index bc0b80b9..3efc03e3 100644
--- a/src/libertem/udf/base.py
+++ b/src/libertem/udf/base.py
@@ -1214,6 +1214,8 @@ class UDFBase(UDFProtocol):
         # wrap numpy results into `ResultBuffer`s:
         results = {}
         for name, arr in results_tmp.items():
+            import dask.array.core
+            assert isinstance(arr, (np.ndarray, dask.array.core.Array))
             self._check_results(decl, arr, name)
             buf_decl = decl[name]
             buf = results_buffer_cls[name](
```

We do not yet support returning `cupy` arrays from `get_results`, for the iCoM case, I'm currently doing a round-trip CPU->GPU->CPU, which is still very fast . In the future, if we want to optimize the "very large scan" use-case, we may want to keep the result buffers on the GPU, for example to directly apply a shader for visualization.

I'm also thinking about including some helper functions on the wrapper type, as there will be some common operations:

* Is everything valid?
* What is the (outer) bounding box (as 2d or nd slices)?
* What is the inner bounding box, meaning the slices for the "rectangular" region where all values are valid?

## Tangent: missing data and dropped frames

So, we also had the topic of missing data some time ago (#256), where we kind of punted out, IIRC with the data set returning zeros for missing portions or similar. Now, with live processing, it will sadly become more common that we drop frames (for example when optimizing a new UDF for live processing and can't keep up, or when the computer used is simply not fast enough to keep up with the chosen computation). With the `valid_nav_mask` or something similar, we now have the chance  to properly communicate this to the UDF in a consistent way. But maybe we actually want to have a separate mask for this, to distinguish dropped frames / missing data from (in the case of STEM) not having visited a given position yet.

A reason for keeping these separately is that UDFs may want to only show results in the "inner bounding box", a rectangular region where all positions have been seen. Otherwise, we may get artifacts/histogram/flickering issues for the live display, at borders etc. If we have a single mask only, and we are missing a single frame, this region will never grow beyond that position.

So maybe, in this PR, `valid_nav_mask` should actually be documented to _include_ positions with missing data, and in the future we can add a second mask especially for the missing data. Agreed?

sk1p:
The `web_api` tests are a bit flaky now, which I managed to reproduce locally, too. I suspect this is related to #1577 and the changes to how the server creates executors etc. - before bugging out, the test was hanging for quite some time in `make_executor`:

```
Thread 1706497 (idle): "LiberTEM-background"
    stop (distributed/utils.py:606)
    close (distributed/client.py:1790)
    __del__ (distributed/client.py:1522)
    _make_invoke_excepthook (threading.py:1324)
    __init__ (threading.py:910)
    __init__ (distributed/utils.py:479)
    _start_unlocked (distributed/utils.py:590)
    start (distributed/utils.py:551)
    start (distributed/client.py:1209)
    __init__ (distributed/client.py:1014)
    connect (libertem/executor/dask.py:701)
    make_executor (libertem/web/state.py:128)
    _do_connect (libertem/web/connect.py:53)
    put (libertem/web/connect.py:45)
    _execute (tornado/web.py:1790)
    _run (asyncio/events.py:80)
    _run_once (asyncio/base_events.py:1914)
    run_forever (asyncio/base_events.py:607)
    run (conftest.py:939)
    _bootstrap_inner (threading.py:1045)
    _bootstrap (threading.py:1002)
```

(where "LiberTEM-background" is the thread used to run the web API in the background for testing - used in fixtures like `base_url`)

matbryan52 (commented):
In comments

src/libertem/common/buffers.py
```diff
@@ -167,6 +180,130 @@ def __del__(self):
...
+    mask (for example, defining a region in the array
+    where there are valid entries)
+
+    This is meant for usage in :meth:`UDF.get_results`,
+    and you can use the convenience method :meth:`UDF.with_mask`
+    to instantiate it.
+    """
+
+    def __init__(self, arr: T, mask: np.ndarray):
+        if isinstance(mask, int):
```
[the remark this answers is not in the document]
sk1p: True, I'll update the type signatures.

src/libertem/udf/base.py
```diff
@@ -1594,6 +1644,9 @@ def buffer(
             return PlaceholderBufferWrapper(kind, extra_shape, dtype, use=use)
         return BufferWrapper(kind, extra_shape, dtype, where, use=use)
 
+    def with_mask(self, data, mask: np.ndarray) -> ArrayWithMask:
+        return ArrayWithMask(data, mask=mask)
```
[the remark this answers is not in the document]
sk1p: Agree on the `@staticmethod`, but I'll keep it in `UDF`, as the method needs to appear in the documentation (and as far as I can see, `UDFBase` doesn't, and mostly contains implementation details)

src/libertem/udf/base.py
```diff
@@ -379,6 +380,10 @@ def __init__(
         self._corrections = corrections
         self._threads_per_worker = threads_per_worker
         self._array_backend = array_backend
+        self._valid_nav_mask = valid_nav_mask
+
+        if valid_nav_mask is not None and roi is not None:
+            assert valid_nav_mask.shape[0] == np.count_nonzero(roi)
```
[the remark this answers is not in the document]
sk1p: True, I'll update it to use our `count_nonzero`

src/libertem/udf/base.py
```diff
@@ -552,6 +557,38 @@ def array_backend(self) -> Optional[ArrayBackend]:
         """
         return self._array_backend
 
+    def get_valid_nav_mask(self, full_nav: bool = False) -> Optional[np.ndarray]:
+        """
+        Return a mask of the already computed nav positions, as flattened
+        1D array.
+
+        Only available in :meth:`~libertem.udf.base.UDF.merge` and
```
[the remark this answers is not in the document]
sk1p: Good catch, I'll also add a test case that ensures `get_valid_nav_mask` always returns `None` in `process_*` (in the case of the inline executor, this apparently wasn't the case, so now I'm explicitly resetting this value to ensure we get a consistent experience across executors)

tests/udf/test_valid_mask.py
```diff
@@ -0,0 +1,388 @@
...
+            full_mask = self.meta.get_valid_nav_mask(full_nav=True)
+            assert full_mask.shape[0] == self.meta.dataset_shape.nav.size, \
+                "when passing `full_nav=True`, the shape must match the flattened ds shape"
+        if self.params.debug:
+            print("get_results", self.meta.get_valid_nav_mask())
+        results = super().get_results()
+        # import pdb; pdb.set_trace()
+        return results
+
+    def process_frame(self, frame):
```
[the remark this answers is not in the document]
sk1p: Good idea, will do!

sk1p:
@uellue could you have a look specifically at the CoM changes?

docs/source/changelog/features/valid-data-masking.rst
```diff
@@ -0,0 +1,10 @@
+[Feature] Valid data masks
+==========================
+
+* Make the mask of already processed navigation elements available
+  to the UDF in :meth:`~libertem.udf.base.UDF.merge` and
+  :meth:`~libertem.udf.base.UDF.get_results` (:issue:`1473`),
+  propagate this information to the result buffers
+  and allow UDFs to override this for results using the new
+  :meth:`~libertem.udf.base.UDF.with_mask` method
+  (:issue:`1449` :pr:`1593`).
```
matbryan52, line 10: Rephrasing to help with the release notes later:

```suggestion
* Add endpoints to provide :meth:`~libertem.udf.base.UDF.merge`
  and :meth:`~libertem.udf.base.UDF.get_results` with information
  on which parts of the dataset have already been processed
  (:issue:`1473`). The information is then propagated to the result
  buffers, and UDFs can override the validity mask for results
  using the new :meth:`~libertem.udf.base.UDF.with_mask` method
  (:issue:`1449` :pr:`1593`).
```

docs/source/udf/advanced.rst
```diff
@@ -275,6 +276,114 @@ by the ROI.
...
+This is useful if there is no direct mapping between the already processed
+navigation elements and the valid results, or if you are using a :code:`kind='single'`
+buffer which is progressively built up, to override the default of all-valid.
+
+Customizing the valid data mask is done in :meth:`~libertem.udf.UDF.get_results`
+with the :meth:`~libertem.udf.UDF.with_mask` method. For example:
+
+.. testsetup::
+
+    from libertem.udf import UDF
```
matbryan52, line 343: Import (unused, but in the test code)


```suggestion
    from libertem.udf import UDF
    from libertem.common.math import count_nonzero    
```

docs/source/udf/advanced.rst
```diff
@@ -275,6 +276,114 @@ by the ROI.
...
+------------------- ------------------
+:code:`'sig'`       All-valid
+------------------- ------------------
+:code:`'nav'`       Parts which have been merged using :meth:`~libertem.udf.UDF.merge`
+=================== ==================
+
+In your UDF, you can also customize the valid mask that should be returned.
+This is useful if there is no direct mapping between the already processed
+navigation elements and the valid results, or if you are using a :code:`kind='single'`
+buffer which is progressively built up, to override the default of all-valid.
```
matbryan52, line 336:
```suggestion
This is useful if the dataset might contain invalid frames discovered at runtime,
or if intermediate results cannot be fully valid due to missing information (e.g.
calculating gradients across parition boundaries). Anothe example is when using a :code:`kind='single'` buffer which is progressively computed, valid-by-default
might be incorrect if not enough information has been seen (e.g. statistical measures
or a count of values).
```

docs/source/udf/advanced.rst
```diff
@@ -275,6 +276,114 @@ by the ROI.
...
 
+Valid data masking
+------------------
+
+In various intermediate states, result buffers may be only partially filled
+with valid data. This comes from the fact that we split the data set into
+partitions, and then merge results together as they become available.
+Also, when doing live processing, you may have a direct mapping of acquired
+detector frames to result items, where naturally only positions where we have
+received and processed data can be valid in the corresponding result buffer.
```
matbryan52, line 287:
```suggestion
In various intermediate states, result buffers may be only partially filled
with valid data. This comes from the fact that we split the dataset into
partitions and then merge results together only as they become available.
When doing live processing we also have the constraint that not all frames
have been acquired yet, and so naturally only positions where we have
received and processed data can be valid in the corresponding result buffer.
```

matbryan52 (approved):
Looks good to me. If there is anything missing it is a description in the doc of how this differs from the existing `damage` on `UDFResults`, which is linked but a less granular information about validity.

sk1p:
> Looks good to me. If there is anything missing it is a description in the doc of how this differs from the existing `damage` on `UDFResults`, which is linked but a less granular information about validity.

Thanks for having a look, I'll add a short note about the relation to `UDFResults.damage`.

src/libertem/common/buffers.py
```diff
@@ -167,6 +180,141 @@ def __del__(self):
...
+    where there are valid entries)
+
+    This is meant for usage in :meth:`UDF.get_results`,
+    and you can use the convenience method :meth:`UDF.with_mask`
+    to instantiate it.
+    """
+
+    def __init__(self, arr: T, mask: Union[np.ndarray, bool]):
+        if isinstance(mask, int):
+            mask = np.array([mask])
```
uellue, line 207: The signature says `ndarray` or `bool`, and in the next line we test for `int`, which is a superset of `bool`. Probably the parameters should be documented in the docstring?

* `ndarray` of what types?
* What is the difference between dtype `int` and `bool` ndarrays?

The behavior of NumPy fancy indexing is different with an `int` mask or `bool` mask: With an `int` mask it returns the elements at those indices. I can't quite imagine how that should work with valid data masking, so we should probably restrict the mask to dtype `bool`?

src/libertem/udf/com.py
```diff
@@ -616,13 +614,12 @@ def get_inp():
             if com_params.regression == -1:
                 pass
             elif com_params.regression == 0:
-                valid = get_valid()
-                valid_field = field[valid]
+                valid_field = field[valid_mask]
```
uellue, line 617: ...see here for a case where a mask of dtype `int` would not work as expected.

src/libertem/udf/com.py
```diff
@@ -680,12 +677,22 @@ def get_results(self):
...
+        regression, inp = self.get_regression(field, valid_mask=valid_mask)
+
         if inp is not None:
             self.apply_lin_regression(regression, inp, field)
         elif not np.allclose(regression[0], 0):
             self.apply_mean_regression(regression, field)
 
+        # zero the field values inside of the invalid-mask
+        # as it might be non-zero from the applied regression:
+        # field[np.broadcast_to(mask, field.shape)] = 0
```
uellue, line 694: Why is this in there, commented out? The variable `mask` is not defined in the previous code, as far as I can tell.

tests/udf/test_valid_mask.py
```diff
@@ -0,0 +1,475 @@
...
+        dest.buf_single += src.buf_single
+        dest.buf_nav[:] = src.buf_nav
+
+
+def test_valid_nav_mask_available():
+    dataset = MemoryDataSet(datashape=[16, 16, 32, 32], num_partitions=4)
+    ctx = Context.make_with('inline')
+    for res in ctx.run_udf_iter(dataset=dataset, udf=ValidNavMaskUDF()):
+        # TODO: maybe compare damage we got in `get_results` with `res.damage` here?
+        pass
```
uellue, line 66: Is there a reason to not use the default fixtures for dataset and inline context?

The pattern repeats in the following tests.

tests/udf/test_valid_mask.py
```diff
@@ -0,0 +1,475 @@
...
+    (64, 64, 4),
+    (1, 1, 4),
+    (1, 1, 1, 1),  # that's too many...
+])
+def test_custom_mask_invalid(mask_shape):
+    """
+    Examples of mask shapes that are incompatible with the 'custom' buffer
+    defined above. Make sure these raise an appropriate exception.
+    """
+    dataset = MemoryDataSet(datashape=[16, 16, 4, 4], num_partitions=4)
```
uellue, line 228: In the same vein we can test that invalid dtypes are rejected

uellue (changes requested):
Looks good, with minor comments on dtypes.

I'll also contribute a test case for correct CoMUDF regression with incomplete input data, so please don't merge yet!

src/libertem/udf/com.py
uellue: Went ahead and limited the application of regression to valid data.

src/libertem/common/buffers.py
sk1p: True, this was confusing, the type should be `bool` or `np.ndarray`. I'll add documentation to `with_mask`, which is the public interface, to clear this up, and I'll consistently use `bool` instead of `int` (for Python this doesn't really matter, and I somehow got into the habit of using `1` instead of `True` sometimes, but it's clear that there's a difference for `numpy`...). The intention of the interface was to pass in boolean masks, not lists of coordinates, which could be added as a follow-up IMHO.

uellue: Yeah, list of coordinates will also modify the shape and contain duplicates, so I think bool only makes sense for now. :-)

tests/udf/test_valid_mask.py
sk1p: For one, I've lost overview over the different fixtures we have available. We may need documentation for the most commonly used fixtures in our contributor docs... I used `make_with` as it's just so convenient :) I'll update the tests to use `lt_ctx` instead, which actually does useful things for testing (replaces the plotting with a dummy, and restricts the number of inline threads).

As for the data set, I think it's good to actually have the code in the test, as that means it is self-contained, and you can directly understand the shape of the test data from the test code. When using the fixture, you need to go hunting first. Let me know if you have concerns with this approach, I can easily update the tests to use the `default_raw` fixture instead.

uellue: Good point about the data shape! I also really like `make_with()`. :-)

uellue (approved):
Let's Go, Try Merge!👍

PR #1599, opened 2024-02-21
Make sure live plots work with the widget backend

uellue:
The newest versions of jupyterlab and notebook don't support the `nbagg` backend anymore. Instead the `widget` backend is supposed to be used: https://matplotlib.org/stable/users/explain/figure/interactive.html#jupyter-notebooks-jupyterlab

With this change the `widget` backend shows and updates the plots live as it is supposed to. It also uses the recommended live drawing functions for matplotlib:
https://matplotlib.org/stable/users/explain/figure/interactive_guide.html#explicitly-spinning-the-event-loop

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/examples/live-plotting.ipynb
+++ b/examples/live-plotting.ipynb
@@ -64,18 +64,7 @@
    "cell_type": "code",
    "execution_count": 3,
    "metadata": {},
-   "outputs": [
-    {
-     "name": "stderr",
-     "output_type": "stream",
-     "text": [
-      "C:\\Users\\weber\\miniconda3\\envs\\lt312\\Lib\\site-packages\\distributed\\node.py:182: UserWarning: Port 8787 is already in use.\n",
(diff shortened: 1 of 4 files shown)


uellue:
CC @sivborg!

uellue:
Hm, why are the data test failing now? The notebook test that failed previously seems successful.

uellue:
Something is not quite right in CI, it seems, Suddenly there are errors after errors.

uellue:
The HyperSpy issue is not reproduced locally on Python 3.11, Windows

sk1p (approved):
LGTM! Tested on both `notebook~=6` and `notebook>=7` with the live plotting notebook, seems to work fine (though the processing with that data set is too fast to actually see intermediate results and to be able to judge how interactive things are etc. - I think this is something we'll only find out in real usage).

PR #1600, opened 2024-02-22
Deprecate gridmatching and fullmatch

uellue:
Depends on releasing
https://github.com/LiberTEM/LiberTEM-blobfinder/pull/83, will fail and should not be merged until then.

Closes #1469

The tests are removed, which allows removing the test fixtures and corresponding imports, which then allows testing for a `DeprecationWarning` on import.

## Contributor Checklist:

* [ ] I have added or updated my entry in [the creators.json file](https://github.com/LiberTEM/LiberTEM/blob/master/packaging/creators.json)
* [ ] I have added [a changelog entry](https://github.com/LiberTEM/LiberTEM/tree/master/docs/source/changelog) for my contribution
* [ ] I have added/updated documentation for all user-facing changes
* [ ] I have added/updated test cases
* [ ] I have included the [rebuilt production build of the client](https://libertem.github.io/LiberTEM/contributing.html?#building-the-client) (only if changes were made to the GUI)

## Reviewer Checklist:

* [ ] `/azp run libertem.libertem-data` passed
* [ ] No import of GPL code from MIT code

PR #1602, opened 2024-02-22
Various CI updates

sk1p:
- Very minimal version of #1600 to speed up CI
- Remove `macos-latest` testing for now (please do let me know if you disagree, maybe we can compromise and keep a single job for "smoke testing"?)

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       matrix:
         python_version: ["3.7", "3.9", "3.10", "3.11", "3.12"]
-        os: [ubuntu-latest, windows-latest, macos-latest, macos-14]
+        os: [ubuntu-latest, windows-latest, macos-14]
         exclude:
           - os: macos-14
             python_version: "3.7"
--- a/tox.ini
+++ b/tox.ini
@@ -16,7 +16,6 @@ deps=
     cuda11x: cupy-cuda11x
     cuda12x: cupy-cuda12x
 extras=
-    hdbscan
     bqplot
     hdf5plugin
 setenv=
@@ -94,6 +93,8 @@ deps=
     notebooks-cuda110: cupy-cuda110
     notebooks-cuda11x: cupy-cuda11x
     notebooks-cuda12x: cupy-cuda12x
+extras=
+    hdbscan
 
 commands=
     pytest --durations=10 --cov=libertem --cov-report=term --cov-report=html --cov-report=xml --cov-config=setup.cfg --junitxml=junit.xml --nbval --nbval-sanitize-with nbval_sanitize.cfg -p no:python --nbval-current-env {posargs:examples/}


sk1p:
@uellue is this fine as a stop-gap solution until blobfinder is released?

@matbryan52 do you agree with reducing the CI coverage on macos/intel? I'm pretty much frustrated and fed up by their performance, a recent run took ~15min, where the comparable arm-based runner takes about 5min... :(

uellue (approved):
LGTM! I have also backported the setuptools from #1600. I checked, hdbscan is still required for docs_requirements.txt here, can only be dropped in #1600.

uellue:
Just another thought -- should we also drop Python 3.9 or 3.10?

uellue:
...or do data tests and regular tests with different Python versions? That way we cover more ground with less resources.

matbryan52:
> @matbryan52 do you agree with reducing the CI coverage on macos/intel? I'm pretty much frustrated and fed up by their performance, a recent run took ~15min, where the comparable arm-based runner takes about 5min... :(

I agree yes! Testing now on Mac M1 should cover the majority of mac particularities, in my view. If we add tests on Mac/Intel to the release checklist as a manual item maybe ??

> Just another thought -- should we also drop Python 3.9 or 3.10?

We already had one strange case where we had a failure in 3.8 uncaught because we skipped this version. That said this should be really rare.

matbryan52 (approved):

matbryan52:
> ...or do data tests and regular tests with different Python versions? That way we cover more ground with less resources.

Could Azure pipelines be used to run one or two jobs of the normal tests, maybe ? Say Py3.9 only on Azure ?

uellue:
Merging for now, we can do another PR for Python versions if necessary. :-)

sk1p:
> If we add tests on Mac/Intel to the release checklist as a manual item maybe ??

Yeah, I kind of dislike adding more steps to the release checklist actually... maybe we could add another GitHub Actions workflow for this, which is triggered manually? Like: a more complete matrix from 3.7 to 3.12 on all OSes, and including the `slow`-marked tests? In a way that workflow would be easier to maintain, as it doesn't need to care about publishing images, packages, docs or things like that... I've done a quick version of this in #1603

> Just another thought -- should we also drop Python 3.9 or 3.10?

(tongue in cheek: maybe we should drop windows support, as that's the slowest jobs by far, now that we've dropped intel mac)

More seriously, I think it might make sense to kind of align with `numpy` on these issues, which right now supports 3.9 - 3.12.

I can actually think of a few jobs we could _add_ that would make sense. For example: for each Python version, install the lowest possible versions of our direct dependencies (with `uv`, we can use `--resolution=lowest-direct`). That would make sure we have correct lower version constraints in our packaging metadata. Then, in addition, we would benefit from automated tests with `cupy` installed and a GPU available. And of course, continuous benchmarks...

For these, I was thinking of using the JSC CI service, if only for thorough pre-release checking. There, we have potentially much larger parallelism available than here, even though we can only run on Linux...

On the other hand, I think a significant amount of time is spent jit-compiling `numba` stuff, so if we manage to switch our I/O code to ahead-of-time compiled code, maybe that is the smarter way to go, as that would speed up not only the CI but also the production code...

PR #1603, opened 2024-02-23
Add thorough manual testing workflow

sk1p:
Already caught issues:

- [x] Update clustering analysis for current `sklearn`
- [x] ~~Need to apply event loop fixes for Python 3.8, too?~~

## TODO

- [ ] For testing, this still triggers on pull requests, to be removed before it's merged.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/.github/workflows/thorough.yml
+++ b/.github/workflows/thorough.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   tests:
     strategy:
+      fail-fast: false
       matrix:
         python_version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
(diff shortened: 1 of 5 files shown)


sk1p:
> If I suppress the error inside `aiohttp` `TimerContext` then I get an error relating to a `Task got Future attached to different event loop`... I'm a bit out of my depth here but reporting just in case.
>[...]

That's a very good hint - I think this has to do with different scopes for the asyncio http client vs the event loop somehow (this is fixed for newer pytest-asyncio versions, which we can't have with py38...).

uellue:
LGTM! Merging

PR #1604, opened 2024-02-26
Work around "UDF run cancelled after 0 partitions"

sk1p:
Cherry-picked from #1578 

Example failure: https://github.com/LiberTEM/LiberTEM/actions/runs/8024972705/job/21924629028

This could be reproduced just by running `test_fd_limit` with a current `dask` version. The different behavior was introduced in dask 2024.2.1 - not sure yet if this is a bug or not, as I haven't managed to minimize the testcase yet.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/executor/dask.py
+++ b/src/libertem/executor/dask.py
@@ -451,7 +451,7 @@ class DaskJobExecutor(CommonDaskMixin, BaseJobExecutor):
         # redirect keys to new values
         handle = str(uuid.uuid4())
         try:
-            fut = self.client.scatter(obj, broadcast=True)
+            fut = self.client.scatter(obj, broadcast=True, hash=False)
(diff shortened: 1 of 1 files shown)


uellue (approved):
LGTM!

It is not so clear if hashing is faster or slower than not hashing anyway, it depends on the circumstances.

uellue:
Discussion with @sk1p: Generally a life time management / reference counting issue with items in the cluster that share the same hash. Attaching a new handle to a pre-existing item can overlap with disposing of said item. If everything gets a new hash always, their life cycles are not connected which keeps it simple. Combining hashing and reusing items in a distributed Dask cluster with garbage collection seems like a complex transaction, better to avoid.

PR #1608, opened 2024-03-01
Exposed Y and X field shifts in CoMUDF

sivborg:
This is only a small change in the :class:`~libertem.udf.com.CoMUDF` class. It exposes the y- and x-components of the `field` result. The reason is so that they are more easy to plot and access, which is already shown to be useful in cases where the other results do not give good information. For example in my case where the entire dataset is magnetized with same magnitude, but different orientation.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/docs/source/changelog/misc/field-components.rst
+++ b/docs/source/changelog/misc/field-components.rst
@@ -2,4 +2,4 @@
 ==============================================
 
 * In :class:`~libertem.udf.com.CoMUDF`, added `field_y` and `field_x` to results, 
-  which are slices of their respective components in the `field` result.
+  which are slices of their respective components in the `field` result (:pr:`1608`).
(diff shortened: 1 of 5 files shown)


docs/source/changelog/misc/field-components.rst
```diff
@@ -0,0 +1,5 @@
+[Feature] Expose y- and x-components in CoMUDF
+==============================================
+
+* In :class:`~libertem.udf.com.CoMUDF`, added `field_y` and `field_x` to results, 
+  which are slices of their respective components in the `field` result.
```
uellue, line 5:
```suggestion
  which are slices of their respective components in the `field` result (:pr:`1608`).
```

uellue (approved):
Thx, LGTM! I'll go ahead and include the PR in the changelog, run test and merge.

PR #1620, opened 2024-04-08
Fixup for malformed EMPAD data

### LINKED ISSUE #1617: Support for malformed EMPAD files
The files in https://er-c-data-mgmt.fz-juelich.de/browse/?path=/adhoc/libertem/libertem-test-data/EMPAD/Magical_DataSet#browse don't work with with current EMPAD reader as deployed on moellenstedt.

sk1p:
For some EMPAD data sets, the raw file corresponds to the shape of the scan parameters with `mode="search"`, not `mode="acquire"`. That means we were expecting more data than actually available.

Note that this means we are more strict in what we accept, in particular this checks (if we are opening via the XML file) that the raw file is exactly as large as specified in the XML file (one of the two `scan_parameters`). Previously, it was probably possible to have a raw file that is larger than the specified scan in the XML file, and we would have probably accepted that.

Fixes #1617 

@uellue is the `Magical_DataSet` going to stay in our test data, or should we extract an example file?

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/docs/source/changelog/bugfix/empad-scan-params.rst
+++ b/docs/source/changelog/bugfix/empad-scan-params.rst
@@ -1,0 +1,7 @@
+[Bugfix] Fix loading of malformed EMPAD data
+============================================
+
+* For some data sets, the actual raw data acquired corresponds to the
+  :code:`mode="search"` scan parameters, and not :code:`mode="acquire"`. This
+  is probably a bug in the acquisition software, but we want to be able to load
+  these files anyways (:issue:`1617`, :pr:`1620`).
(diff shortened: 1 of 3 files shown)


uellue (approved):
Thank you for the fix! I've modified it a bit so that both the main and alternate version are included in the error message. I've also added a test case to that end.

That includes adding a test file version where the size doesn't fit on purpose.

Since the files are interesting for this test, I've kept acquisition_1 and deleted all the other ones, including unrelated stuff about a tilt series.

uellue:
The files were actually saved with "search" settings in the GUI. Apparently the file size and file name are the main indication which set of parameters are valid.

In conclusion, the files were not malformed, and the fix here implements the correct method to determine parameters.

PR #1618, opened 2024-03-27
Adjust event loop on Windows when launching libertem-server

matbryan52:
Closes #1616 

Seems like the web server also needs coaxing into using `WindowsSelectorEventLoopPolicy`. Without this change we had a `ProactorEventLoop` despite the `Selector` policy being returned from `asyncio.get_event_loop_policy()` (??).

Probably needs more thought later, but if you can check on your Windows 11 machine @uellue we could merge as a quick fix? I've checked this now works on Windows 10, Windows Server 2016 and a Ubuntu Server 20.

This was not caught by the smoke `test_cli.py` case because it only checks for a synchronous `log.info`, we could make it actually `curl` the server address to check if we actually get a response!

## Contributor Checklist:

* [x] I have added or updated my entry in [the creators.json file](https://github.com/LiberTEM/LiberTEM/blob/master/packaging/creators.json)
* [ ] I have added [a changelog entry](https://github.com/LiberTEM/LiberTEM/tree/master/docs/source/changelog) for my contribution
* [ ] I have added/updated documentation for all user-facing changes
* [ ] I have added/updated test cases
* [ ] I have included the [rebuilt production build of the client](https://libertem.github.io/LiberTEM/contributing.html?#building-the-client) (only if changes were made to the GUI)

## Reviewer Checklist:

* [ ] `/azp run libertem.libertem-data` passed
* [x] No import of GPL code from MIT code

PR #1622, opened 2024-04-09
Add `libertem-server` `test_cli.py` GET

matbryan52:
Followup to #1618 by checking we get a 200 response to a GET request to `libertem-server`. Basic but is an improvement on no check. Could go further by checking the response content for some likely strings.

I don't know if we'd like to have a shorter timeout? The `http_client` fixture seems to have a default 120 second timeout.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/tests/server/test_cli.py
+++ b/tests/server/test_cli.py
@@ -37,10 +37,6 @@ async def test_libertem_server_cli_startup(http_client):
         if 'LiberTEM listening on' in line:
             break
 
-    addr = line[line.find("http"):].strip()
-    async with http_client.get(addr, timeout=30.0) as response:
-        assert response.status == 200, "Failed to GET from libertem-server"
-
     async def _debug():
         while True:
             line = await asyncio.wait_for(p.stderr.readline(), 5)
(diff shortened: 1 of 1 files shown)


sk1p (approved):
> Followup to #1618 by checking we get a 200 response to a GET request to `libertem-server`. Basic but is an improvement on no check. Could go further by checking the response content for some likely strings.

:+1: I'm already happy with a simple check.

> I don't know if we'd like to have a shorter timeout? The `http_client` fixture seems to have a default 120 second timeout.

Hmm, this is a trade-off between making the test flaky in case of very slow CI, and making the error case fail faster. Maybe 30 seconds is fast enough, and still likely enough to succeed in case of slow CI?

tests/server/test_cli.py
```diff
@@ -37,6 +37,10 @@ async def test_libertem_server_cli_startup():
         if 'LiberTEM listening on' in line:
             break
 
+    addr = line[line.find("http"):].strip()
+    async with http_client.get(addr) as response:
```
sk1p, line 41: I think this should work:

```suggestion
    async with http_client.get(addr, timeout=30.0) as response:
```

matbryan52: If the check were moved into the `try/except/finally` block below would that do the job? The assert would send us via `except` i.e. `p.terminate() ; p.kill()`

sk1p: Yup, that seems to work! Tested by setting the timeout to a very small non-zero value, and no `libertem-server` seems to be running afterwards.

PR #1624, opened 2024-04-10
Update license to GPL 3.0 ~~or later~~ only

### LINKED ISSUE #1553: Fix license to GPL-3 <del>or later</del> only in Zenodo template
The current template gives GPL-2 only.

Should also be updated in other projects that use the same template mechanism.

uellue:
The current one maps to GPL 3 only.

Closes #1553 

List of valid license identifier string obtained by

```
import requests
response = requests.get('https://sandbox.zenodo.org/api/licenses/?size=1000')
print(response.json())
```

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/packaging/zenodo-upload.json
+++ b/packaging/zenodo-upload.json
@@ -18,7 +18,7 @@
                 }
             ],
             "access_right": "open",
-            "license": "gpl-3.0-or-later",
+            "license": "gpl-3.0-only",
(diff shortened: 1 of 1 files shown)


sk1p:
#1553 says:

> The current template gives GPL-2 only.

Which I agree would be incorrect.

But here:

> The current one maps to GPL 3 only.

We actually don't specify the "or later" for the license. For example, from the README:

> LiberTEM is licensed under GPLv3. The I/O parts are also available under the MIT license, please see LICENSE files in the subdirectories for details.

And in the `setup.py`:

```
    license='GPL v3',
[...]
        'License :: OSI Approved :: GNU General Public License v3 (GPLv3)',
```

So, I think this change might be incorrect.

uellue:
Good point! I've updated this PR to 3.0 only. We'll see with the next RC if this is right

PR #1625, opened 2024-04-18
Update release procedure

sk1p:
* Update and re-structure release checklist
* No longer use zenodo sandbox and test.pypi.org; instead upload directly to production deployments (streamlines things a bit, and makes installing an RC easier; also probably more stable, especially regarding wipes of the zenodo sandbox)
* Add missing Python 3.12 classifier

## Contributor Checklist:

## Reviewer Checklist:


matbryan52 (approved):
I like the changes to the release checklist. I would question whether these two are still practical given the number of datasets and the potential number of ways the system could fail:

```
* [ ] Check what happens when trying to open non-existent files by scripting.
    * [ ] Proper understandable error message? TODO automate?
* [ ] Check what happens when opening all file types with bad parameters by scripting
    * [ ] Proper understandable error message? TODO automate?
 ```

We could say that automating these tasks should actually happen for 0.14 - I will try to write a `/data` test which actually collects a bunch of error messages that a human could quickly read over...

sk1p:
>  I would question whether these two are still practical given the number of datasets and the potential number of ways the system could fail:

Agreed...

> We could say that automating these tasks should actually happen for 0.14

:+1: yeah, not automating something just means it will be unreliably (or not) tested before releases.

uellue (approved):
:+1: :+1: :+1:

PR #1627, opened 2024-04-18
Section on air-gapped installation

uellue:
## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -181,6 +181,8 @@ running the LiberTEM GUI. In other environments, like Jupyter notebooks, you
 need to restart the Python interpreter to make sure the new version is used,
 for example by restarting the ipython kernel.
 
+.. _`airgapped`:
+
 Air-gapped installation
 -----------------------
 
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -181,6 +181,42 @@ running the LiberTEM GUI. In other environments, like Jupyter notebooks, you
 need to restart the Python interpreter to make sure the new version is used,
 for example by restarting the ipython kernel.
 
+Air-gapped installation
+-----------------------
+
+Many microscope control computers are not connected to the internet, which means
+that the usual installation methods don't work. It is not straightforward to
+package a Python application into a self-contained executable or installer for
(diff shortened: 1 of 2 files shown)


docs/source/install.rst
```diff
@@ -181,6 +181,44 @@ running the LiberTEM GUI. In other environments, like Jupyter notebooks, you
...
+On the target
+^^^^^^^^^^^^^
+
+* If necessary, install the same Python version as on the host.
+* Start a new command shell and confirm the Python version: :code:`python --version`
+* Create a virtual environment in a folder of your choice: :code:`python -m venv libertem-venv`
+* Activate the environment: On Windows cmd.exe :code:`libertem-venv\\Scripts\\activate.bat`
+* Change to the directory with the Python packages.
+* Install these packages without dependencies or using a package index:
+  On Windows cmd.exe :code:`FOR %a in (*.whl) DO python -m pip install --no-index --no-deps %a`
```
sk1p, line 220: Hm, this depends on the wheels folder being clean, i.e. no extra packages or multiple versions of the same package being in there. There was a similar way of doing this, using `--find-links`, that would allow you to use the dependency machinery of pip with this method, making sure you get compatible versions. Is there any reason to not use `--find-links` (or whatever replacement there is on modern `pip`)?

docs/source/install.rst
```diff
@@ -181,6 +181,44 @@ running the LiberTEM GUI. In other environments, like Jupyter notebooks, you
...
+work reliably. In order to install LiberTEM on an air-gapped machine, you need a
+computer with the same operating system, architecture and Python version as a
+host system where you can prepare all required packages for the target system.
+
+On the host
+^^^^^^^^^^^
+
+* Create a folder where you collect all required files, for example :code:`wheels`.
+* If necessary, download and install a recent version of Python supported by LiberTEM from
+  https://www.python.org/downloads/. This installer works without internet access.
```
sk1p, line 203: Maybe we should explicitly say that this version of course has to be compatible with the operating system version you will install on, which could be older on air-gapped systems (as we depend on Python 3.9, that means currently Windows 8+ IIUC)

sk1p (approved):

PR #1628, opened 2024-04-19
WIP PyInstaller progress

uellue:
Current status: libertem-server starts up, but doesn't find the static files where they are expected

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/packaging/pyinstaller/libertem-server.spec
+++ b/packaging/pyinstaller/libertem-server.spec
@@ -1,0 +1,68 @@
+# -*- mode: python ; coding: utf-8 -*-
+
+import sys ; sys.setrecursionlimit(sys.getrecursionlimit() * 5)
+
+dask_datas = []
+
+dask_files = [
+    ('Lib/site-packages/dask/dask.yaml', 'dask'),
+    ('Lib/site-packages/distributed/distributed.yaml', 'distributed')
+]
+
+for p in sys.path:
+    for d, target in dask_files:
+        path = os.path.join(p, d)
+        if os.path.exists(path):
+            dask_datas.append((path, target))
+
+libertem_datas = [('../../client', 'libertem')]
+
+datas = []
+
+try:
+    import opentelemetry.sdk
+    datas += copy_metadata('opentelemetry.sdk')
+except ModuleNotFoundError:
+    pass
+
+a = Analysis(
+    ['libertem-server.py'],
+    pathex=[],
+    binaries=[],
+    datas=datas + dask_datas + libertem_datas,
+    hiddenimports=[],
+    hookspath=['.\\extra-hooks\\'],
+    hooksconfig={},
+    runtime_hooks=[],
+    excludes=[],
+    noarchive=False,
+    optimize=0,
+)
+pyz = PYZ(a.pure)
+
+exe = EXE(
+    pyz,
+    a.scripts,
+    [],
+    exclude_binaries=True,
+    name='libertem-server',
+    debug=False,
+    bootloader_ignore_signals=False,
+    strip=False,
+    upx=True,
+    console=True,
+    disable_windowed_traceback=False,
+    argv_emulation=False,
+    target_arch=None,
+    codesign_identity=None,
+    entitlements_file=None,
+)
+coll = COLLECT(
+    exe,
+    a.binaries,
+    a.datas,
+    strip=False,
+    upx=True,
+    upx_exclude=[],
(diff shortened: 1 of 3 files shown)


packaging/pyinstaller/libertem-server.spec
```diff
@@ -0,0 +1,68 @@
+# -*- mode: python ; coding: utf-8 -*-
+
+import sys ; sys.setrecursionlimit(sys.getrecursionlimit() * 5)
+
+dask_datas = []
+
+dask_files = [
+    ('Lib/site-packages/dask/dask.yaml', 'dask'),
```
rokm, line 8: This should be unnecessary, as [existing hook for dask](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-dask.py) should collect that file.

packaging/pyinstaller/libertem-server.spec
```diff
@@ -0,0 +1,68 @@
...
+    ('Lib/site-packages/distributed/distributed.yaml', 'distributed')
```
rokm, line 9: This could be handled in more portable way by creating a hook for `distributed` that calls `collect_data_files('distributed')`.

rokm:
And to avoid a future issue report - if your code uses a package that in turn uses `multiprocessing` (I think `dask.distributed` / `distributed` fits the bill), make sure to call `multiprocessing.freeze_support()` at the start of your program. Otherwise, use of `multiprocessing` will likely result in endless program spawn loop. See https://pyinstaller.org/en/stable/common-issues-and-pitfalls.html#multi-processing

packaging/pyinstaller/libertem-server.spec
```diff
@@ -0,0 +1,68 @@
...
+]
+
+for p in sys.path:
+    for d, target in dask_files:
+        path = os.path.join(p, d)
+        if os.path.exists(path):
+            dask_datas.append((path, target))
+
+libertem_datas = [('../../client', 'libertem')]
```
sk1p, line 18: > Current status: libertem-server starts up, but doesn't find the static files where they are expected

Do they end up in the `site-packages` in the installation? Is the relative path correct (whats the working directory when this code runs?), or does it maybe need a `os.path.join(os.path.dirname(__file__), '..', '..', 'client')` or similar?

uellue:
@rokm thank you for the pointers, much appreciated!

packaging/pyinstaller/libertem-server.spec
rokm: If relative paths are given like that, PyInstaller assumes they are relative to the spec file (in order to factor out current working directory during the build).

I think part of the problem here is that the contents of that `client` directory are collected into `libertem` directory, as instructed. The target directory should be `libertem\web\client`, as it is when the package is installed.

But if the build process here assumes that `libertem` is installed (in non-editable mode) in the python environment prior to the build (which I suspect is the case, otherwise PyInstaller will have no way of finding `libertem` modules...), then the data files could be simply collected by `hook-libertem.py`:
```python
from PyInstaller.utils.hooks import collect_data_files
datas = collect_data_files('libertem')
```

The second part of the problem seems to be that in frozen application, `tornado.template.Loader` receives relative `root_directory` (`libertem\web`), which it incorrectly resolves to absolute path using current working directory. When running unfrozen, that `root_directory` is already a correct absolute path. Will need to dig around a bit more to see what exactly is going wrong...

rokm: > The second part of the problem seems to be that in frozen application, `tornado.template.Loader` receives relative `root_directory` (`libertem\web`), which it incorrectly resolves to absolute path using current working directory. When running unfrozen, that `root_directory` is already a correct absolute path. Will need to dig around a bit more to see what exactly is going wrong...

Ah yes, [`tornado` is trying to infer the path from `co_filename` of code objects](https://github.com/tornadoweb/tornado/blob/master/tornado/web.py#L1076-L1083), which are anonymized relative paths in PyInstaller builds.

I suppose the easiest way for you to work around this insanity is to set `template_path` in the application settings, similarly to how it is done for [`static_path`](https://github.com/uellue/LiberTEM/blob/e20a2495cc49aa7af83f448c07fdce70bf57d7df/src/libertem/web/server.py#L100). In this case, the value should be `os.path.dirname(__file__)`.

PR #1635, opened 2024-05-03
PickShiftedUDF

sivborg:
## Contributor Checklist:

## Reviewer Checklist:

This is an addition of a new PickUDF that is able to shift the data that is retrieved. This is useful because it makes it possible to correct raw data for unintended shifts, for example descan which can obstruct electromagnetic signal in STEM-DPC. This functions as the previous PickUDF, but you can give corrective regression parameters indicating how much to shift the data. This can be found for example by using the CoMUDF:
```
from libertem import api
from libertem.udf.com import CoMUDF
from libertem.udf.raw import PickShiftedUDF

ctx = api.Context()

dataset = ctx.load('dask', dask_array=s.data, sig_dims=2)

comudf = CoMUDF.with_params(cy=128, cx=128, r=1000, regression=1)
comresult = ctx.run_udf(dataset=dataset, udf=comudf)

pickudf = PickShiftedUDF(regression_coefficients=comresult["regression"].data)
result = ctx.run_udf(udf=pickudf, dataset=dataset, roi=(10,10))
```

The way this is implemented is with `scipy.ndimage.shift`, done as a postprocessing step in the UDF. The shifts are found with the `self.meta.coordinates` attribute, which is passed around as a result buffer. So perhaps there is a cleaner way of doing this. The UDF seems to work well for my practical purposes and I have set up some accompanying tests.


src/libertem/udf/raw.py
```diff
@@ -75,3 +76,107 @@ def merge_all(self, ordered_results):
...
+    regression_coefficients : (3, 2) matrix
+        A vector of regression coefficients for x- and y-directions.
+        The regression coefficients make up two planes that describe
+        the descan and are used to remove it.
+        The coefficients are arranged in the following manner:
+        ((y, x), (dy/dy, dx/dy), (dy/dx, dx/dx)) where y and x are
+        the shifts at coordinates (0, 0). The displayed image will be
+        shifted in x- and y-directions by these parameters. By default
+        `None`, which corresponds to no shifts.
+    '''
```
uellue, line 99: What about supporting the same options as `CoMUDF`, or even use the `CoMParams` class: https://github.com/LiberTEM/LiberTEM/blob/59529c19ca4fcaac63aef25d524c80e3389e97c9/src/libertem/udf/com.py#L30-L44

`scan_rotation` and `flip_y` would apply perfectly for this use case and should be easy to implement with the `scipy.ndimage` functions.

In particular when dealing with DPC, it can help if in the picked frames everything points in the right directions relative to the scan! By using the same parameters, one can use them for both. As a side effect one could use this to create a corrected dataset! Could be merged with corresponding code in pyxem as a perspective?

src/libertem/udf/raw.py
```diff
@@ -75,3 +76,107 @@ def merge_all(self, ordered_results):
...
+            for intens, shift in zip(intensity, shifts):
+                shifted.append(scipy.ndimage.shift(intens, -shift, mode="constant"))
+
+            if shifted:
+                intensity = np.stack(shifted)
+
+        return {
+            "intensity": intensity,
+            "coordinates": self.results.get_buffer('coordinates').raw_data
+        }
```
uellue, line 152: What about using `process_frame()` and doing all this there? It would be easier to read, and the processing would be performed on the workers, not on the main process.

Later one could look into using `process_partition()` or `postprocess()` to process larger chunks, which would be beneficial for using GPUs.

tests/udf/test_pick.py
```diff
@@ -44,3 +44,47 @@ def test_pick_empty_roi(lt_ctx, delayed_ctx):
...
+    res_delayed = delayed_ctx.run_udf(dataset=dataset, udf=udf, roi=roi)
+
+    assert np.allclose(data[roi], res['intensity'].data)
+    assert np.allclose(data[roi], res_delayed['intensity'].data)
+
+    assert data[roi].shape == res['intensity'].data.shape
+    assert data[roi].shape == res_delayed['intensity'].data.shape
+
+    assert data.dtype == res['intensity'].data.dtype
+    assert data.dtype == res_delayed['intensity'].data.dtype
```
uellue, line 90: The regression coefficients should all be different to make sure we are not mixing them anywhere. In particular, the dy/dx ... coefficients require tests.

We should also confirm that the coefficients act the same as in `CoMUDF`, in particular if we support more parameters, as suggested above. A simple and comprehensive test would be comparing `CoMUDF` on an uncorrected dataset vs. `CoMUDF` on a corrected dataset.

src/libertem/udf/raw.py
```diff
@@ -75,3 +76,107 @@ def merge_all(self, ordered_results):
         intensity_chunks = [b.intensity for b in ordered_results.values()]
         intensity_sum = np.stack(intensity_chunks, axis=0).sum(axis=0)
         return {'intensity': intensity_sum}
+
+
+class PickShiftedUDF(UDF):
+    '''
```
uellue, line 82:
```suggestion
    '''
    .. versionadded:: 0.15.0

```

src/libertem/udf/raw.py
```diff
@@ -26,8 +27,8 @@ class PickUDF(UDF):
     >>> result["intensity"].raw_data[0].shape
     (32, 32)
     '''
-    def __init__(self):
-        super().__init__()
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
```
uellue, line 31: Hm, why this change? The `PickShiftedUDF` is subclassed from `UDF`, so `PickUDF` doesn't need to pass parameters through. If one wanted a common subclass to share code, it would be cleaner to have a `BasePickUDF` with the common parts, and derive both `PickUDF` and `PickShiftedUDF` from that, each with a final constructor. That way the parameters are clear for autocompletion etc.

uellue (changes requested):
Looks like a good idea and a good start!

For changelog you can add a file in `docs/source/changelog/features/`. There are already two, you can just follow the pattern. :-)

PR #1656, opened 2024-06-18
Fix compatibility with numpy>=2.0.0

### LINKED ISSUE #1653: `numpy` v2 compatibility
Now that `numba` v0.60 is released, we can start to support `numpy` v2. I'm working on a PR with the necessary changes.

To keep libertem installable, we should do a point release ASAP.

sk1p:
Fixes #1653

 * npy: use `ast.literal_eval` instead of `numpy.lib.utils.safe_eval`
 * k2is: fix encode test case; simplify encoder function
 * dm: convert image count from `np.uint32` to Python integer, as it is used in arithmetics which might not fit into the original dtype
 * mib,raw: update decoders/encoders to explicitly use fitting dtypes for shifted `uint8`s

Note that the notebook tests and the docs checks don't currently run with numpy 2 because of incompatibilities in dependencies (hyperspy->pint and scikit-image), so I've constrained these to `numpy<2` for now.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/docs/source/changelog/features/numpy2.rst
+++ b/docs/source/changelog/features/numpy2.rst
@@ -1,0 +1,6 @@
+[Feature] numpy 2 compatibility
+===============================
+
+* Support for numpy 2.x. Note that some examples may not yet run, as the
+  dependencies might not be numpy 2 compatible. For these cases, please
+  downgrade to :code:`numpy<2` (:pr:`1656`, :issue:`1653`).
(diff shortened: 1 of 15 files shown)


uellue:
`pytest -m "not dist" tests` passes with data and CuPy on Python 3.12 with numpy 2 on Linux and Win. LGTM, thx for taking care!

PR #1663, opened 2024-06-19
Remove compatibility code

uellue:
Follow https://github.com/numpy/numpy/blob/main/numpy/lib/format.py where this code was copied from.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/io/dataset/npy.py
+++ b/src/libertem/io/dataset/npy.py
@@ -1,13 +1,10 @@
-import io
 import os
-import sys
 import typing
 from typing import Optional
 import logging
 
 import numpy as np
-from ast import literal_eval
-from numpy.lib.format import read_magic
+from numpy.lib.format import open_memmap
 from libertem.common.messageconverter import MessageConverter
 
 from libertem.io.dataset.base import (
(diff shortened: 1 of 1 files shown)


uellue:
Looking at https://github.com/numpy/numpy/blob/main/numpy/lib/format.py, we won't get away so easily. The array format changed with NumPy 2.0, so this will be a bigger project to make sure we support old and new .npy.

matbryan52:
> Looking at https://github.com/numpy/numpy/blob/main/numpy/lib/format.py, we won't get away so easily. The array format changed with NumPy 2.0, so this will be a bigger project to make sure we support old and new .npy.

Will hopefully not take too long to do, and I would imagine we won't encounter newer format files often. I get the impression they are using the older format whenever possible for compatibility and only using the newer for unicode header names or many column structures.

uellue:
> > Looking at https://github.com/numpy/numpy/blob/main/numpy/lib/format.py, we won't get away so easily. The array format changed with NumPy 2.0, so this will be a bigger project to make sure we support old and new .npy.
> 
> Will hopefully not take too long to do, and I would imagine we won't encounter newer format files often. I get the impression they are using the older format whenever possible for compatibility and only using the newer for unicode header names or many column structures.

Ah, good to know! Thinking about it, since we are creating .npy files with NumPy 2 in the tests, we should be more or less fine as long as they pass.

uellue:
Instead of copying the functions, perhaps we can memory map the file, get the parameters and immediately close the memory map? I'd expect/hope that memory mapping is a noop until actually accessing the memory region.

sk1p (approved):
LGTM! I didn't see the new npy file format though - as far as I can see, format 3.0 is the latest and was added in numpy 1.17:

https://github.com/numpy/numpy/blob/594657d38b12b1fdc7f620ed026c8df93f271cb3/numpy/lib/format.py#L426-L429

Anyway, using `open_memmap`, we should be be as compatible as we can.

Issue #1649, opened 2024-06-07
Permissive licensing


uellue:
Discussion with JK Weiss, TESCAN:

If LiberTEM was under a permissive license like MIT, it would allow integration in TESCAN's software. That would be of concrete interest for TESCAN.

Since we have a lot of code contributed by a small number of contributors, and we can probably contact a large percentage of our contributors, re-licensing would be a realistic option.

uellue:
Assigning to the next milestone to make sure we follow up in a timely manner.

uellue:
```
    install_requires=[
        "numpy", # 3-Clause BSD
        "scipy>=1.4.1", # 3-Clause BSD
        "sparse", # 3-Clause BSD License
        "distributed>=2.19.0", # 3-Clause BSD License
        "dask!=2023.6.1",  # 3-Clause BSD License
        "click", # 3-Clause BSD License
        "tornado>=5", # Apache 2.0
        "matplotlib", # Matplotlib only uses BSD compatible code, and its license is based on the [PSF](https://docs.python.org/3/license.html) license
        "pillow", # Like PIL, Pillow is licensed under the open source HPND License (permissive)
        "h5py", # 3-Clause BSD License
        "psutil", # 3-Clause BSD License
        # https://github.com/LiberTEM/LiberTEM/issues/763
        # https://github.com/LiberTEM/LiberTEM/issues/783
        "numba>=0.51", # BSD 2-Clause "Simplified" License
        "ncempy>=1.10", # !!! GPL-3
        'pywin32!=226;platform_system=="Windows"', # 3-Clause BSD License
        # FIXME pull request #259
        # https://github.com/LiberTEM/LiberTEM/pull/259#discussion_r251877431
        'scikit-image', # 3-Clause BSD License
        'cloudpickle', # 3-Clause BSD License
        'jsonschema', # MIT License
        'scikit-learn', # 3-Clause BSD License
        'tqdm', # MIT "it's complicated" https://github.com/tqdm/tqdm/blob/master/LICENCE
        'threadpoolctl>=3.0', # 3-Clause BSD License
        'nbformat', # 3-Clause BSD License
        'nbconvert', # 3-Clause BSD License
        'autopep8', # MIT License
        'empyre>=0.3.0', # !!! EMPyRe is licensed under [GPLv3](https://iffgit.fz-juelich.de/empyre/empyre/-/blob/master/LICENSE).
        'defusedxml', # Python Software Foundation License (PSFL)
[...]

PR #1687, opened 2024-10-23
Replace colormap and remove empyre dependency

uellue:
Replace vector field colormap from `empyre` with custom implementation based on `colorcet` and LUV color space manipulation to remove dependency on GPL code. See also #1649

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/viz/base.py
+++ b/src/libertem/viz/base.py
@@ -111,9 +111,7 @@ def rgb_from_2dvector(y, x, vmax=None):
 
 
 def make_colormap():
-    # including 360 to close the circle: it interpolates
-    # in the interval [0, 1] and color(0) == color(1)
-    angles = np.arange(361)
+    angles = np.arange(360)
     angles_rad = angles/180*np.pi
     yy, xx = -np.sin(angles_rad), -np.cos(angles_rad)
     cols = rgb_from_2dvector(y=yy, x=xx) / 255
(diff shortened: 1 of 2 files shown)


uellue:
This modifies semi-official code since our templates import color maps from `libertem.viz` that are not part of the official API. The `cmaps` variable and vector color maps were used only in a very few places, and using the 3D vectors with 2D data always looked a bit awkward to me. For that reason I propose to replace `cmaps` with a single isoluminant cyclic color map and CMAP_CIRCULAR_DEFAULT with the `rgb_from_2dvector()` function. However, that breaks notebooks that were generated from our templates. Not sure if one should keep the previous API for compatibility!

uellue:
Comparison of old and new color map, I tried to tweak it to be similar:

Figure 2

uellue:
TODO: Color map for angle and value has different colors for the same direction:

Figure 4
Figure 5(1)

Possibly the same function and ajdustments can be used for both. In any case, the previous angle plotting code was incorrect because it didn't set vmin to -np.pi and vmax to np.pi

uellue:
Figure 5(2)

Phase with new colormap, same colors as complex_0_1

uellue:
The async notebook seems to hang. It runs finw with `tox` locally.

sk1p (approved):
LGTM!

PR #1700, opened 2025-02-20
Ensure we properly close internal generators/iterators

sk1p:
If we don't, we may swallow exceptions. This should improve reliability (especially the hangs observed recently with Python 3.13).

See #1697 for some additional analysis and stack traces.

Without this, running `pytest tests/executor/test_pipelined.py -n12` on Python 3.13 hung quite reproducibly in `CommsDispatcher.__exit__`, waiting for the `monitor_queue` thread to return. I'm not 100% sure how this hang came to be, but this appears to be fixed by properly propagating exceptions.

(what brought me to this solution: I broke the deadlock by making the `_message_q.get` in `CommsDispatcher.monitor_queue` into a non-blocking variant, and exiting the thread via an event. This showed messages like "Exception ignored in: <generator object ...>", which points to generators not being fully consumed or [closed](https://docs.python.org/3/reference/expressions.html#generator.close), causing these ignored exceptions when the generator is garbage collected)

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/udf/base.py
+++ b/src/libertem/udf/base.py
@@ -2581,15 +2581,19 @@ class UDFRunner:
                     yield params_handle
 
                     if tasks:
-                        for res in executor.run_tasks(
+                        task_gen = executor.run_tasks(
                             tasks,
                             params_handle,
                             cancel_id,
                             task_comm_handler,
-                        ):
-                            if progress:
-                                pman.finalize_task(res[1])
-                            yield res
+                        )
+                        try:
+                            for res in task_gen:
+                                if progress:
+                                    pman.finalize_task(res[1])
+                                yield res
+                        finally:
+                            task_gen.close()
             finally:
                 if progress and tasks and pman is not None:
                     pman.close()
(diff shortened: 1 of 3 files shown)


src/libertem/udf/base.py
```diff
@@ -2581,15 +2581,19 @@ def _inner():
...
                             task_comm_handler,
-                        ):
-                            if progress:
-                                pman.finalize_task(res[1])
-                            yield res
+                        )
+                        try:
+                            for res in task_gen:
+                                if progress:
+                                    pman.finalize_task(res[1])
```
sk1p, line 2593: I'm not sure why this isn't covered, even with the additional test case; `progress` should be passed down properly, and the next line has coverage...

uellue: When I add a `raise Exception` before that line in the same `if` statement, `tests/executor/test_pipelined.py::test_with_progress` fails exactly there. I'd say it is definitely covered provided that test runs in CI (and why shouldn't it), and Codecov is mistaken.

uellue (approved):
LGTM, thx! :-)

Issue #1701, opened 2025-02-21
Improve partitioning of "sig-heavy" data


sk1p:
Sig-heavy meaning here data that has shapes like `(10, 4096, 4096)` instead of the usual 4D-STEM shapes where we have many more scan points. Right now, working on such a data set gives a message like the following:

```
[...]/site-packages/libertem/io/dataset/base/partition.py:75: RuntimeWarning: dataset contains fewer frames than partitions, setting num_part to 1 to allow processing (use fewer workers?)
  warnings.warn(
```

For reference, this was introduced in https://github.com/LiberTEM/LiberTEM/pull/1337 

As we have quite a lot of work to do in some cases per-frame, that means we are missing out in parallelization here; it could be 10x faster at least (maybe even more if we have mostly GPU workers). In other cases, we may have little work per-frame, meaning splitting "small" partitions into even smaller ones results in more overheads.

There are different approaches:

1) Have a tweakable (maybe when loading data) to let the user influence the partitioning (fixes the issue for advanced users)
2) By default, go the other way and split the data into single-frame partitions (set [`num_partitions=num_frames` instead of `num_partitions=1`](https://github.com/LiberTEM/LiberTEM/pull/1337/files#diff-e4757907cb1d6345ded30677b85509ddeea92ea5be743c62b48f0945e37f4c36R70-R76))
3) Have a heuristic decide this
4) Have the UDF communicate this (needs a new mechanism similar to tile shape negotiation)

For the short term, I'm tending towards 2), and maybe with a later addition of 1) to allow to override this decision.

sk1p:
I'll actually go ahead with a minimal PR for option 2 for discussion and possibly benchmarking of overheads etc.

matbryan52:
I think I chose the `num_partitions = 1` because it followed the existing heuristic that a partition should be around 512 MB, so even if you have many workers it didn't make sense from an overheads perspective to split a small dataset up beyond this. In the case of an image stack the existing heuristic is not the right one, and can only really be decided from the work that needs to be done on the data. I can imagine in some cases that setting `num_partitions == num_workers` is optimal, too !

The reality is that any 4D-STEM dataset which was small enough to run into the `num_partitions = 1` branch is trivially quick to compute on, anyway, so it doesn't really matter that this will now launch `num_frames` tasks instead. You'll have to wait for the overhead but this should be a few seconds maximum, then the trivial processing time.

[...]

PR #1702, opened 2025-02-21
Allow small partitions; add manual override

sk1p:
In case a `DataSet` has less frames than partitions (i.e. workers), we previously combined the frames into a single partition. This is a performance issue if there is a lot of work per-frame. Instead, create one partition per frame so we get more parallelization.

Also add a manual override in case the automatically chosen number of partition is inappropriate.

See #1701 for some initial discussion and other options.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/docs/source/changelog/features/num-partitions.rst
+++ b/docs/source/changelog/features/num-partitions.rst
@@ -4,7 +4,7 @@
   implementations. This means (expert) users can override the number of
   partitions when loading data, in cases where the default heuristic doesn't
   work well. Also changes the number of partitions in case there are fewer
-  frames than workers, where it is an advantage to more have small (1-frame)
+  frames than workers, where it is an overall advantage to have small (1-frame)
   partitions instead of aggregating all frames into a single partition,
   especially if there is a lot of processing done per frame (:issue:`1701`
   :pr:`1702`).
--- a/docs/source/changelog/features/num-partitions.rst
+++ b/docs/source/changelog/features/num-partitions.rst
@@ -1,0 +1,10 @@
+[Feature] Allow overriding the number of partitions
+===================================================
+* Adds a parameter :code:`num_partitions` to most :code:`DataSet`
+  implementations. This means (export) users can override the number of
+  partitions when loading data, in cases where the default heuristic doesn't
+  work well. Also changes the number of partitions in case there are fewer
+  frames than workers, where it is an advantage to more have small (1-frame)
+  partitions instead of aggregating all frames into a single partition,
+  especially if there is a lot of processing done per frame (:issue:`1701`
+  :pr:`1702`).
--- a/docs/source/changelog/features/num-partitions.rst
+++ b/docs/source/changelog/features/num-partitions.rst
@@ -1,7 +1,7 @@
 [Feature] Allow overriding the number of partitions
 ===================================================
 * Adds a parameter :code:`num_partitions` to most :code:`DataSet`
-  implementations. This means (export) users can override the number of
+  implementations. This means (expert) users can override the number of
   partitions when loading data, in cases where the default heuristic doesn't
   work well. Also changes the number of partitions in case there are fewer
   frames than workers, where it is an advantage to more have small (1-frame)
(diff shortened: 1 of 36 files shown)


src/libertem/io/dataset/base/dataset.py
```diff
@@ -81,6 +86,8 @@ def get_num_partitions(self) -> int:
         native dtype. At least :code:`self._cores` partitions
         are created.
         """
+        if self._user_num_partitions is not None:
+            return self._user_num_partitions
```
matbryan52, line 90: Since there is no situation where zero or negative partitions could make sense, better to clip this here early rather than get a strange index out of range or div-by-0 error later? Could also `raise` explicitly of course.

```suggestion
            return max(1, self._user_num_partitions)
```

src/libertem/io/dataset/raw.py
```diff
@@ -95,10 +95,27 @@ class RawFileDataSet(DataSet):
     dtype: numpy dtype
         The dtype of the data as it is on disk. Can contain endian indicator, for
         example >u2 for big-endian 16bit data.
+
+    num_partitions: int, optional
+        Override the number of partitions. This is useful if the
+        default heuristic doesn't work well.
```
matbryan52, line 101: Perhaps rather than "heuristic" we could mention why we added this option.


```suggestion
        Override the number of partitions. This is useful if the
        default number of partitions, chosen based on common workloads,
        creates partitions which are too large (or small) for the UDFs
        being run on this dataset.
```

src/libertem/io/dataset/base/partition.py
```diff
@@ -73,11 +73,11 @@ def make_slices(cls, shape, num_partitions, sync_offset=0):
         num_frames = shape.nav.size
         if num_partitions > num_frames:
             warnings.warn(
-                "dataset contains fewer frames than partitions, "
-                "setting num_part to 1 to allow processing (use fewer workers?)",
+                f"dataset contains fewer frames than partitions, "
+                f"setting num_partitions to {num_frames} to allow processing",
```
matbryan52, line 77: Improving my warning message from the last PR on this:

```suggestion
                "dataset contains fewer frames than specified partitions, "
                f"setting num_partitions == num_frames == {num_frames} "
                "to avoid creating empty partitions",
```

matbryan52 (changes requested):
Aside the comment / docstring changes, I don't see any issue with this. 

I do also think there are a few places in the test suite where we already force a number of partitions using `ds.set_num_cores(n)` that could be replaced with this feature?

sk1p:
The intermittent `web_api` failure and consecutive hang seems unrelated; re-running fixed it and I couldn't reproduce locally. This should be ready to be merged now.

docs/source/changelog/features/num-partitions.rst
```diff
@@ -0,0 +1,10 @@
+[Feature] Allow overriding the number of partitions
+===================================================
+* Adds a parameter :code:`num_partitions` to most :code:`DataSet`
+  implementations. This means (expert) users can override the number of
+  partitions when loading data, in cases where the default heuristic doesn't
+  work well. Also changes the number of partitions in case there are fewer
+  frames than workers, where it is an advantage to more have small (1-frame)
```
uellue, line 7:
```suggestion
  frames than workers, where it is an overall advantage to have small (1-frame)
```

uellue (approved):
Looks simple enough, great!

Just to make sure, the code path where the number of partitions is not set is also tested somewhere somehow for all the datasets, right? Just to make sure that the default for an optional parameter actually works. Otherwise it could slip through if we refactor the code later and require a different default etc.

sk1p:
> Just to make sure, the code path where the number of partitions is not set is also tested somewhere somehow for all the datasets, right? Just to make sure that the default for an optional parameter actually works. Otherwise it could slip through if we refactor the code later and require a different default etc.

Good point! I'll add new test cases to explicitly make sure these cases are covered.

sk1p:
@uellue @matbryan52 should be ready to go now. If you agree, feel free to press merge :sparkles:

PR #1699, opened 2025-02-18
Fix deprecation warning on sparse.COO coords

uellue:
Convert to ndarray or use alternative methods.

Also include filter in pytest.ini that makes this deprecation warning an error during tests.

## Contributor Checklist:

* [x] I have added or updated my entry in [the creators.json file](https://github.com/LiberTEM/LiberTEM/blob/master/packaging/creators.json)
* [N/A] I have added [a changelog entry](https://github.com/LiberTEM/LiberTEM/tree/master/docs/source/changelog) for my contribution
* [N/A] I have added/updated documentation for all user-facing changes
* [x] I have added/updated test cases
* [N/A] I have included the [rebuilt production build of the client](https://libertem.github.io/LiberTEM/contributing.html?#building-the-client) (only if changes were made to the GUI)

## Reviewer Checklist:

* [x] `/azp run libertem.libertem-data` passed
* [x] Only import code that is compatible with MIT license

PR #1710, opened 2025-03-27
Use more sparseconverter

### LINKED ISSUE #1373: Better separate ROI handling from other sparse conversions?; port libertem.common.sparse to sparseconverter
https://github.com/LiberTEM/LiberTEM/blob/1b823125adc681263cbbf828ef55a999ad513bdf/src/libertem/common/sparse.py#L19-L43

Currently this function is responsible for interpreting the ROI argument, but its behavior may also bleed into other parts of LiberTEM like the MaskContainer. Maybe the ROI argument interpretation should have its own function so that these concerns can be separated better?

In the same direction, the functions in libertem.common.sparse could now be ported to the new sparseconverter spin-off package to remove duplication.

uellue:
Closes #1373

The rest is a bit specific and would require re-engineering the MaskContainer, ApplyMasksEngine and their use.

Most transition to sparseconverter was already done when fixing the COO coordinate deprecation in #1699.

## Contributor Checklist:

## Reviewer Checklist:

* [N/A] Only import code that is compatible with MIT license

### DIFF
--- a/src/libertem/common/sparse.py
+++ b/src/libertem/common/sparse.py
@@ -10,12 +10,11 @@ if TYPE_CHECKING:
 
 
 def to_dense(a):
-    if isinstance(a, sparse.SparseArray):
-        return a.todense()
-    elif sp.issparse(a):
-        return a.toarray()
-    else:
+    # If unsupported by sparseconverter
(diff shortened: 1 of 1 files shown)


sk1p (approved):
LGTM!

PR #1712, opened 2025-03-28
Rework GMS examples

### LINKED ISSUE #1418: DM integration examples freeze DM
## How to reproduce

Run at least the inline or local example for DM from the examples folder. Win11, version 3.52.3932.0.

## What happens

The "local" example seems to start the cluster. At some point after that it freezes DM, though.

Deferred to 0.12 release for now, to be tested with latest DM version and fresh DM Python environment.

uellue:
They work flawlessly in GMS 3.61.4719.0! Also, no nested threads required since the fixes in GMS. :yay:

- Use publicly available file
- gms.py: Make sure we don't have a view because GMS doesn't like it
- Replace inline with threads
- Update to current API
- Include live plot in both examples instead of separate one
- Don't teach starting a process-based executor -- threads or external are better choices.

Closes #1418

## Contributor Checklist:

## Reviewer Checklist:


docs/source/changelog/bugfix/gms.rst
```diff
@@ -0,0 +1,3 @@
+[Bugfix] Update GMS integration
+===============================
+* Update GMS examples and fix live plotting for GMS 3.61 (:issue:`1418`, :pr:`1712``).
```
sk1p, line 3:
```suggestion
* Update GMS examples and fix live plotting for GMS 3.61 (:issue:`1418`, :pr:`1712`).
```

sk1p (approved):
LGTM! Small typo in the changelog

PR #1715, opened 2025-04-10
Support sparse 0.16

uellue:
The previous implementation throws an error with the newly released sparse 0.16.0.

Replace product with custom function that works directly on the COO data structure since I didn't find an efficient solution for a dense-sparse product that remains sparse and supports broadcasting as required.

Note: This only works with 2D sig dimensions. However, it seems that radial_bins() only supports 2D results anyway.

Drive-by:

- Update tests cases to use assert_allclose()
- Use complex64 for masks by default. Previously, values were cast to complex128 on the way, likely by accident.

## Contributor Checklist:

## Reviewer Checklist:


uellue:
See https://github.com/hyperspy/rosettasciio/issues/390 for cause of notebook failure.

src/libertem/analysis/radialfourier.py
```diff
@@ -113,30 +118,49 @@ def stack():
...
 
+@numba.njit(cache=True, fastmath=True, nogil=True)
+def _radial_mask_product(ring_stack_coords, ring_stack_data_inout, modulator):
+    '''
+    Perform the product between rings and modulator for radial_mask_factory()
+
+    Work on the COO data structure since `sparse` doesn't support
+    efficient sparse-dense product with broadcasting since 0.16, apparently.
+    '''
+    for i in range(ring_stack_data_inout.shape[0]):
```
sk1p, line 157: For coverage, should we mark at least one of the test cases as `with_numba`?

src/libertem/analysis/radialfourier.py
```diff
@@ -113,30 +118,49 @@ def stack():
...
+                ring_stack_data_inout=ring_stack.data,
+                modulator=modulator
+            )
         else:
             ring_stack = rings[:, np.newaxis, ...] * modulator
         return ring_stack.reshape((-1, detector_y, detector_x))
     return stack
 
 
+@numba.njit(cache=True, fastmath=True, nogil=True)
```
sk1p, line 149: Is `fastmath` needed for performance? I'd rather only add it if it brings a significant speed-up; as I understand it, there are some safety concerns, for example NaN-values causing UB etc.

tests/analysis/test_analysis_radialfourier.py
```diff
@@ -65,6 +66,14 @@ def test_sparse(ds_random, lt_ctx):
     lt_ctx.run(analysis)
 
 
+def test_sparse_multi(ds_random, lt_ctx):
```
sk1p, line 69:
```suggestion
@pytest.mark.with_numba
def test_sparse_multi(ds_random, lt_ctx):
```

sk1p (commented):
Looks good to me overall, see some small comments and suggestions inline

tests/analysis/test_analysis_radialfourier.py
uellue: Thx! I put it on the other test since it is a smaller mask stack, I think.

PR #1726, opened 2025-05-26
Find existing deposition by concept ID

### LINKED ISSUE #1645: Fix Zenodo upload script
Since recently the script cannot update an existing draft anymore, instead it triggers a "404". We should figure out what is going on and fix the script.

As a work-around one can drop any drafts on Zenodo before minting a new RC or release.

uellue:
Unfortunately it is no longer possible to find a draft in the previous way. Instead the script now lists all depositions and matches them by concept ID to find the right one.

Closes #1645 

Note: I've already added the LT_RELEASE_UPLOAD_ZENODO_CONCEPT variable to the repo.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/scripts/zenodo_upload
+++ b/scripts/zenodo_upload
@@ -370,12 +370,12 @@ def make_data():
 
 
 @click.command()
-@click.option('--parent', default="264899")
+@click.option('--concept', default="260887")
 @click.option('--url', default='https://sandbox.zenodo.org/api/')
 @click.option('--verbose', '-v', is_flag=True)
 @click.option('--mask-zenodo-exception', is_flag=True)
 @click.argument('files', nargs=-1, required=True, type=click.File())
-def main(files, parent, url, verbose, mask_zenodo_exception):
+def main(files, concept, url, verbose, mask_zenodo_exception):
     try:
         # raise ZenodoException("Test")
         global BASE_URL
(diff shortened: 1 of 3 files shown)


scripts/zenodo_upload
```diff
@@ -384,17 +384,34 @@ def main(files, parent, url, verbose, mask_zenodo_exception):
...
+        if submitted:
+            log("Creating new version of deposition %s..." % deposition_id)
+            r = new_version(parent_id=deposition_id)
             prettylog(r)
+            deposition_id = r.json()['id']
 
-        deposition_id = get_latest_draft(r)
         log("Updating draft deposition %s..." % deposition_id)
         r = update_deposition(deposition_id=deposition_id, data=make_data())
         prettylog(r)
```
sk1p, line 417: So this means that in any case, `update_deposition` will be called?

uellue: Yes exactly: It looks for depositions that match the concept. A deposition can be identical to the record, in that case `deposition['submitted'] is True`. If a draft deposition exists, it will find that instead where `deposition['submitted'] is False`. So far it returned either the record or the draft, but I'm not sure if that will always be the case. For that reason I check if there are multiple entries for the same concept and error out to help future me figure it out.

If the script finds a submitted record, it creates a new draft for it. After that the deposition_id is guaranteed :tm:  to be an unsubmitted draft, and the script updates it as before.

The bug appeared when the API returned the ID of the record even if a draft existed, without a way to figure out what the ID of an existing draft is. In the next steps updating the record failed because that only works with drafts. It worked without existing draft because it just created the new draft and then used the ID from that return value for the next steps. That way it got the ID of the freshly created draft. Essentially, it continued working by accident.

PR #1743, opened 2025-06-13
Make constraints more selective

uellue:
How to reproduce
================

Install LiberTEM in GMS_VENV_PYTHON for Digital Micrograph 3.61 that requires Python 3.10.9 and NumPy 1.23.5

`pip install libertem==0.15.0 numpy==1.23.5`

What happens
============

Unresolvable version conflict because numba>=0.61 depends on NumPy 2

What should happen
==================

Minimum Numba version on 3.10 is 0.55

Analysis
========

The previous constraint `"numba>=0.61;python_version < '3.14' and python_version >= '3.10'"` seems to be responsible. It looks like the last matching one was taken, not the first.

Fix
===

Make sure only one numba constraint matches per Python version. With these updated constraints it installs `numba-0.60.0` without version conflict.

## Contributor Checklist:

## Reviewer Checklist:

* [ ] Only import code that is compatible with MIT license

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,10 +59,10 @@ dependencies = [
     # Minimum constraints of numba for all Python versions we support
     # See https://numba.readthedocs.io/en/stable/release-notes-overview.html
     "numba>=0.53;python_version < '3.10'",
-    "numba>=0.55;python_version < '3.11'",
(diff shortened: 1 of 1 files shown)


uellue:
Workflow name updated directly in master. Workflow seems to work as designed!

PR #1759, opened 2025-07-22
Make GPU processing available on the main process

uellue:
* Parameter to `Context.make_with()` to specify GPU
* Activate GPU in relevant parts of `UDFRunner`
* Make available through the `UDF.xp` property

Selecting a GPU in CuPy is confirmed to be thread local. Included tests that confirm correctly setting `UDF.xp` for main process functions without interfering with `UDF.xp` in functions that run on workers.

This change may break UDFs that have previously used `self.xp` on the main node and expected to always receive NumPy. However, given the documentation of `UDF.xp`, users should always expect one of NumPy or CuPy from this property, so no change of documented API behavior.

Drive-by: Fix title of changelog entry; remove duplicate `detect()` call.

## Contributor Checklist:

## Reviewer Checklist:

### DIFF
--- a/src/libertem/api.py
+++ b/src/libertem/api.py
@@ -199,6 +199,14 @@ class Context:
 
         .. versionadded:: 0.7.0
 
+    main_process_gpu : int or None
+
+        .. versionadded:: 0.16.0
+
+        GPU to make available for UDFs on the main process,
+        i.e. for merging and computing results. Currently, this switches
+        :attr:`libertem.udf.base.UDF.xp` to CuPy if available.
+
     Attributes
     ----------
 
@@ -223,12 +231,14 @@ class Context:
         self,
         executor: Optional[JobExecutor] = None,
         plot_class: Optional['Live2DPlot'] = None,
+        main_process_gpu: Optional[int] = None,
     ):
         import traceback
         if executor is None:
             executor = self._create_local_executor()
         self.executor = executor
         self._plot_class = plot_class
+        self._main_process_gpu = main_process_gpu
         self._created_at = traceback.format_stack()
         self._register_at_exit()
 
@@ -241,6 +251,7 @@ class Context:
         gpus: Optional[Union[int, Iterable[int]]] = None,
         plot_class: Optional['Live2DPlot'] = None,
         snooze_timeout: Optional[float] = None,
+        main_process_gpu: Optional[Union[int, bool]] = None,
     ) -> 'Context':
         '''
         Create a Context with a specific kind of executor.
@@ -323,6 +334,22 @@ class Context:
             environment, for example. The executor is automatically brought back
             up when used again after snoozing. Currently only supported for the
             :class:`~libertem.executor.dask.DaskJobExecutor`.
+        main_process_gpu : int or bool, optional
+
+            .. versionadded:: 0.16.0
+
+            Activate GPU processing on the main process. Currently, this switches
+            :attr:`libertem.udf.base.UDF.xp` to CuPy if available.
+
+            True:
+                Use any available GPU, throw an error if none are available.
+            int:
+                Specify GPU ID to use, throw an error if it is not present.
+            False:
+                Don't use GPU on the main process.
+            None:
+                Default behavior. Currently activates GPU processing if a GPU is
+                available to catch any potential issues with this feature.
 
         Raises
         ------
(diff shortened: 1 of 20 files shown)


src/libertem/api.py
```diff
@@ -223,12 +231,14 @@ def __init__(
...
         executor: Optional[JobExecutor] = None,
         plot_class: Optional['Live2DPlot'] = None,
+        main_process_gpu: Optional[int] = None,
     ):
         import traceback
         if executor is None:
             executor = self._create_local_executor()
         self.executor = executor
         self._plot_class = plot_class
+        self._main_process_gpu = main_process_gpu
```
matbryan52, line 241: A quick question, I might be missing something technical, but does this need to live on the `Context` itself? It seems that so far the details of *how* the computation happens are confined to the various Executors. So `main_process_gpu` could be passed into the executor creation by `make_with`, then the executor itself handles supplying the GPU `xp` during any main process code? (Or the executor directly created with main node GPU enabled, of course, and the Context would not need to know).

uellue: That's a good point! The executor is currently not involved in running the main process part, only the Tasks on the workers. The UDFRunner is ephemeral. Going up the call stack I ended up at the Context to store the state and feed the parameter into the UDFRunner.

If one wanted to let an Executor handle this, one could perhaps use a second one, for example a Threaded one, for the main process? That would also allow running the merge and results functions of several UDFs in parallel in the future.

I'll see how much work it would be to implement a simple version of this, just to get the API right -- `__init__()` should accept a second executor in that case.

I think it would mean adding GPU support to the executor, I'll see what would be the best way for this.

matbryan52: Looking at it a little, could we do something like using the executor itself to call `runner.run_for_dataset_sync`, with something like an `executor.run_main_process` and so it could apply the `use_gpu` context manager as appropriate to its setting?

src/libertem/common/executor.py
```diff
@@ -107,6 +115,38 @@ def enter(self):
...
+        enable : bool, default True
+            Flag that allows using the GPU conditionally without changing the
+            code structure. If set to :code:`False`, this function doesn't
+            perform any action.
+        """
+        if enable:
+            with use_gpu(self._gpu_id):
+                yield
+        else:
+            yield
```
sk1p, line 137: What do you think about not having a separate `use_gpu` contextmanager, but instead having that as part of `enter`? Before, this class had one specific use: you specify the environment with the parameters of the constructor, and you `enter` it. Now, there are basically two environments you can specify with different parameters of the constructor, and two context managers. So I would propose to either 1) integrate into `enter` or 2) have a separate class altogether.

sk1p (approved):
Two main things to improve: a change to `Environment.enter` and `use_gpu`, and possibly getting rid of the executor mixin? Details inside. Thanks!

sk1p:
> Two main things to improve: a change to `Environment.enter` and `use_gpu`, and possibly getting rid of the executor mixin? Details inside. Thanks!

... actually wanted to select "request changes", but in general I approve :)

src/libertem/udf/base.py
```diff
@@ -2058,6 +2064,24 @@ def task_frames(self) -> int:
...
+        previous_id = None
+        try:
+            # Avoid importing if not used
+            import cupy
+            previous_id = cupy.cuda.Device().id
+            cupy.cuda.Device(device).use()
+            yield
+        finally:
+            if previous_id is not None:
+                cupy.cuda.Device(previous_id).use()
```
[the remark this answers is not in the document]
uellue: Unfortunately it was hard to reproduce. It happened sometimes with tests in async mode. What about leaving it for now until we hit a more stable reproducer or in case it turns out to also affect the current solution?

uellue:
Hm, coverage report looks odd

src/libertem/executor/base.py
```diff
@@ -37,6 +55,20 @@ def ensure_async(self, pool=None):
         """
         return AsyncAdapter(wrapped=self, pool=pool)
 
+    def run_process_local(self, task: GenericTaskProtocol, args=(), kwargs={}):
+        """
+        run a callable :code:`fn` in the context of the current process.
+        """
+        env = self._get_local_env()
+        return task(args, kwargs, environment=env)
```
sk1p, line 63: Small issue here: the `kwargs` are a quasi-global variable; if a first task is called with the default `kwargs` and calls, for example, `kwargs.update(...)`, the next calls of `run_process_local` with default `kwargs` will get the updated `kwargs`, too.

sk1p (commented):
:+1: generally looks good, see one last issue in the comments!

One small issue, as I re-read the code: as we set the `cupy` GPU ID for a larger block of code, users of `run_udf_iter` will have their `cupy` GPU ID changed, too, at least in the thread that calls `run_udf_iter`. So maybe benchmarking the other option of only enabling the environment while calling `merge` etc. is needed?

sk1p (approved):
LGTM!

